PR #45, opened 2016-12-10
Moving to use Spring Boot for configuration using dependency injection

ppatierno:


vietj:
my 2 cents : I have nothing against Spring Boot, but with this choice you force your users to use it.

ppatierno:
@vietj Nope. The configuration object can be instantiated even without Spring Boot and then set to the bridge for its configuration. For example this is the way I use in the unit tests where Spring Boot isn't used.

vietj:
doesn't it force to use Spring as dependency ?

ppatierno:
Yes on this you are right but this bridge isn't supposed to be a generic component. It's an application for bridging AMQP to/from Kafka that the user should use as it is. It's not like other Vert.x components like AMQP bridge or MQTT server which have only on side well defined (AMQP and MQTT) and are then open on the other side.

vietj:
imho (that's just my opinion), I would provide a Spring independant Verticle that can be deployed in Vert.x without Spring and I would provide a Spring Boot enabled component that just configures the Verticle from Spring Boot.

vietj:
i.e I would provide a Vert.x service (http://vertx.io/docs/vertx-service-factory/java/) and a Spring Boot component that configures and deploys this Vert.x Service.

vietj:
I can contribute that if you are ok

ppatierno:
I agree on that I should avoid Spring Boot dependency. Regarding the verticle configuration do you think that DeploymentOptions using an external json file could be a good choice ? As described following :  

http://vertx.io/blog/vert-x-application-configuration/

The big advantage of using Spring Boot is the possibility to have property configured and mapped directly to environment variables that are most used in a microservices scenario using Kubernetes and OpenShift.

PR #55, opened 2017-04-20
Added support to retrieve topic name

pinkyrathore:
@ppatierno please review and let me know your thoughts

We have a requirement to retrieve the topic name at receiver end, so added support for it.
This also fixes issue https://github.com/EnMasseProject/amqp-kafka-bridge/issues/54

### DIFF
--- a/src/test/java/enmasse/kafka/bridge/example/BridgeReceiver.java
+++ b/src/test/java/enmasse/kafka/bridge/example/BridgeReceiver.java
@@ -139,7 +139,8 @@ public class BridgeReceiver {
 								Object partition = messageAnnotations.getValue().get(Symbol.getSymbol("x-opt-bridge.partition"));
 								Object offset = messageAnnotations.getValue().get(Symbol.getSymbol("x-opt-bridge.offset"));
 								Object key = messageAnnotations.getValue().get(Symbol.getSymbol("x-opt-bridge.key"));
-								LOG.info("... on partition {} [{}], key = {}", partition, offset, key);
+								Object topicName = messageAnnotations.getValue().get(Symbol.valueOf("x-opt-bridge.topic"));
+								LOG.info("... on topic {} partition {} [{}], key = {}", topicName, partition, offset, key);
 							}
 						})
 						.open();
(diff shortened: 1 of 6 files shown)


pinkyrathore:
After running some more tests, I found one more issue -

Scenario -

Messages are delivered in order but they are marked committed out of order.

Example -

Offsets = 0_61, 0_62, 0_63, 0_64, 0_65

1. Message 61 is tracked, delivered but **not committed yet**. 

Offset state -

firstUsettledOffset (now) = 61
offsetsFlag = true

2. Message 62 tracked and delivered 

It will go in this condition because 61 is not committed yet -
 
else if (offset > this.firstUnsettledOffsets.get(partition)) { // do nothing }

Offset state -

offsetSettlements = {62=true}
firstUsettledOffset = 61
offsetsFlag = true

3. Message 61 marked committed

Offset state -

firstUsettledOffset (now) = 62
offsetsFlag = false

4. Message 63, 64, 65 also marked delivered but go into below condition 
else if (offset > this.firstUnsettledOffsets.get(partition)) { // do nothing }

Now, as offsetsFlag has been set to false for the partition in step 3, getOffsets method will not find any unsettled offsets. As a result, the offsets from 62 to 65 are never committed in kafka or in local map and results in **LAG**

pinkyrathore:
Sure.

pinkyrathore:
@ppatierno I have removed fix for issue 54 from the PR. it's only the enhancement now.

src/test/java/enmasse/kafka/bridge/example/BridgeReceiver.java
```diff
@@ -139,7 +139,8 @@ public void run(Vertx vertx) {
 								Object partition = messageAnnotations.getValue().get(Symbol.getSymbol("x-opt-bridge.partition"));
 								Object offset = messageAnnotations.getValue().get(Symbol.getSymbol("x-opt-bridge.offset"));
 								Object key = messageAnnotations.getValue().get(Symbol.getSymbol("x-opt-bridge.key"));
-								LOG.info("... on partition {} [{}], key = {}", partition, offset, key);
+								Object topicName = messageAnnotations.getValue().get(Symbol.valueOf("x-opt-bridge.topic"));
```
ppatierno: can you change this just to "topic" (not "topicName") to be consistent with the annotation that is ok.

pinkyrathore: Done.

ppatierno:
@pinkyrathore I merged this PR for completeness, btw the receiver should already knows the topic when a message is received because it's in the attached address (i.e. my_topic/group.id/my_group, so the topic is "my_topic"). When it receives a message on an AMQP receiver link it has the information about topic (in the address) through the "source" terminal.

pinkyrathore:
@ppatierno ok. yes, that's right. Thanks for pointing it out.

PR #57, opened 2017-04-27
Start Kafka cluster within tests (copied from vert-x3/vertx-kafka-cli…

tombentley:
…ent)

The new files KafkaClusterTestBase.java and KafkaTestBase.java are copied straight from https://github.com/vert-x3/vertx-kafka-client/blob/master/src/test/java/io/vertx/kafka/client/tests/ which is also ASL 2 licensed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,8 @@
 		<junit.version>4.12</junit.version>
 		<vertx-unit.version>3.2.1</vertx-unit.version>
 		<spring-boot.version>1.4.2.RELEASE</spring-boot.version>
+		<debezium.version>0.4.0</debezium.version>
+
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -66,6 +69,31 @@
 			<version>${vertx-unit.version}</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.vertx</groupId>
+			<artifactId>vertx-kafka-client</artifactId>
+			<version>${vertx-kafka-client.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
```
ppatierno: I guess that this dependency isn't needed just for the tests part. It should be part of the next step of using the Vert.x Kafka client. Isn't it ?

src/test/java/enmasse/kafka/bridge/KafkaClusterTestBase.java
```diff
@@ -0,0 +1,48 @@
+package enmasse.kafka.bridge;
```
ppatierno: Could you add the license header please ?

src/test/java/enmasse/kafka/bridge/KafkaTestBase.java
```diff
@@ -0,0 +1,77 @@
+package enmasse.kafka.bridge;
```
ppatierno: I guess that this PR should be useful only for adding the possibility to run tests without the need of an external Kafka cluster right ? So this class should not be needed. We don't use it in the tests we already have in the bridge. For now we need the Before and After methods for starting the embedded Kafka cluster. You could get rid of the "extends KafkaTestBase" in the previous KafkaClusterTestBase. It's possible that we'll re-add it when moving to use the Vert.x Kafka client, but let's do step by step. wdyt ?

pom.xml
tombentley: Indeed, a subsequent patch will change it to compile scope. I can make it compile scope from the start if you prefer.

src/test/java/enmasse/kafka/bridge/KafkaClusterTestBase.java
tombentley: Sure. tbh I wasn't sure whether to include the headers or not (since both projects are ASL2 and none of the other files in this project have headers), but having a heading in one file and not in the other is careless.

tombentley:
@ppatierno changes made.

pom.xml
ppatierno: I would prefer remove it for now if not necessary in order to have a consistent patch. We'll add the Kafka client then, with the other related PR.

tombentley:
@ppatierno vertx-kafka-client now with compile scope.

tombentley:
Sorry I managed to drop a file, now fixed.

PR #59, opened 2017-05-10
OffsetTracker

tombentley:
Tidies up the `OffsetTracker` API and implementations in a number of ways:

1. Drops using the opaque `String tag` in the API for `int partition` and `long offset`, since the offset tracker doesn't need to know about the tags used for the vertx `EventBus`
2. Refactors the implementations of `OffsetTracker` to avoid using multiple `Map`s

### DIFF
--- a/src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
@@ -451,7 +451,7 @@ public class SinkBridgeEndpoint implements BridgeEndpoint {
 		
 		if (partition == null && offset != null) {
 			// no meaning only offset without partition
-			condition = new ErrorCondition(Symbol.getSymbol(Bridge.AMQP_ERROR_NO_PARTITION_FILTER), "No partition filter specied");
+			condition = new ErrorCondition(Symbol.getSymbol(Bridge.AMQP_ERROR_NO_PARTITION_FILTER), "No partition filter specified");
 			return condition;
 		}
 		
--- a/src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
@@ -293,14 +293,18 @@ public class SinkBridgeEndpoint implements BridgeEndpoint {
 								Object obj = this.vertx.sharedData().getLocalMap(this.context.getEbName()).remove(deliveryTag);
 								
 								if (obj instanceof KafkaMessage<?, ?>) {
-	
+									// get partition and offset from delivery tag : <partition>_<offset>
+									int underscore = deliveryTag.indexOf("_");
+									int messagePartition = Integer.valueOf(deliveryTag.substring(0, underscore));
+									long messageOffset = Long.valueOf(deliveryTag.substring(underscore + 1));
+									
 									KafkaMessage<String, byte[]> kafkaMessage = (KafkaMessage<String, byte[]>) obj;
 									record = kafkaMessage.getRecord();
 									
 									Message message = converter.toAmqpMessage(this.sender.getSource().getAddress(), record);
 									
 									// record (converted in AMQP message) is on the way ... ask to tracker to track its delivery
-									this.offsetTracker.track(deliveryTag, record);
+									this.offsetTracker.track(messagePartition, messageOffset, record);
 									
 									LOG.debug("Tracked {} - {} [{}]", record.topic(), record.partition(), record.offset());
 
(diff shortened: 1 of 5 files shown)


src/main/java/enmasse/kafka/bridge/tracker/SimpleOffsetTracker.java
```diff
@@ -32,52 +32,53 @@
...
 		
 		if (this.offsets.containsKey(partition)) {
 			
 			// map already contains partition but to handle "out of order" delivery
 			// we have to check that the current delivered offset is greater than
 			// the last committed offset
-			if (offset > this.offsets.get(partition)) {
-				this.offsets.put(partition, offset);
-				this.offsetsFlag.put(partition, true);
+			PartitionState x = this.offsets.get(partition);
```
ppatierno: Can we use something like "state" or "partitionState" as variable name instead of just "x" ?

src/main/java/enmasse/kafka/bridge/tracker/SimpleOffsetTracker.java
```diff
@@ -109,9 +109,10 @@ public synchronized void commit(Map<TopicPartition, OffsetAndMetadata> offsets)
 			
 				// if offset tracked isn't changed during Kafka committing operation 
 				// (it means no other messages were acknowledged)
-				if (this.offsets.get(offset.getKey().partition()) == offset.getValue().offset()) {
+				PartitionState x = this.offsets.get(offset.getKey().partition());
```
ppatierno: Can we use something like "state" or "partitionState" as variable name instead of just "x" ?

src/main/java/enmasse/kafka/bridge/tracker/SimpleOffsetTracker.java
```diff
@@ -32,52 +32,53 @@
  */
 public class SimpleOffsetTracker<K, V> implements OffsetTracker<K, V> {
 
+	private static class PartitionState {
+		public long offset;
+		public boolean changed;
```
ppatierno: In general I prefer to use private fields and getter/setter. What do you think about that ? Just to know your opinion ;)

src/main/java/enmasse/kafka/bridge/tracker/FullOffsetTracker.java
```diff
@@ -40,15 +40,31 @@
...
+	 */
+	private static class PartitionState {
+		/** Insertion-ordered map of offset to settlement */
+		public Map<Long, Boolean> settlements;
+		/** The current offset of this partition */
+		public long offset;
+		// has this message been received by remote AMQP peer
+		public boolean flag;
+		
+		public long firstUnsettled;
```
ppatierno: In general I prefer to use private fields and getter/setter. What do you think about that ? Just to know your opinion ;)

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -293,22 +293,26 @@ public void handle(ProtonLink<?> link) {
 								Object obj = this.vertx.sharedData().getLocalMap(this.context.getEbName()).remove(deliveryTag);
 								
 								if (obj instanceof KafkaMessage<?, ?>) {
-	
+									// get partition and offset from delivery tag : <partition>_<offset>
+									int underscore = deliveryTag.indexOf("_");
+									int messagePartition = Integer.valueOf(deliveryTag.substring(0, underscore));
+									long messageOffset = Long.valueOf(deliveryTag.substring(underscore + 1));
```
ppatierno: just "partition" and "offset" instead of long names ?

src/main/java/enmasse/kafka/bridge/tracker/SimpleOffsetTracker.java
tombentley: Sure

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
tombentley: `partition` and `offset` are being used in a enclosing scope so it would be a compile time error to declare them.

src/main/java/enmasse/kafka/bridge/tracker/SimpleOffsetTracker.java
tombentley: I would usually agree, and I don't exactly disagree in this particular case. At the time of writing it was easier to go though statement by statement replacing the old code with the new using fields.

In this particular case the inner class never leaks outside the enclosing class, and the code is all pretty trivial: It's easy to see the field readers and writers because they're all in the same file. Adding getters and setter would just make the code longer for no benefit, imo. In a large class, or if the inner class was externally accessible I would certainly use accessors. 

I'll happily encapsulate if that's what you prefer.

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
ppatierno: yes ... sorry you are right. Sometime we should check if we can move this "big" inline callback to a method in order to avoid something like that, because offset and partition are declared outside the callback here. For now leave as it is please.

src/main/java/enmasse/kafka/bridge/tracker/SimpleOffsetTracker.java
ppatierno: I got your point. Agree. ;)

PR #60, opened 2017-05-17
Vertx sink

tombentley:
These commits refactor the BridgeSinkEndpoint to use Vertx's `KafkaConsumer` instead of using the native Kafka `Consumer` running in a separate thread. 

It also adds some mock based tests. There is definitely scope to add further tests.

There is still scope to tidy up our error handling (at the least factoring out some methods to do the AMQP detach in the presence of errors when setting up the link), but for now I propose to do this in a separate PR.


pom.xml
```diff
@@ -12,14 +12,16 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
-		<kafka.version>0.10.1.0</kafka.version>
+		<kafka.version>0.10.2.0</kafka.version>
```
ppatierno: Let's review the needed Kafka version as for the #61

pom.xml
```diff
@@ -12,14 +12,16 @@
...
 		<slf4j-log4j12.version>1.7.18</slf4j-log4j12.version>
 		<log4j.version>1.2.17</log4j.version>
 		<vertx-proton.version>3.4.2-SNAPSHOT</vertx-proton.version>
+		<vertx-kafka-client.version>3.4.1</vertx-kafka-client.version>
 		<junit.version>4.12</junit.version>
 		<vertx-unit.version>3.2.1</vertx-unit.version>
+		<vertx-kafka-client.version>3.4.2-SNAPSHOT</vertx-kafka-client.version>
 		<spring-boot.version>1.4.2.RELEASE</spring-boot.version>
-		<debezium.version>0.4.0</debezium.version>
+		<debezium.version>0.5.1-SNAPSHOT</debezium.version>
```
ppatierno: Let's review the needed Debezium version as for the #61

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -473,4 +377,258 @@ private ErrorCondition checkFilters(Object partition, Object offset) {
...
+			LOG.debug("Received from Kafka partition {} [{}], key = {}, value = {}", record.partition(), record.offset(), record.key(), record.value());
+			
+			// 1. start message sending
+			protonSend(new KafkaMessage<K, V>(record.partition(), record.offset(), record.record()));
+			
+			if (endOfBatch()) {
+				LOG.debug("End of batch in {} mode => commitOffsets()", this.qos);
+				try {
+					// 2. commit all tracked offsets for partitions
+					commitOffsets(false);
```
ppatierno: Let's imagine that we read 5 records from Kafka cluster and we have to send as AT LEAST ONCE ... in this case we add these 5 records to the offset tracker and send messages. Then in the handler when we are at last record (so endofbatch), the commitOffsets is called but .... no commit is executed because the previous 5 records aren't confirmed yet by the receiver (they are in the offset tracker but not delivered). Imagine that no other messages we'll be received from Kafka (because no one is sending). The handler won't be called anymore and so the commitOffsets won't be called; even if we receive the disposition from receiver and we call the offset tracker delivered, the commit to the Kafka cluster isn't called. In the previous implementation we have the commitOffsets code were executed in any case, trying to commit something if needed because we were in a loop (while) and the poll was executed but if no records were returned the commit part was executed in any case.

tombentley:
@ppatierno I've rebased the PR slightly, `@Ignore`ing the new tests (not all of which pass). This makes travis happy. 

I've pushed the branch `vertx-sink-part-2` which has some further commits refactoring the error handling (which makes the mock tests pass), as well as adding more tests around various error cases which run a Kafka cluster. These latter tests depend on using the SNAPSHOT version of debezium, because they depend on https://github.com/debezium/debezium/pull/235 which hasn't yet made it into a release version.

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -473,4 +388,244 @@ private ErrorCondition checkFilters(Object partition, Object offset) {
...
+		}
+		LOG.debug("Getting partitions for " + kafkaTopic);
+		List<PartitionInfo> availablePartitions = partitionsResult.result();
+		Optional<PartitionInfo> requestedPartitionInfo = availablePartitions.stream().filter(p -> p.getPartition() == this.partition).findFirst();
+		
+		if (requestedPartitionInfo.isPresent()) {
+			LOG.debug("Requested partition {} present", partition);
+			if (this.offset != null) {
+				// Don't consume messages until we've called seek()
+				this.consumer.pause();
```
ppatierno: With latest modification on the Vert.x Kafka client, can we avoid pausing before assigning ans seeking right ?

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -473,4 +388,244 @@ private ErrorCondition checkFilters(Object partition, Object offset) {
...
+							String message = seekResult.cause().getMessage();
+							ErrorCondition condition =
+									new ErrorCondition(Symbol.getSymbol(Bridge.AMQP_ERROR_KAFKA_SUBSCRIBE),
+											"Error getting partition info" + (message != null ? ": " + message : ""));
+							sendProtonError(condition);
+							return;
+						}
+						partitionsAssigned();
+						// Now we can consume messages from the offset we've seek()ed to
+						this.consumer.resume();
```
ppatierno: so as for pause() we don't need this anymore ?

src/main/java/enmasse/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -473,4 +388,244 @@ private ErrorCondition checkFilters(Object partition, Object offset) {
...
+							//sendProtonError(newError(Bridge.AMQP_ERROR_KAFKA_SYNC, "Error in commit", ar.cause()));
+						}
+					}
+				});
+			}
+			// 2. commit succeeded, so we can enqueue record for sending
+			LOG.debug("Received from Kafka partition {} [{}], key = {}, value = {}", record.partition(), record.offset(), record.key(), record.value());
+			
+			// 3. start message sending
+			protonSend(new KafkaMessage<K, V>(record.partition(), record.offset(), record.record()));
```
ppatierno: the previous version used the commitSync() on the native Kafka client so we sent the AMQP messages only when commit went well. With this way we are sending the record even if the commit fails (it's execute asynchronously). It means that AT_MOST_ONCE isn't guaranteed. For example ... the commit fails, the message is sent in any case and if the bridge crashes and then restarts, the client reconnects it could receive the previous message twice. Wdyt ?

tombentley: Yes, I agree. The call to `protonSend()` should be in the `commit()` callback.

ppatierno: Yes but at same time the commit is called only on start batch while we should send messages even for the subsequent records (in the batch).

ppatierno:
Finally after some changes and very long useful discussions I'm going to merge ! ;)

tombentley:
Thanks!

PR #61, opened 2017-05-18
Vertx source

tombentley:
Change the BridgeSourceEndpoint to use the VertX `KafkaProducer` API.


tombentley:
Oops, try with a more suitable `pom.xml`.

PR #67, opened 2017-06-06
Vertx sink part 2

tombentley:
Tidy up various things, and improve tests, follow merge of #60


src/main/java/enmasse/kafka/bridge/Bridge.java
```diff
@@ -396,4 +404,46 @@ private void processOpenSender(ProtonConnection connection, ProtonSender sender)
...
+
+	static void detachWithError(ProtonLink<?> link, ErrorCondition error) {
+		LOG.error("Detaching link {} due to eror {}, description: {}", link, error.getCondition(), error.getDescription());
+		link.setSource(null)
+		.open()
+		.setCondition(error)
+		.close();
+	}
+	
+	static MessageConverter<?, ?> instantiateConverter(ProtonLink<?> link, String className) {
```
ppatierno: aren't we doing too much responsibilities to this method related to send the error on AMQP as well ? Maybe the endpoint sink/source which uses this method for converter creation can check and send AMQP error to its client counter part on the network. Wdyt ?

tombentley: It's a fair point. I was mostly concerned with removing the duplicate code in the sink and source. If we just return null in the event of error (letting the caller call `detachWithError()`) then we lose the detail of exactly what went wrong.

Wrapping up the error string and description into an exception, and catching that in the source and sink is neater, and could also be used by `SinkBridgeEndpoint.checkFilters()`. Such exceptions can't be thrown in the async callbacks though. To prevent that being done accidentally we could make it a checked exception (not usually a fan of checked exceptions, but in this case I think it's useful).

PR #75, opened 2017-09-11
Add AMQPS support (TLS) in the CLIENT mode

scholzj:
In the CLIENT mode, when the environment variable `AMQP_CERT_DIR` / configuration property `amqp.certDir` is specified, the bridge will open the AMQP connection with SASL EXTERNAL and the TLS certificates from the `$AMQP_CERT_DIR` path (ca.crt, tls.crt and tls.key).

Additionally `run_bridge.sh` is now using `$MESSAGING_SERVICE_PORT_AMQPS_BROKER` for the port where to connect. This is the port where link routing is enabled in EnMasse.

### DIFF
--- a/src/main/sh/run_bridge.sh
+++ b/src/main/sh/run_bridge.sh
@@ -11,6 +11,6 @@ fi
 # configuring the bridge to work in "client" mode connecting to the messaging (router) layer
 export AMQP_MODE="CLIENT"
 export AMQP_HOST=$MESSAGING_SERVICE_HOST
-export AMQP_PORT=$MESSAGING_SERVICE_PORT_INTERNAL
+export AMQP_PORT=$MESSAGING_SERVICE_PORT_AMQPS_BROKER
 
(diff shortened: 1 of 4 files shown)


scholzj:
The second part of this change is in EnMasseProject/enmasse#199.

src/main/sh/run_bridge.sh
```diff
@@ -11,6 +11,6 @@ fi
 # configuring the bridge to work in "client" mode connecting to the messaging (router) layer
 export AMQP_MODE="CLIENT"
 export AMQP_HOST=$MESSAGING_SERVICE_HOST
-export AMQP_PORT=$MESSAGING_SERVICE_PORT_INTERNAL
+export AMQP_PORT=$MESSAGING_SERVICE_PORT_AMQPS_BROKER
```
ppatierno: of course this change is needed and I'm ok with that but I have a comment on that. I'll open an issue involving you and Ulf as well to discuss about my proposal.

PR #127, opened 2019-03-22
Fixed #126

### LINKED ISSUE #126: HttpBridgeTest uses deprecated httpclient post methods
io.vertx.core.http.HttpClient
provides post methods that are deprectaed (they will break vert.x 4.x)

consider the usage of vertx-web-client instead.

hifly81:


hifly81:
This PR fixes:
https://github.com/strimzi/strimzi-kafka-bridge/issues/126

PR #138, opened 2019-04-02
HTTP POST producer error handling

stanlyDoge:
Bridge returns:
400 invalid request
422 POST request is empty
40401 topic does not exist
40402 partition does not exist (is higher than topic.partitions)

Not implemented "schema" error codes:
42201, 42202, 42205


src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -45,6 +50,10 @@
     private KafkaProducer<String, byte[]> producerUnsettledMode;
     private KafkaProducer<String, byte[]> producerSettledMode;
 
+    protected final String SEPARATOR = "#";
+    protected final int TOPIC_NOT_FOUND = 40401;
+    protected final int PARTITION_NOT_FOUND = 40402;
```
ppatierno: These are more specific to HTTP, can we move them to the related HTTP source bridge endpoint? Wdyt?

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -80,11 +89,33 @@ protected void handleClose() {
...
 
-        log.debug("Sending to Kafka on topic {} at partition {} and key {}", krecord.topic(), krecord.partition(), krecord.key());
-        if (handler == null) {
-            this.producerSettledMode.write(krecord);
-        } else {
-            this.producerUnsettledMode.write(krecord, handler);
+        Properties props = new Properties();
+        KafkaConfig consumerConfig = this.bridgeConfigProperties.getKafkaConfig();
+        props.put("bootstrap.servers", consumerConfig.getBootstrapServers());
+        AdminClient ac = AdminClient.create(props);
```
ppatierno: not sure that we need the `AdminClient` which checks stuff before sending ... this will drive a slower client experience I guess. You should check what kind of exceptions are raised just sending a message when the above conditions happen (topic or partition don't exist). Finally, I am not sure it's the right place to do this check or at the upper level. Remember that this part is shared with the AMQP bridge so it should be agnostic. But we have to think more about that in terms of architecture.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -198,9 +198,18 @@ private void processRequests(HttpServerRequest httpServerRequest) {
                             .end();
                 }
                 break;
-
+            case EMPTY:
+                httpServerRequest.response()
+                        .setStatusCode(422)
+                        .setStatusMessage("records may not be empty")
```
ppatierno: "cannot be"

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -69,16 +69,18 @@ public void handle(Endpoint<?> endpoint) {
 
             // wait for ALL futures completed
             CompositeFuture.join(sendHandlers).setHandler(done -> {
+
                 for (int i = 0; i < sendHandlers.size(); i++) {
                     // check if, for each future, the sending operation is completed successfully or failed
-                    if (done.result().succeeded(i)) {
+                    if (done.result() != null && done.result().succeeded(i)) {
```
ppatierno: good catch!

src/test/java/io/strimzi/kafka/bridge/amqp/AmqpBridgeTest.java
```diff
@@ -109,6 +109,13 @@ public void sendSimpleMessages(TestContext context) {
 		String topic = "sendSimpleMessages";
 		kafkaCluster.createTopic(topic, 1, 1);
 
+		try {
```
ppatierno: why the need for sleep now? The tests were all passing up to now.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeErrorsTest.java
```diff
@@ -0,0 +1,339 @@
...
+        this.vertx.deployVerticle(this.httpBridge, context.asyncAssertSuccess());
+    }
+
+    @After
+    public void after(TestContext context) {
+
+        this.vertx.close(context.asyncAssertSuccess());
+    }
+
+    //@Test
```
ppatierno: is it disabled on purpose?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -74,6 +74,12 @@ public void before(TestContext context) {
         this.httpBridge = new HttpBridge(this.bridgeConfigProperties);
 
         this.vertx.deployVerticle(this.httpBridge, context.asyncAssertSuccess());
+        try {
```
ppatierno: do you really need this sleep? Tests are passing even on Travis without that, are you facing any problems?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -742,4 +773,44 @@ public void receiveSimpleMessageFromPartitionAndOffset(TestContext context) {
 
         deleteAsync.await();
     }
+
+    //@Test
+    public void sendToOnePartitionTest(TestContext context) {
```
ppatierno: is it disabled on purpose?

src/test/java/io/strimzi/kafka/bridge/amqp/AmqpBridgeTest.java
stanlyDoge: I turned off the autocreation of the topics to test non-existing topics. Before, it just created a topic when a message was sent to it.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
stanlyDoge: Yeah. From the API spec I see `POST /topics/(string: topic_name)/partitions/(int: partition_id)` request should do `"Produce messages to one partition of the topic"` but in our bridge, it is fully ignored (AFAICS). So this is the test which we can allow after this is implemented.

ppatierno: But using that endpoint is another way to do that. You can already send to one partition when you specify "partition" field in a record in the "records" collection when you do a POST on `/topics/topic_name`

ppatierno: Ah! I think I got it. You added test but not implementing the feature yet I guess. Well, I would avoid that, it creates just confusion imho. This PR is about producer error handling not implementing that endpoint.

stanlyDoge: Ok, I just wanted to write a test which uses this endpoint. And I did O:) However, it does not cope properly with the situation which is described as "If no partition is provided, one will be chosen based on the hash of the key. If no key is provided, the partition will be chosen for each message in a round-robin fashion."

ppatierno: That description works with the generic `/topics/topic-name` endpoint of course, not this one.

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -80,11 +82,25 @@ protected void handleClose() {
...
         if (handler == null) {
             this.producerSettledMode.write(krecord);
         } else {
-            this.producerUnsettledMode.write(krecord, handler);
+            log.debug("Sending to topic " + krecord.topic() + " at partition {}", krecord.partition());
+            this.producerUnsettledMode.partitionsFor(krecord.topic(), part -> {
+                if (part.succeeded()) {
+                    if (krecord.partition() == null) {
+                        //TODO should not to be!
+                        this.producerUnsettledMode.write(krecord, handler);
```
stanlyDoge: So you say this is applying round-robin somehow and I can just delete this comment? @ppatierno

ppatierno: When the partition is null and the key is null, the producer uses round robin but it's not a logic we have to add it's in the Kafka producer. The code here hasn't to be changed. You should try to catch the right exception for these kinds of errors. Please try and see the right exception raised by the producer.

stanlyDoge: In that case, everything is ok :)

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -86,21 +86,7 @@ protected void send(KafkaProducerRecord<String, byte[]> krecord, Handler<AsyncRe
...
-                        this.producerUnsettledMode.write(krecord, handler);
-
-                    } else {
-                        handler.handle(Future.failedFuture(Integer.toString(ErrorCodeEnum.PARTITION_NOT_FOUND.getValue())));
-                    }
-                } else {
-                    handler.handle(Future.failedFuture(Integer.toString(ErrorCodeEnum.TOPIC_NOT_FOUND.getValue())));
-                }
-            });
+            this.producerUnsettledMode.write(krecord, handler).exceptionHandler(k -> handler.handle(Future.failedFuture(k)));
```
ppatierno: I would change the order setting the exception handler first and before actually writing.

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -80,11 +82,11 @@ protected void handleClose() {
      */
     protected void send(KafkaProducerRecord<String, byte[]> krecord, Handler<AsyncResult<RecordMetadata>> handler) {
 
-        log.debug("Sending to Kafka on topic {} at partition {} and key {}", krecord.topic(), krecord.partition(), krecord.key());
         if (handler == null) {
             this.producerSettledMode.write(krecord);
         } else {
-            this.producerUnsettledMode.write(krecord, handler);
+            log.debug("Sending to topic " + krecord.topic() + " at partition {}", krecord.partition());
+            this.producerUnsettledMode.exceptionHandler(e -> handler.handle(Future.failedFuture(e))).write(krecord, handler);
```
ppatierno: I have just a doubt but you could check easily. Are we sure that the `handler` isn't called twice?
So one time by the exceptionHandler and the other time by the write but with the async result as failed (which is something we check in the handler).

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -96,6 +98,7 @@ public void open() {
         props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, this.bridgeConfigProperties.getKafkaConfig().getProducerConfig().getKeySerializer());
         props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, this.bridgeConfigProperties.getKafkaConfig().getProducerConfig().getValueSerializer());
         props.put(ProducerConfig.ACKS_CONFIG, this.bridgeConfigProperties.getKafkaConfig().getProducerConfig().getAcks());
+        props.put(ProducerConfig.MAX_BLOCK_MS_CONFIG, 10000);
```
ppatierno: why this one?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -194,13 +194,23 @@ private void processRequests(HttpServerRequest httpServerRequest) {
...
+                            .setStatusCode(ErrorCodeEnum.UNKNOWN_REQUEST.getValue())
+                            .setStatusMessage("endpoint not found")
                             .end();
                 }
                 break;
-
+            case EMPTY:
+                httpServerRequest.response()
+                        .setStatusCode(ErrorCodeEnum.EMPTY_REQUEST.getValue())
+                        .setStatusMessage("records cannot not be empty")
```
ppatierno: we don't know which kind of request this is. Cannot say nothing about the payload and what it should contain. It could be a POST request related to commit offsets so it doesn't contain records.

src/main/java/io/strimzi/kafka/bridge/http/ErrorCodeEnum.java
```diff
@@ -0,0 +1,34 @@
+package io.strimzi.kafka.bridge.http;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum ErrorCodeEnum {
+    TOPIC_NOT_FOUND(40401),
+    PARTITION_NOT_FOUND(40402),
+    INVALID_REQUEST(400),
```
ppatierno: let's use HTTP name so "bad request"

src/main/java/io/strimzi/kafka/bridge/http/ErrorCodeEnum.java
```diff
@@ -0,0 +1,34 @@
...
+    UNKNOWN_REQUEST(404),
```
ppatierno: not found

src/main/java/io/strimzi/kafka/bridge/http/ErrorCodeEnum.java
```diff
@@ -0,0 +1,34 @@
...
+    EMPTY_REQUEST(422),
```
ppatierno: unprocessable entity

src/main/java/io/strimzi/kafka/bridge/http/ErrorCodeEnum.java
```diff
@@ -0,0 +1,34 @@
...
+    UNKNOWN_ERROR(500);
```
ppatierno: internal server error

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -47,16 +47,21 @@ public HttpSourceBridgeEndpoint(Vertx vertx, HttpBridgeConfig bridgeConfigProper
...
-        messageConverter = new HttpJsonMessageConverter();
 
         // split path to extract params
         String[] params = httpServerRequest.path().split("/");
 
         // path is like this : /topics/{topic_name}, topic will be at the last position of params
-        String topic = params[params.length - 1];
+        String topic = params[2];
+
+        if (params.length == 5) {
```
ppatierno: Why this distinction? what is param number 4? If it's about partition-id on the URL we should remove that, it's not part of this PR.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
```diff
@@ -30,6 +30,17 @@
...
+    private Integer partition;
+
+    public HttpJsonMessageConverter(int partition) {
+        this.partition = partition;
+    }
+
+    public HttpJsonMessageConverter() {
+        super();
+        this.partition = null;
+    }
```
ppatierno: as said partition-id on the URL is not part of this PR, please remove all the stuff related to it.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
```diff
@@ -38,14 +49,29 @@
...
+                // TODO
+                // If no partition is provided, one will be chosen based on the hash of the key.
+                // If no key is provided, the partition will be chosen for each message in a round-robin fashion.
+                /*if (key != null) {
+                    partition = key.hashCode();
+                } else {
+                    partition = roundRobin();
+                }*/
+                partition = 0;
+            }
```
ppatierno: is that the same here? logic for partition-id on the URL?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeErrorsTest.java
```diff
@@ -0,0 +1,266 @@
...
+                    String statusMessage = offsets.getJsonObject(0).getString("error");
+
+                    context.assertEquals(ErrorCodeEnum.TOPIC_NOT_FOUND.getValue(), code);
+                    context.assertEquals("Topic " + kafkaTopic + " not found", statusMessage);
+                    async.complete();
+                });
+    }
+
+    //@Test
+    public void sendToNonExistingTopicNonExistingPartitionTest(TestContext context) {
```
ppatierno: instead of commenting, let's use @Ignore so that we have trace during the tests that something is ignored and it the future we can re-activate that.

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -80,11 +81,13 @@ protected void handleClose() {
      */
     protected void send(KafkaProducerRecord<String, byte[]> krecord, Handler<AsyncResult<RecordMetadata>> handler) {
 
-        log.debug("Sending to Kafka on topic {} at partition {} and key {}", krecord.topic(), krecord.partition(), krecord.key());
         if (handler == null) {
             this.producerSettledMode.write(krecord);
         } else {
-            this.producerUnsettledMode.write(krecord, handler);
+            log.debug("Sending to topic " + krecord.topic() + " at partition {}", krecord.partition());
```
ppatierno: let's move this log line before the `if` statement, because it's the same for both producers.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -194,13 +194,23 @@ private void processRequests(HttpServerRequest httpServerRequest) {
...
+                            .setStatusCode(ErrorCodeEnum.NOT_FOUND.getValue())
+                            .setStatusMessage("endpoint not found")
                             .end();
                 }
                 break;
-
+            case EMPTY:
+                httpServerRequest.response()
+                        .setStatusCode(ErrorCodeEnum.UNPROCESSABLE_ENTITY.getValue())
+                        .setStatusMessage("records cannot not be empty")
```
ppatierno: I still see "records" in this message. But it could be not a POST request about records but about offsets for example.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -47,16 +47,17 @@ public HttpSourceBridgeEndpoint(Vertx vertx, HttpBridgeConfig bridgeConfigProper
...
         HttpServerRequest httpServerRequest = (HttpServerRequest) endpoint.get();
 
-        messageConverter = new HttpJsonMessageConverter();
 
         // split path to extract params
         String[] params = httpServerRequest.path().split("/");
 
         // path is like this : /topics/{topic_name}, topic will be at the last position of params
-        String topic = params[params.length - 1];
+        String topic = params[2];
```
ppatierno: you are still thinking about having more after the topic_name (so the partition-id) ... please revert this as well :-)

src/main/java/io/strimzi/kafka/bridge/http/RequestIdentifier.java
```diff
@@ -44,6 +45,9 @@ static RequestType getRequestType(HttpServerRequest request){
                 if (method == HttpMethod.POST && params.length == 2){
                     return RequestType.PRODUCE;
                 }
+                if (method == HttpMethod.POST && params.length == 4 && params[2].equals("partitions") && isNumeric(params[3])){
+                    return RequestType.PRODUCE;
+                }
```
ppatierno: still a piece of partition-id in the URL :-)

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
```diff
@@ -30,6 +30,10 @@
 
 public class HttpJsonMessageConverter implements MessageConverter<String, byte[], Buffer, Buffer> {
 
+    public HttpJsonMessageConverter() {
+        super();
+    }
```
ppatierno: why do we need this ctor?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeErrorsTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@RunWith(VertxUnitRunner.class)
+public class HttpBridgeErrorsTest extends KafkaClusterTestBase {
```
ppatierno: do we really need a complete new class for testing errors. Imho we can just having tests about errors in the current class we already have.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -742,4 +748,285 @@ public void receiveSimpleMessageFromPartitionAndOffset(TestContext context) {
...
+                    context.assertEquals(0L, metadata.getLong("offset"));
+                    async.complete();
+                });
+    }
+
+    @Ignore
+    @Test
+    public void partitionPriorityTest(TestContext context) {
+        // partition is specified in the message body and in the POST request
+        // what does have higher priority? Lets say POST one
```
ppatierno: you are right that the priority is on the URL because the client is doing a POST to a specific endpoint but I would return an error about wrong payload because you cannot specify the "partition" in the JSON. Pushing silently, can confuse the user which thinks that we pushed the message to the partition in the JSON while it went to the one in the URL.

Issue #118, opened 2019-03-13
HTTP POST consumer subscription error handling


ppatierno:
Currently, there is no error handling on possible consumer subscription errors (i.e. the consumer instance doesn't exist). We should handle that with proper HTTP status codes.

PR #140, opened 2019-04-09
HTTP POST consumer error handling

stanlyDoge:
@ppatierno do you have any hint how to test "Subscription to topics, partitions and pattern are mutually exclusive." error?

Fixes https://github.com/strimzi/strimzi-kafka-bridge/issues/118


src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -271,7 +271,7 @@ private void automaticPartitionAssignment() {
             this.handleSubscribe(subscribeResult);
 
             if (subscribeResult.failed()) {
-                return;
+                throw new IllegalStateException(subscribeResult.cause());
```
ppatierno: genuine question, did you need this exception for something useful?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -91,7 +91,14 @@ public void handle(Endpoint<?> endpoint) {
                         }
                     });
 
-                    this.subscribe(false);
+                    try {
+                        this.subscribe(false);
+                    } catch (IllegalArgumentException e) {
+                        httpServerRequest.response().setStatusCode(ErrorCodeEnum.SUBSCRIPTION_EXCLUSIVE.getValue())
```
ppatierno: are you sure that for this kind of scenario, an exception is raised by the `subscribe` method?
You are catching the `IllegalArgumentException` you added before but are we sure that the subscribe fails just for this reason and catching the exception is the way to go?

stanlyDoge: No, I am not sure with this at all. Sadly I did not find a way how to test this path and implement it correctly. This is "scratch" how it could be but we need test definitely.

ppatierno: So I would say, let's remove it from this PR and will open an issue for that one.
Remove even raising the `IllegalArgumentException` above.

PR #144, opened 2019-04-11
Modified detection of existing consumer

stanlyDoge:


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -41,7 +41,7 @@
 
     private Map<HttpConnection, SourceBridgeEndpoint> httpSourceEndpoints;
 
-    private Map<String, SinkBridgeEndpoint> httpSinkEndpoints;
+    protected static Map<String, SinkBridgeEndpoint> httpSinkEndpoints;
```
ppatierno: Instead of exposing as static the endpoints collection to ... an HTTP endpoint itself (where you check if the consumerId already exists), we should do the check in the HTTP bridge as we do for the `DELETE` operation.

ppatierno: Offline you noticed that the `name` comes from the body not URL as for `DELETE` so to avoid this `static` field I would ...

* create an `HttpBridgeContext` for holding the source and sink endpoints
* allowing to pass the context to the `HttpSinkBridgeEndpoint` in the ctor so that it can access to the sinks and checking for existing consumer-id

Wdyt?

stanlyDoge: I can do that.

src/main/java/io/strimzi/kafka/bridge/http/ErrorCodeEnum.java
```diff
@@ -7,7 +7,8 @@
     TOPIC_NOT_FOUND(40401),
     PARTITION_NOT_FOUND(40402),
     CONSUMER_NOT_FOUND(40403),
-    SUBSCRIPTION_EXCLUSIVE(40902),
+    CONSUMER_ALREADY_EXISTS(40902),
+    SUBSCRIPTION_EXCLUSIVE(40903),
```
ppatierno: I would remove this subscription exclusive error for now

src/main/java/io/strimzi/kafka/bridge/http/HttpBridgeContext.java
```diff
@@ -0,0 +1,40 @@
...
+package io.strimzi.kafka.bridge.http;
+
+import io.strimzi.kafka.bridge.SinkBridgeEndpoint;
+import io.strimzi.kafka.bridge.SourceBridgeEndpoint;
+import io.vertx.core.http.HttpConnection;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class HttpBridgeContext {
```
ppatierno: a little bit of Javadoc here and on methods

PR #149, opened 2019-04-18
GET consumer error handling

stanlyDoge:
The test for the `GET /consumers/:group_name/instances/:instance/records` endpoint. The missing user situation was already handled before. I do not see any other errors which can happen there.


ppatierno:
This close #149

PR #157, opened 2019-04-23
Improved apicurio configuration

stanlyDoge:
Added error codes + example request bodies.

### DIFF
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -13,6 +13,14 @@
         },
         "version": "1.0.0"
     },
+    "servers": [
+        {
+            "url": "https://bridge.swagger.io/v2"
+        },
+        {
+            "url": "http://bridge.swagger.io/v2"
+        }
+    ],
     "paths": {
         "/consumers/{groupid}": {
             "post": {
@@ -95,6 +103,7 @@
                     "Consumers"
                 ],
                 "summary": "Commit a list of offsets for the consumer",
+                "operationId": "commit",
                 "requestBody": {
                     "$ref": "#/components/requestBodies/Offsets"
                 },
@@ -412,6 +421,7 @@
                     "Consumers"
                 ],
                 "summary": "Overrides the fetch offsets that the consumer will use for the next set of records to fetch",
+                "operationId": "seek",
                 "requestBody": {
                     "$ref": "#/components/requestBodies/Offsets"
                 },
@@ -444,51 +454,7 @@
             ]
         }
     },
-    "tags": [
-        {
-            "name": "Consumers",
-            "description": "Consumers management"
-        },
-        {
-            "name": "Topics",
-            "description": "Topics management"
-        }
-    ],
-    "externalDocs": {
-        "description": "Find out more about Swagger",
-        "url": "http://swagger.io"
-    },
-    "servers": [
-        {
-            "url": "https://petstore.swagger.io/v2"
-        },
-        {
-            "url": "http://petstore.swagger.io/v2"
-        }
-    ],
     "components": {
-        "requestBodies": {
-            "Offsets": {
-                "content": {
-                    "application/json": {
-                        "schema": {
-                            "$ref": "#/components/schemas/Offsets"
-                        }
-                    }
-                },
-                "required": true
-            },
-            "PartitionsForPositions": {
-                "content": {
-                    "application/json": {
-                        "schema": {
-                            "$ref": "#/components/schemas/PartitionsForPositions"
-                        }
-                    }
-                },
-                "required": true
-            }
-        },
         "schemas": {
             "ApiResponse": {
                 "type": "object",
@@ -662,6 +628,42 @@
                 },
                 "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
             }
+        },
+        "requestBodies": {
+            "Offsets": {
+                "content": {
+                    "application/json": {
+                        "schema": {
+                            "$ref": "#/components/schemas/Offsets"
+                        }
+                    }
+                },
+                "required": true
+            },
+            "PartitionsForPositions": {
+                "content": {
+                    "application/json": {
+                        "schema": {
+                            "$ref": "#/components/schemas/PartitionsForPositions"
+                        }
+                    }
+                },
+                "required": true
+            }
         }
+    },
+    "tags": [
+        {
+            "name": "Consumers",
+            "description": "Consumers management"
+        },
+        {
+            "name": "Topics",
+            "description": "Topics management"
+        }
+    ],
+    "externalDocs": {
+        "description": "Find out more about Swagger",
+        "url": "http://swagger.io"
     }
 }
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -42,7 +42,14 @@
                 },
                 "responses": {
                     "200": {
-                        "description": "OK"
+                        "description": "OK",
+                        "content": {
+                            "*/*": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/createdConsumer"
+                                }
+                            }
+                        }
                     },
                     "409": {
                         "description": "Consumer instance with the specified name already exists"
@@ -190,7 +197,39 @@
                 "operationId": "poll",
                 "responses": {
                     "200": {
-                        "description": "OK"
+                        "description": "OK",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/Records2"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": [
+                                            {
+                                                "topic": "topic",
+                                                "key": "key1",
+                                                "value": {
+                                                    "foo": "bar"
+                                                },
+                                                "partition": 0,
+                                                "offset": 2
+                                            },
+                                            {
+                                                "topic": "topic",
+                                                "key": "key2",
+                                                "value": [
+                                                    "foo2",
+                                                    "bar2"
+                                                ],
+                                                "partition": 1,
+                                                "offset": 3
+                                            }
+                                        ]
+                                    }
+                                }
+                            }
+                        }
                     },
                     "404": {
                         "description": "Consumer instance not found"
@@ -324,7 +363,34 @@
                 },
                 "responses": {
                     "200": {
-                        "description": "OK"
+                        "description": "OK",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/OffsetsWithoutTopic"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": {
+                                            "offsets": [
+                                                {
+                                                    "partition": 2,
+                                                    "offset": 0
+                                                },
+                                                {
+                                                    "partition": 1,
+                                                    "offset": 1
+                                                },
+                                                {
+                                                    "partition": 2,
+                                                    "offset": 2
+                                                }
+                                            ]
+                                        }
+                                    }
+                                }
+                            }
+                        }
                     },
                     "404": {
                         "description": "Topic not found"
@@ -362,7 +428,62 @@
                 },
                 "responses": {
                     "200": {
-                        "description": "OK"
+                        "description": "OK",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/Partitions"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": [
+                                            {
+                                                "partition": 1,
+                                                "leader": 1,
+                                                "replicas": [
+                                                    {
+                                                        "broker": 1,
+                                                        "leader": true,
+                                                        "in_sync": true
+                                                    },
+                                                    {
+                                                        "broker": 2,
+                                                        "leader": false,
+                                                        "in_sync": true
+                                                    },
+                                                    {
+                                                        "broker": 3,
+                                                        "leader": false,
+                                                        "in_sync": false
+                                                    }
+                                                ]
+                                            },
+                                            {
+                                                "partition": 2,
+                                                "leader": 2,
+                                                "replicas": [
+                                                    {
+                                                        "broker": 1,
+                                                        "leader": false,
+                                                        "in_sync": true
+                                                    },
+                                                    {
+                                                        "broker": 2,
+                                                        "leader": true,
+                                                        "in_sync": true
+                                                    },
+                                                    {
+                                                        "broker": 3,
+                                                        "leader": false,
+                                                        "in_sync": false
+                                                    }
+                                                ]
+                                            }
+                                        ]
+                                    }
+                                }
+                            }
+                        }
                     },
                     "404": {
                         "description": "Topic or Partition not found"
@@ -459,7 +580,7 @@
                     }
                 },
                 "responses": {
-                    "200": {
+                    "204": {
                         "description": "OK"
                     },
                     "404": {
@@ -716,6 +837,67 @@
                     }
                 },
                 "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
+            },
+            "createdConsumer": {
+                "title": "Root Type for createdConsumer",
+                "description": "The root of the createdConsumer type's schema.",
+                "type": "object",
+                "properties": {
+                    "instance_id": {
+                        "description": "Unique ID for the consumer instance in the group",
+                        "type": "string"
+                    },
+                    "base_uri": {
+                        "description": "Base URI used to construct URIs for subsequent requests against this consumer instance",
+                        "type": "string"
+                    }
+                },
+                "example": "{\n    \"instance_id\": \"consumer1\",\n    \"base_uri\": \"http://localhost:8080/consumers/my-group/instances/consumer1\"\n}"
+            },
+            "OffsetsWithoutTopic": {
+                "title": "Root Type for OffsetsWithoutTopic",
+                "description": "The root of the OffsetsWithoutTopic type's schema.",
+                "type": "object",
+                "properties": {
+                    "offsets": {
+                        "type": "array",
+                        "items": {}
+                    }
+                },
+                "example": "{\n    \"offsets\": [\n        {\n            \"partition\": 2,\n            \"offset\": 0\n        },\n        {\n            \"partition\": 1,\n            \"offset\": 1\n        },\n        {\n            \"partition\": 2,\n            \"offset\": 2\n        }\n    ]\n}"
+            },
+            "Records2": {
+                "title": "Root Type for Records2",
+                "description": "The root of the Records2 type's schema.",
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "topic": {
+                            "type": "string"
+                        },
+                        "key": {
+                            "type": "string"
+                        },
+                        "value": {
+                            "type": "object",
+                            "properties": {
+                                "foo": {
+                                    "type": "string"
+                                }
+                            }
+                        },
+                        "partition": {
+                            "format": "int32",
+                            "type": "integer"
+                        },
+                        "offset": {
+                            "format": "int32",
+                            "type": "integer"
+                        }
+                    }
+                },
+                "example": "[\n    {\n        \"topic\": \"topic\",\n        \"key\": \"key1\",\n        \"value\": {\n            \"foo\": \"bar\"\n        },\n        \"partition\": 0,\n        \"offset\": 2\n    },\n    {\n        \"topic\": \"topic\",\n        \"key\": \"key2\",\n        \"value\": [\n            \"foo2\",\n            \"bar2\"\n        ],\n        \"partition\": 1,\n        \"offset\": 3\n    }\n]"
             }
         },
         "requestBodies": {
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -13,6 +13,14 @@
         },
         "version": "1.0.0"
     },
+    "servers": [
+        {
+            "url": "https://bridge.swagger.io/v2"
+        },
+        {
+            "url": "http://bridge.swagger.io/v2"
+        }
+    ],
     "paths": {
         "/consumers/{groupid}": {
             "post": {
@@ -22,6 +30,7 @@
                 "summary": "Create a new consumer instance in the consumer group",
                 "operationId": "createConsumer",
                 "requestBody": {
+                    "description": "Consumer specification",
                     "content": {
                         "application/json": {
                             "schema": {
@@ -47,6 +56,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -74,6 +84,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -82,6 +93,7 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -112,6 +124,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -120,6 +133,7 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -150,6 +164,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -158,6 +173,7 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -188,6 +204,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -196,6 +213,7 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -229,6 +247,7 @@
                 "summary": "Subscribe to the given list of topics",
                 "operationId": "subscribe",
                 "requestBody": {
+                    "description": "A list of topics to subscribe",
                     "content": {
                         "application/json": {
                             "schema": {
@@ -269,6 +288,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -277,6 +297,7 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -314,6 +335,7 @@
                 {
                     "name": "topicname",
                     "in": "path",
+                    "description": "Name of the topic",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -354,6 +376,7 @@
                 {
                     "name": "topicname",
                     "in": "path",
+                    "description": "Name of the topic",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -362,6 +385,7 @@
                 {
                     "name": "partitionid",
                     "in": "path",
+                    "description": "ID of the partition",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -377,6 +401,7 @@
                 "summary": "Overrides the fetch offsets that the consumer will use for the next set of records to fetch",
                 "operationId": "seek",
                 "requestBody": {
+                    "description": "A list of offsets",
                     "content": {
                         "application/json": {
                             "schema": {
@@ -399,6 +424,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -407,6 +433,7 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -422,6 +449,7 @@
                 "summary": "Commit a list of offsets for the consumer",
                 "operationId": "commit",
                 "requestBody": {
+                    "description": "A list of offsets to commit for partitions",
                     "content": {
                         "application/json": {
                             "schema": {
@@ -443,6 +471,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
+                    "description": "The name of the consumer group",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -451,6 +480,58 @@
                 {
                     "name": "name",
                     "in": "path",
+                    "description": "The ID of the consumer instance",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/assignments": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "summary": "Manually assign a list of partitions to the consumer",
+                "operationId": "assign",
+                "requestBody": {
+                    "description": "A list of partitions to assign to the consumer",
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/PartitionsForPositions"
+                            }
+                        }
+                    },
+                    "required": true
+                },
+                "responses": {
+                    "204": {
+                        "description": "OK"
+                    },
+                    "404": {
+                        "description": "Consumer instance not found"
+                    },
+                    "409": {
+                        "description": "Subscription to topics, partitions and pattern are mutually exclusive"
+                    }
+                }
+            },
+            "parameters": [
+                {
+                    "name": "groupid",
+                    "in": "path",
+                    "description": "The name of the consumer group",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
+                },
+                {
+                    "name": "name",
+                    "in": "path",
+                    "description": "The ID of the consumer instance",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -459,41 +540,7 @@
             ]
         }
     },
-    "tags": [
-        {
-            "name": "Consumers",
-            "description": "Consumers management"
-        },
-        {
-            "name": "Topics",
-            "description": "Topics management"
-        }
-    ],
-    "externalDocs": {
-        "description": "Find out more about Swagger",
-        "url": "http://swagger.io"
-    },
-    "servers": [
-        {
-            "url": "https://bridge.swagger.io/v2"
-        },
-        {
-            "url": "http://bridge.swagger.io/v2"
-        }
-    ],
     "components": {
-        "requestBodies": {
-            "PartitionsForPositions": {
-                "content": {
-                    "application/json": {
-                        "schema": {
-                            "$ref": "#/components/schemas/PartitionsForPositions"
-                        }
-                    }
-                },
-                "required": true
-            }
-        },
         "schemas": {
             "ApiResponse": {
                 "type": "object",
@@ -524,7 +571,7 @@
                     "auto.offset.reset": {
                         "type": "string"
                     },
-                    "enable.auto.commit": {
+                    "auto.commit.enable": {
                         "type": "string"
                     },
                     "fetch.min.bytes": {
@@ -534,7 +581,7 @@
                         "type": "string"
                     }
                 },
-                "example": "{\n    \"name\": \"consumer1\",\n    \"format\": \"binary\",\n    \"auto.offset.reset\": \"earliest\",\n    \"enable.auto.commit\": \"false\",\n    \"fetch.min.bytes\": \"512\",\n    \"consumer.request.timeout.ms\": \"30000\"\n}"
+                "example": "{\n    \"name\": \"consumer1\",\n    \"format\": \"binary\",\n    \"auto.offset.reset\": \"earliest\",\n    \"auto.commit.enable\": \"false\",\n    \"fetch.min.bytes\": \"512\",\n    \"consumer.request.timeout.ms\": \"30000\"\n}"
             },
             "Offset": {
                 "title": "Offset",
@@ -670,6 +717,33 @@
                 },
                 "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
             }
+        },
+        "requestBodies": {
+            "PartitionsForPositions": {
+                "description": "A list of partitions",
+                "content": {
+                    "application/json": {
+                        "schema": {
+                            "$ref": "#/components/schemas/PartitionsForPositions"
+                        }
+                    }
+                },
+                "required": true
+            }
         }
+    },
+    "tags": [
+        {
+            "name": "Consumers",
+            "description": "Consumers management"
+        },
+        {
+            "name": "Topics",
+            "description": "Topics management"
+        }
+    ],
+    "externalDocs": {
+        "description": "Find out more about Swagger",
+        "url": "http://swagger.io"
     }
 }
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -623,7 +623,10 @@
                 "type": "object",
                 "properties": {
                     "offsets": {
-                        "$ref": "#/components/schemas/Offset"
+                        "type": "array",
+                        "items": {
+                            "$ref": "#/components/schemas/Offset"
+                        }
                     }
                 },
                 "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -13,14 +13,6 @@
         },
         "version": "1.0.0"
     },
-    "servers": [
-        {
-            "url": "https://bridge.swagger.io/v2"
-        },
-        {
-            "url": "http://bridge.swagger.io/v2"
-        }
-    ],
     "paths": {
         "/consumers/{groupid}": {
             "post": {
@@ -72,44 +64,6 @@
                 "responses": {
                     "204": {
                         "description": "No content"
-                    },
-                    "404": {
-                        "description": "Consumer instance not found"
-                    }
-                }
-            },
-            "parameters": [
-                {
-                    "name": "groupid",
-                    "in": "path",
-                    "required": true,
-                    "schema": {
-                        "type": "string"
-                    }
-                },
-                {
-                    "name": "name",
-                    "in": "path",
-                    "required": true,
-                    "schema": {
-                        "type": "string"
-                    }
-                }
-            ]
-        },
-        "/consumers/{groupid}/instances/{name}/offsets": {
-            "post": {
-                "tags": [
-                    "Consumers"
-                ],
-                "summary": "Commit a list of offsets for the consumer",
-                "operationId": "commit",
-                "requestBody": {
-                    "$ref": "#/components/requestBodies/Offsets"
-                },
-                "responses": {
-                    "200": {
-                        "description": "OK"
                     },
                     "404": {
                         "description": "Consumer instance not found"
@@ -423,7 +377,14 @@
                 "summary": "Overrides the fetch offsets that the consumer will use for the next set of records to fetch",
                 "operationId": "seek",
                 "requestBody": {
-                    "$ref": "#/components/requestBodies/Offsets"
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/Offsets"
+                            }
+                        }
+                    },
+                    "required": true
                 },
                 "responses": {
                     "204": {
@@ -452,9 +413,87 @@
                     }
                 }
             ]
+        },
+        "/consumers/{groupid}/instances/{name}/offsets": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "summary": "Commit a list of offsets for the consumer",
+                "operationId": "commit",
+                "requestBody": {
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/Offsets"
+                            }
+                        }
+                    }
+                },
+                "responses": {
+                    "200": {
+                        "description": "OK"
+                    },
+                    "404": {
+                        "description": "Consumer instance not found"
+                    }
+                }
+            },
+            "parameters": [
+                {
+                    "name": "groupid",
+                    "in": "path",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
+                },
+                {
+                    "name": "name",
+                    "in": "path",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
+                }
+            ]
         }
     },
+    "tags": [
+        {
+            "name": "Consumers",
+            "description": "Consumers management"
+        },
+        {
+            "name": "Topics",
+            "description": "Topics management"
+        }
+    ],
+    "externalDocs": {
+        "description": "Find out more about Swagger",
+        "url": "http://swagger.io"
+    },
+    "servers": [
+        {
+            "url": "https://bridge.swagger.io/v2"
+        },
+        {
+            "url": "http://bridge.swagger.io/v2"
+        }
+    ],
     "components": {
+        "requestBodies": {
+            "PartitionsForPositions": {
+                "content": {
+                    "application/json": {
+                        "schema": {
+                            "$ref": "#/components/schemas/PartitionsForPositions"
+                        }
+                    }
+                },
+                "required": true
+            }
+        },
         "schemas": {
             "ApiResponse": {
                 "type": "object",
@@ -631,42 +670,6 @@
                 },
                 "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
             }
-        },
-        "requestBodies": {
-            "Offsets": {
-                "content": {
-                    "application/json": {
-                        "schema": {
-                            "$ref": "#/components/schemas/Offsets"
-                        }
-                    }
-                },
-                "required": true
-            },
-            "PartitionsForPositions": {
-                "content": {
-                    "application/json": {
-                        "schema": {
-                            "$ref": "#/components/schemas/PartitionsForPositions"
-                        }
-                    }
-                },
-                "required": true
-            }
         }
-    },
-    "tags": [
-        {
-            "name": "Consumers",
-            "description": "Consumers management"
-        },
-        {
-            "name": "Topics",
-            "description": "Topics management"
-        }
-    ],
-    "externalDocs": {
-        "description": "Find out more about Swagger",
-        "url": "http://swagger.io"
     }
 }
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -524,7 +524,7 @@
                     "auto.offset.reset": {
                         "type": "string"
                     },
-                    "auto.commit.enable": {
+                    "enable.auto.commit": {
                         "type": "string"
                     },
                     "fetch.min.bytes": {
@@ -534,7 +534,7 @@
                         "type": "string"
                     }
                 },
-                "example": "{\n    \"name\": \"consumer1\",\n    \"format\": \"binary\",\n    \"auto.offset.reset\": \"earliest\",\n    \"auto.commit.enable\": \"false\",\n    \"fetch.min.bytes\": \"512\",\n    \"consumer.request.timeout.ms\": \"30000\"\n}"
+                "example": "{\n    \"name\": \"consumer1\",\n    \"format\": \"binary\",\n    \"auto.offset.reset\": \"earliest\",\n    \"enable.auto.commit\": \"false\",\n    \"fetch.min.bytes\": \"512\",\n    \"consumer.request.timeout.ms\": \"30000\"\n}"
             },
             "Offset": {
                 "title": "Offset",
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -1,8 +1,8 @@
 {
     "openapi": "3.0.0",
     "info": {
-        "title": "HTTP Kafka Bridge",
-        "description": "This is a HTTP Kafka Bridge server",
+        "title": "Strimzi HTTP Bridge for Apache Kafka",
+        "description": "The Strimzi HTTP Bridge for Apache Kafka provides a REST API for integrating client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol. The HTTP Bridge can be deployed on bare metal or in a container.",
         "termsOfService": "http://swagger.io/terms/",
         "contact": {
             "email": "strimzi@redhat.com"
@@ -27,10 +27,10 @@
                 "tags": [
                     "Consumers"
                 ],
-                "summary": "Create a new consumer instance in the consumer group",
+                "description": "Creates a consumer instance in a given consumer group. You can optionally specify a consumer name and supported configuration options.",
                 "operationId": "createConsumer",
                 "requestBody": {
-                    "description": "Consumer specification",
+                    "description": "Name and configuration of the consumer. If a name is not specified, a randomly generated name is assigned. All parameters are optional. The supported configuration options are shown in the following example.",
                     "content": {
                         "application/json": {
                             "schema": {
@@ -52,10 +52,10 @@
                         }
                     },
                     "409": {
-                        "description": "Consumer instance with the specified name already exists"
+                        "description": "A consumer instance with the specified name already exists in your Kafka cluster."
                     },
                     "422": {
-                        "description": "Invalid consumer configuration. One of the settings specified in the request contained an invalid value"
+                        "description": "One or more consumer configuration options have invalid values. Ensure that the values of the configuration options specified in the request body are valid."
                     }
                 }
             },
@@ -63,7 +63,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
-                    "description": "The name of the consumer group",
+                    "description": "ID of the consumer group in which to create the consumer.",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -76,14 +76,14 @@
                 "tags": [
                     "Consumers"
                 ],
-                "summary": "Destroy the consumer instance",
+                "description": "Deletes a consumer instance from your Kafka cluster.",
                 "operationId": "deleteConsumer",
                 "responses": {
                     "204": {
-                        "description": "No content"
+                        "description": "No content."
                     },
                     "404": {
-                        "description": "Consumer instance not found"
+                        "description": "The specified consumer instance was not found."
                     }
                 }
             },
@@ -91,7 +91,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
-                    "description": "The name of the consumer group",
+                    "description": "ID of the consumer group to which the consumer belongs.",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -100,7 +100,7 @@
                 {
                     "name": "name",
                     "in": "path",
-                    "description": "The ID of the consumer instance",
+                    "description": "Name of the consumer that you want to delete.",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -113,17 +113,25 @@
                 "tags": [
                     "Consumers"
                 ],
-                "summary": "Seek to the first offset for each of the given partitions",
+                "description": "Configures a subscribed consumer to seek to (and subsequently read from) the first offset in one or more given topic partitions.",
                 "operationId": "seekToBeginning",
                 "requestBody": {
-                    "$ref": "#/components/requestBodies/PartitionsForPositions"
+                    "description": "List of topic partitions to which the consumer is subscribed. The consumer will seek to the first offset in the specified partitions.",
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/PartitionsForPositions"
+                            }
+                        }
+                    },
+                    "required": true
                 },
                 "responses": {
                     "204": {
-                        "description": "No Content"
+                        "description": "No content."
                     },
                     "404": {
-                        "description": "Consumer instance not found"
+                        "description": "The specified consumer instance was not found."
                     }
                 }
             },
@@ -131,7 +139,7 @@
                 {
                     "name": "groupid",
                     "in": "path",
-                    "description": "The name of the consumer group",
+                    "description": "ID of the consumer group to which the subscribed consumer belongs.",
                     "required": true,
                     "schema": {
                         "type": "string"
@@ -140,7 +148,7 @@
                 {
                     "name": "name",
                     "in": "path",
-                    "description": "The ID of the consumer instance",
+                    "description": "Name of the subscribed consumer.",
                     "required": true,
                     "schema": {
                         "type": "string"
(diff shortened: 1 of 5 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -232,6 +232,7 @@ private void seek(RoutingContext routingContext) {
 
     private void seekToBeginning(RoutingContext routingContext) {
         //TODO
+
```
ppatierno: Why this space?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -1187,5 +1187,6 @@ public void offsetsConsumerDoesNotExist(TestContext context) {
                     context.assertEquals("Consumer instance not found", status);
                     commitAsync.complete();
                 });
+        commitAsync.await();
```
ppatierno: You know, it's not related to this PR :) please revert

tombentley:
Any progress on this?

tombentley:
Apparently this is awaiting review by @laidan6000.

laidan6000:
Sure thing. I'm reviewing the descriptions, etc. and will let you know when done, @stanlyDoge.

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    "type": "string"
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/offsets": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "summary": "Commit a list of offsets for the consumer",
```
laidan6000:
```suggestion
                "summary": "Commits a list of consumer offsets to the consumer offsets commit log. To commit offsets for all records fetched by the consumer, leave the request body empty.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                "operationId": "commit",
+                "parameters": [
+                    {
+                        "name": "body",
+                        "in": "body",
+                        "description": "A list of offsets to commit for partitions",
```
laidan6000:
```suggestion
                        "description": "List of consumer offsets to commit to the consumer offsets commit log. You can specify one or more topic partitions to commit offsets for.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                            "$ref": "#/definitions/Offsets"
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "OK"
+                    },
+                    "404": {
+                        "description": "Consumer instance not found"
```
laidan6000:
```suggestion
                        "description": "The specified consumer instance was not found."
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    }
+                }
+            },
+            "parameters": [
+                {
+                    "name": "groupid",
+                    "in": "path",
+                    "description": "The name of the consumer group",
```
laidan6000:
```suggestion
                    "description": "ID of the consumer group to which the consumer belongs.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    "required": true,
+                    "type": "string"
+                },
+                {
+                    "name": "name",
+                    "in": "path",
+                    "description": "The ID of the consumer instance",
```
laidan6000:
```suggestion
                    "description": "Name of the consumer.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    "type": "string"
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/assignments": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "summary": "Manually assign a list of partitions to the consumer",
```
laidan6000:
```suggestion
                "summary": "Assigns one or more topic partitions to a consumer.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                "operationId": "assign",
+                "parameters": [
+                    {
+                        "name": "body",
+                        "in": "body",
+                        "description": "A list of partitions to assign to the consumer",
```
laidan6000:
```suggestion
                        "description": "List of topic partitions to assign to the consumer.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                            "$ref": "#/definitions/PartitionsForPositions"
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "OK"
+                    },
+                    "404": {
+                        "description": "Consumer instance not found"
```
laidan6000:
```suggestion
                        "description": "The specified consumer instance was not found."
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    },
+                    "409": {
+                        "description": "Subscription to topics, partitions and pattern are mutually exclusive"
```
laidan6000:
```suggestion
                        "description": "Subscription to topics, partitions, and pattern are mutually exclusive."
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    }
+                }
+            },
+            "parameters": [
+                {
+                    "name": "groupid",
+                    "in": "path",
+                    "description": "The name of the consumer group",
```
laidan6000:
```suggestion
                    "description": "ID of the consumer group to which the consumer belongs.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    "required": true,
+                    "type": "string"
+                },
+                {
+                    "name": "name",
+                    "in": "path",
+                    "description": "The ID of the consumer instance",
```
laidan6000:
```suggestion
                    "description": "Name of the consumer to which you want to assign topic partitions.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,875 @@
...
+                    "required": true,
+                    "type": "string"
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/offsets": {
```
laidan6000: In my testing, the endpoint appeared to be named `offset` not `offsets`. Please check.

stanlyDoge:
After applying your suggestions the Apicurio complains about endpoints summary lenght (which should be 120 characters max ). Do you agree to move this summary to the description?

laidan6000:
Yes, I'm fine with moving the text from the "summary" field to the "description" field. The docs looks exactly the same when I preview the output in my browser.

laidan6000:
For the body of each POST request, is it possible to rename "Example Value" to "Request Body"?

Also, for the new example responses, is it possible to rename "Example Value" to "Example Response"?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
...
+                "type": {
+                    "type": "string"
+                },
+                "message": {
+                    "type": "string"
+                }
+            }
+        },
+        "Consumer": {
+            "title": "Root Type for Consumer",
```
PaulRMellor: Not sure of how we're capitalizing root type in this doc, but consumer has been lower case up to now. 
"Root type for the consumer"
This comment applies to all titles that include 'Root Type...'

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
...
+            "description": "The root of the Consumer type's schema.",
```
PaulRMellor: Not sure what this means. Is it 'The schema for the consumer root type'? 
This comment applies to all descriptions that begin: 'The root of the...'

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
...
+            },
+            "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
+        },
+        "createdConsumer": {
+            "title": "Root Type for createdConsumer",
+            "description": "The root of the createdConsumer type's schema.",
+            "type": "object",
+            "properties": {
+                "instance_id": {
+                    "description": "Unique ID for the consumer instance in the group",
```
PaulRMellor:
```suggestion
                    "description": "Unique ID for the consumer instance in the group.",
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                },
+                "example": "{\n    \"offsets\": [\n        {\n            \"topic\": \"topic\",\n            \"partition\": 0,\n            \"offset\": 15\n        },\n        {\n            \"topic\": \"topic\",\n            \"partition\": 1,\n            \"offset\": 42\n        }\n    ]\n}"
+            },
+            "createdConsumer": {
+                "title": "Root Type for createdConsumer",
+                "description": "The root of the createdConsumer type's schema.",
+                "type": "object",
+                "properties": {
+                    "instance_id": {
+                        "description": "Unique ID for the consumer instance in the group",
```
PaulRMellor:
```suggestion
                        "description": "Unique ID for the consumer instance in the group.",
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                        "type": "string"
+                    },
+                    "base_uri": {
+                        "description": "Base URI used to construct URIs for subsequent requests against this consumer instance",
```
PaulRMellor:
```suggestion
                        "description": "Base URI used to construct URIs for subsequent requests against this consumer instance.",
```

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
...
+                    "type": "string"
+                },
+                "base_uri": {
+                    "description": "Base URI used to construct URIs for subsequent requests against this consumer instance",
```
PaulRMellor:
```suggestion
                    "description": "Base URI used to construct URIs for subsequent requests against this consumer instance.",
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+        "/consumers/{groupid}/instances/{name}": {
+            "delete": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Deletes a consumer instance from your Kafka cluster.",
+                "operationId": "deleteConsumer",
+                "responses": {
+                    "204": {
+                        "description": "No content."
```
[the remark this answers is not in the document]
laidan6000: Good point. @stanlyDoge - is this possible?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}/positions/beginning": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Configures a subscribed consumer to seek to (and subsequently read from) the first offset in one or more given topic partitions.",
```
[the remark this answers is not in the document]
laidan6000: Looks good to me. @stanlyDoge - please accept.

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                "operationId": "seekToBeginning",
+                "requestBody": {
+                    "description": "List of topic partitions to which the consumer is subscribed. The consumer will seek to the first offset in the specified partitions.",
```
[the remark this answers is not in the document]
laidan6000: Looks good to me. @stanlyDoge - please accept.

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}/positions/end": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Configures a subscribed consumer to seek to (and subsequently read from) the last offset in one or more given topic partitions. This is the default behavior.",
```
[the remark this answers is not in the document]
laidan6000: Looks good to me.

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                "operationId": "seekToEnd",
+                "requestBody": {
+                    "description": "List of topic partitions to which the consumer is subscribed. The consumer will seek to the last offset in the specified partitions.",
```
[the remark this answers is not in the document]
laidan6000: Looks good to me.

src/main/resources/openapiv2.json
laidan6000: Good point. @stanlyDoge - Should the "Models" section even be visible in the API documentation?

src/main/resources/openapiv2.json
laidan6000: Looks good to me.

src/main/resources/openapi.json
laidan6000: Looks good to me.

src/main/resources/openapi.json
laidan6000: Looks good to me.

src/main/resources/openapiv2.json
laidan6000: Looks good to me.

src/main/resources/openapiv2.json
stanlyDoge: These titles are autogenerated. I can change them, but I think it is necessary to have models in this spec file.

src/main/resources/openapi.json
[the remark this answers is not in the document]
stanlyDoge: We can put there more detailed message but I am not sure about this one. @ppatierno what do you think?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}/records": {
+            "get": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Retrieves information about the records processed by a subscribed consumer, including message values, topics, and partitions. Uses the `subscribe` or `assign` API.",
```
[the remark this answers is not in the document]
stanlyDoge: By using `assign` we can assign a specific partition to the consumer. With the `subscribe`, the partitions are assigned dynamically. I am not sure whether this info is "TLDR". What do you think @ppatierno ?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    "404": {
+                        "description": "The specified consumer instance was not found."
+                    },
+                    "406": {
+                        "description": "The format of the consumer does not match the embedded format requested in the Accept header."
```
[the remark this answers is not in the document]
stanlyDoge: When we send HTTP request, it should contain for example `Accept: application/json` format specification. The produced response should have the same format.

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
...
+                                    ]
+                                }
+                            ]
+                        }
+                    },
+                    "404": {
+                        "description": "The specified topic partition was not found."
+                    },
+                    "422": {
+                        "description": "Unprocessable Entity"
```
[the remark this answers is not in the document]
stanlyDoge: I do not think we can customize the message for different endpoints. WDYT @ppatierno ?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
+{
+    "swagger": "2.0",
+    "info": {
+        "title": "Strimzi HTTP Bridge for Apache Kafka",
+        "description": "The Strimzi HTTP Bridge for Apache Kafka provides a REST API for integrating client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol. The HTTP Bridge can be deployed on bare metal or in a container.",
```
ppatierno: "... interacting HTTP based client applications ..."
" ... or in a container running on Kubernetes or OpenShift platform"

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,876 @@
...
+        "license": {
+            "name": "Apache 2.0",
+            "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+        },
+        "version": "1.0.0"
+    },
+    "host": "bridge.swagger.io",
+    "basePath": "/v2",
+    "schemes": [
+        "https",
```
ppatierno: for now we don't support HTTPS

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                },
+                "responses": {
+                    "204": {
+                        "description": "No content."
+                    },
+                    "404": {
+                        "description": "The specified consumer instance was not found."
+                    },
+                    "409": {
+                        "description": "Subscription to topics, partitions, and pattern are mutually exclusive."
```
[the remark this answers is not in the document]
tombentley:
```suggestion
                        "description": "Subscription to topics, partitions, and patterns are mutually exclusive."
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                    }
+                }
+            ]
+        },
+        "/topics/{topicname}": {
+            "post": {
+                "tags": [
+                    "Topics"
+                ],
+                "description": "Sends one or more records to a given topic, optionally specifying a partition, key, or both.",
```
tombentley: "optionally specifying a partition" that's only available via a different endpoint AFAICS.

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+        },
+        "/consumers/{groupid}/instances/{name}/offsets": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Commits a list of consumer offsets to the consumer offsets commit log. To commit offsets for all records fetched by the consumer, leave the request body empty.",
+                "operationId": "commit",
+                "requestBody": {
+                    "description": "List of consumer offsets to commit to the consumer offsets commit log. You can specify one or more topic partitions to commit offsets for.",
```
tombentley: If an empty body means "commit offsets for all records fetched by the consumer" then does a partial list of consumer offsets mean "commit just these offsets for these partitions, and don't commit anything for the other assigned partitions", or does it mean "commit these offsets for these partitions and for the others commit the offsets fetched by the consumer"?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                    }
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/assignments": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Assigns one or more topic partitions to a consumer.",
```
tombentley: If multiple calls are made are the effects additive, or do subsequent calls have the side-effect of unsubscribing previously assigned partitions?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                },
+                "responses": {
+                    "204": {
+                        "description": "OK"
+                    },
+                    "404": {
+                        "description": "The specified consumer instance was not found."
+                    },
+                    "409": {
+                        "description": "Subscription to topics, partitions, and pattern are mutually exclusive."
```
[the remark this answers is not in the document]
tombentley:
```suggestion
                        "description": "Subscription to topics, partitions, and patterns are mutually exclusive."
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+            "Consumer": {
+                "title": "Root Type for Consumer",
+                "description": "The root of the Consumer type's schema.",
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string"
+                    },
+                    "format": {
+                        "type": "string"
```
tombentley: Can you document the valid values for this and other properties?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                    },
+                    "auto.offset.reset": {
+                        "type": "string"
+                    },
+                    "auto.commit.enable": {
+                        "type": "string"
+                    },
+                    "fetch.min.bytes": {
+                        "type": "string"
```
tombentley: surely this is an integer?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                    },
+                    "format": {
+                        "type": "string"
+                    },
+                    "auto.offset.reset": {
+                        "type": "string"
+                    },
+                    "auto.commit.enable": {
+                        "type": "string"
```
tombentley: surely this is a boolean?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string"
+                    },
+                    "format": {
+                        "type": "string"
+                    },
+                    "auto.offset.reset": {
+                        "type": "string"
```
tombentley: boolean?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                    },
+                    "auto.commit.enable": {
+                        "type": "string"
+                    },
+                    "fetch.min.bytes": {
+                        "type": "string"
+                    },
+                    "consumer.request.timeout.ms": {
+                        "type": "string"
```
tombentley: integer/long?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                    "partition": {
+                        "format": "int32",
+                        "type": "integer"
+                    },
+                    "leader": {
+                        "format": "int32",
+                        "type": "integer"
+                    },
+                    "replicas": {
+                        "type": "object"
```
tombentley: Can this be constrained further than "object"? And is it really an object or an array of objects?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,935 @@
...
+                "type": "object",
+                "properties": {
+                    "offsets": {
+                        "type": "array",
+                        "items": {}
+                    }
+                },
+                "example": "{\n    \"offsets\": [\n        {\n            \"partition\": 2,\n            \"offset\": 0\n        },\n        {\n            \"partition\": 1,\n            \"offset\": 1\n        },\n        {\n            \"partition\": 2,\n            \"offset\": 2\n        }\n    ]\n}"
+            },
+            "Records2": {
```
tombentley: Why do we need two?

tombentley:
One other general comment: We're using the term records here, but the rest of Strimzi uses the term messages. We should be consistent.

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "description": "Name of the subscribed consumer.",
+                    "required": true,
+                    "type": "string"
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/positions/end": {
+            "post": {
+                "tags": [
+                    "Consumers"
```
ppatierno: "Seek" tag

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        "description": "List of topic partitions to which the consumer is subscribed. The consumer will seek the last offset in the specified partitions.",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/PartitionsForPositions"
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "No content."
```
ppatierno: "Seek operation done successfully" @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "type": "string"
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/records": {
+            "get": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Retrieves information about the records processed by a subscribed consumer, including message values, topics, and partitions. Uses the `subscribe` or `assign` API.",
```
ppatierno: It's not about retrieving information about records ... it's about retrieving the records/messages from the topic partitions on which the consumer is subscribed to. @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                "operationId": "poll",
+                "produces": [
+                    "application/json"
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
```
ppatierno: "Poll request executed successfully" ? @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                                    "partition": 1,
+                                    "offset": 3
+                                }
+                            ]
+                        }
+                    },
+                    "404": {
+                        "description": "The specified consumer instance was not found."
+                    },
+                    "406": {
```
ppatierno: We don't have this one for now.

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "name": "groupid",
+                    "in": "path",
+                    "description": "ID of the consumer group to which the subscribed consumer belongs.",
+                    "required": true,
+                    "type": "string"
+                },
+                {
+                    "name": "name",
+                    "in": "path",
+                    "description": "Name of the subscribed consumer for which you want to retrieve information on processed records.",
```
ppatierno: they are no information but actual records/messages @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "required": true,
+                    "type": "string"
+                },
+                {
+                    "name": "timeout",
+                    "in": "query",
+                    "description": "The maximum amount of time in ms that the HTTP Bridge spends retrieving records before timing out the request.",
```
ppatierno: should we use "milliseconds" instead of just "ms"? @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        "description": "List of topics to which the consumer will subscribe.",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/Topics"
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "No content."
```
ppatierno: "Consumer subscribed successfully" @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+            },
+            "delete": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Unsubscribes a consumer from all topics.",
+                "operationId": "unsubscribe",
+                "responses": {
+                    "204": {
+                        "description": "No content."
```
ppatierno: "Consumer unsubscribed successfully" @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        "in": "body",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/Records"
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
```
ppatierno: "Messages/records sent successfully" ? @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "description": "Name of the consumer that you want to unsubscribe from topics.",
+                    "required": true,
+                    "type": "string"
+                }
+            ]
+        },
+        "/topics/{topicname}": {
+            "post": {
+                "tags": [
+                    "Topics"
```
ppatierno: Adding "Producer" tag?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "description": "Name of the topic to which you want to send records.",
+                    "required": true,
+                    "type": "string"
+                }
+            ]
+        },
+        "/topics/{topicname}/partitions/{partitionid}": {
+            "post": {
+                "tags": [
+                    "Topics"
```
ppatierno: "Producer" tag?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        "description": "List of records to send to a given topic partition, including a value (required) and a key (optional).",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/RecordsWithoutPartitions"
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
```
ppatierno: "Messages/records sent successfully" ? @PaulRMellor @laidan6000

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        "schema": {
+                            "$ref": "#/definitions/Partitions"
```
ppatierno: are you sure about that? we should have records here

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        },
+                        "examples": {
+                            "application/json": [
```
ppatierno: I think this example is wrong as the schema ref

src/main/resources/openapiv2.json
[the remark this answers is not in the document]
ppatierno: I think that this description is not something that goes into the HTTP response. I think it's just a description about the meaning of the response code so we can have something different for each request. It just describes what 422 means for this endpoint. Or you read differently from openAPI specification?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "description": "ID of the partition to which you want to send records.",
+                    "required": true,
+                    "type": "string"
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/positions": {
+            "post": {
+                "tags": [
+                    "Consumers"
```
ppatierno: Add "Seek" tag

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                ],
+                "description": "Configures a subscribed consumer to fetch offsets from a particular offset ID the next time it fetches a set of records from a given topic partition. This overrides the default fetch behavior for consumers. You can specify one or more topic partitions.",
```
ppatierno: just "offset" not "offset ID"

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                ],
+                "parameters": [
+                    {
+                        "name": "body",
+                        "in": "body",
+                        "description": "List of partition offsets from which the subscribed consumer will next fetch records.",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/Offsets"
+                        }
```
ppatierno: what about an example here?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "No content."
```
ppatierno: "Seek done successfully" ? @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "application/json"
+                ],
+                "parameters": [
+                    {
+                        "name": "body",
+                        "in": "body",
+                        "description": "List of consumer offsets to commit to the consumer offsets commit log. You can specify one or more topic partitions to commit offsets for.",
+                        "required": false,
+                        "schema": {
+                            "$ref": "#/definitions/Offsets"
```
ppatierno: what about an example here?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "OK"
```
ppatierno: "Commit done successfully"? @laidan6000 @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                    "application/json"
+                ],
+                "parameters": [
+                    {
+                        "name": "body",
+                        "in": "body",
+                        "description": "List of topic partitions to assign to the consumer.",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/PartitionsForPositions"
```
ppatierno: an example here?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        }
+                    }
+                ],
+                "responses": {
+                    "204": {
+                        "description": "OK"
```
ppatierno: "Partitions assigned successfully" ? @laidan6000 @PaulRMellor

src/main/resources/openapi.json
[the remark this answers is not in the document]
ppatierno: I answered about the "description" field in another answer. I guess it's just about to describe the response but it's not going to be part of the HTTP response.

src/main/resources/openapi.json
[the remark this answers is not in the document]
ppatierno: I don't think we need to add "Uses the `subscribe` or `assign` API.". What information it's trying to provide to the user?

src/main/resources/openapiv2.json
PaulRMellor: First offset found successfully?

src/main/resources/openapiv2.json
PaulRMellor: Commit made successfully?

src/main/resources/openapi.json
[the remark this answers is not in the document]
ppatierno: I don't think we will have this support for the 1.2.
The values in the records will be just strings. I am thinking if it's something we have to address for binary data or not.
So right now you could send binary data in base64 encoded format but the bridge doesn't decode to binary but just saves the base64 string so the consumer should be aware of that on the other side.
The final point is to have the bridge decoding the base64 and writing the binary data to the kafka topic.

src/main/resources/openapi.json
ppatierno: It just uses the native `assign` in Kafka consumer and the official doc says "This interface does not allow for incremental assignment and will replace the previous assignment (if there is one)"

src/main/resources/openapiv2.json
ppatierno: It's not about searching offset but it's about seeking to the offset.

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                "fetch.min.bytes": {
+                    "type": "string"
+                },
+                "consumer.request.timeout.ms": {
+                    "type": "string"
+                }
+            },
+            "example": "{\n    \"name\": \"consumer1\",\n    \"format\": \"binary\",\n    \"auto.offset.reset\": \"earliest\",\n    \"auto.commit.enable\": \"false\",\n    \"fetch.min.bytes\": \"512\",\n    \"consumer.request.timeout.ms\": \"30000\"\n}"
+        },
+        "createdConsumer": {
```
ppatierno: please capital "C" so "CreatedConsumer"

src/main/resources/openapi.json
stanlyDoge: `Records` contains nested `records` array. The `Records` contains a list of records. I know it is not ideal but I wanted to be accurate.

src/main/resources/openapi.json
ppatierno: I don't think we need this custom type at all, it's about a GET operations on topics endpoint which we don't support.

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "$ref": "#/definitions/createdConsumer"
+                        },
+                        "examples": {}
```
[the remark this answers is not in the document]
stanlyDoge: Take a look into `createdConsumer`. The example is here.

src/main/resources/openapi.json
tombentley: I don't really think calling one of these things `Records2` ireally improves accuracy. Would it make sense to rename one of them `RecordList` or something like that, if one of them is really just a holder for an array of records?

ppatierno: Why? `Records` defines an array type containing elements of type `Record`. Why we need `Records2`?

src/main/resources/openapiv2.json
```diff
@@ -0,0 +1,889 @@
...
+                        "description": "Name and configuration of the consumer. If a name is not specified, a randomly generated name is assigned. All parameters are optional. The supported configuration options are shown in the following example.",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/Consumer"
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
```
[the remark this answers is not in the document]
stanlyDoge: That mean we need to change this message also in the bridge code.

src/main/resources/openapiv2.json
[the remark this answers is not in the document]
ppatierno: But as you have for GET records I think the example should be here as well.

src/main/resources/openapiv2.json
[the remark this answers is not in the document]
ppatierno: No my understanding is that it's not the message that we send as body in the HTTP response. It's just a description for people trying to understand what 200 means reading the specification. Let's ask @tombentley as well. Wdyt?

src/main/resources/openapiv2.json
PaulRMellor: interacting with HTTP-based client applications

src/main/resources/openapiv2.json
PaulRMellor: Then 'Seek performed successfully'?

src/main/resources/openapi.json
stanlyDoge: @ppatierno WDYT?

src/main/resources/openapi.json
stanlyDoge: It is used as data type of array in `Partitions`.

src/main/resources/openapiv2.json
stanlyDoge: The examples are taken from the data type definition. Check the rendered documentation.

src/main/resources/openapi.json
stanlyDoge: @laidan6000 @PaulRMellor can you create some descriptions there, please?

PaulRMellor: @stanlyDoge - picking up and will add directly in apicurio

src/main/resources/openapi.json
ppatierno: And we don't need `Partitions` as well

ppatierno:
About "records" (`Record`, `RecordList` and `Records`) I took a look and we are talking about produced records and consumed records.
So `Record` is a produced record data type with just key, value and partition ... and `Records` is a list of these records sent by the producer.
The `RecordsList` is an array of consumed records with topic, key, value, partition and offset.
I would try to following this name and semantic, is it possible?

tombentley:
Naming them `ProducerRecord`, `ProducerRecordList` (`ProducerRecords` is just soo easily confused with the singular version) and `ConsumerRecordList` would imho be much clearer.

ppatierno:
Exactly

ppatierno:
@tombentley regarding your question about commit offsets, if the request is empty it just does a commit for all the partitions related to received records (so native Kafka client means `commitSync` without parameter) if only some of them are specified, then it commits just for those partitions (so `commitSync` with a collection of TopicPartition and related OffsetAndMetadata)

tombentley:
@ppatierno I wasn't asking for my own edification: Tell it to the docs!

ppatierno:
@tombentley sorry it sounded as a question to me! @laidan6000 @PaulRMellor ^^

src/main/resources/openapiv2.json
```diff
@@ -302,13 +274,31 @@
                 ],
```
ppatierno: The subscription endpoint not only allow a list ot topics (described by `Topics` type) but also topic_pattern field. They are mutually exclusive, something that we should write in the description as well. @stanlyDoge @PaulRMellor

src/main/resources/openapiv2.json
```diff
@@ -652,6 +686,103 @@
                     "type": "string"
                 }
             ]
+        },
+        "/consumers/{groupid}/instances/{name}/records": {
+            "get": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Retrieves records for subscribed consumer, including message values, topics, and partitions.",
```
ppatierno: even in this case we have to hightlight that the request must be sent to the Kafka bridge instance where the consumer instance lives.

src/main/resources/openapiv2.json
```diff
@@ -671,75 +802,130 @@
...
-            "title": "Record",
-            "required": [
-                "value"
-            ],
-            "type": "object",
-            "properties": {
-                "partition": {
-                    "type": "integer"
+                "format": {
+                    "description": "The allowable message format for the consumer, which can be binary (default), avro or json. The messages are converted into a JSON format. ",
```
ppatierno: Remove `avro` we don't have it for now.

src/main/resources/openapiv2.json
```diff
@@ -671,75 +802,130 @@
...
+                "format": "binary",
+                "auto.offset.reset": "earliest",
+                "auto.commit.enable": "false",
+                "fetch.min.bytes": "512",
+                "consumer.request.timeout.ms": "30000"
             }
         },
-        "Topics": {
-            "title": "Topics",
+        "Offsets": {
```
ppatierno: it's used on commit and on seek operations so I would call OffsetsToCommitSeek

src/main/resources/openapiv2.json
```diff
@@ -671,75 +802,130 @@
...
+                        "topic": "topic",
+                        "partition": 1,
+                        "offset": 42
+                    }
+                ]
+            }
         },
-        "Records": {
-            "title": "Records",
+        "OffsetsWithoutTopic": {
```
ppatierno: It's used in the HTTP response related to records sent, so I would call it OffsetsRecordsSent

src/main/resources/openapiv2.json
```diff
@@ -671,75 +802,130 @@
...
+                        "offset": 1
+                    },
+                    {
+                        "partition": 2,
+                        "offset": 2
                     }
+                ]
+            }
+        },
+        "Partition": {
```
ppatierno: I don't think we use this one, it contains information about administration things like who is the leader etc etc. What am I missing?

src/main/resources/openapiv2.json
```diff
@@ -671,75 +802,130 @@
...
+                    {
+                        "broker": 2,
+                        "leader": false,
+                        "in_sync": false
+                    }
+                ]
+            }
         },
-        "Partitions": {
+        "PartitionsForPositions": {
```
ppatierno: it's not just for positions endpoint but used on seek as well right? Let's call it just `Partitions` if the previous one is really not used in our case.

src/main/resources/openapiv2.json
```diff
@@ -750,47 +936,58 @@
...
+                    {
+                        "topic": "topic",
+                        "partition": 1
+                    }
+                ]
+            }
         },
-        "Offsets": {
-            "title": "Offsets",
+        "RecordsWithoutPartitions": {
```
ppatierno: Are we using it? I cannot find it in the rest of the json file.

src/main/resources/openapiv2.json
```diff
@@ -859,27 +1035,111 @@
...
+                        "value": "msg2",
+                        "partition": 1
+                    },
+                    {
+                        "value": "msg3"
+                    }
+                ]
+            }
+        },
+        "ErrorBody": {
```
ppatierno: Just `Error` is better

src/main/resources/openapiv2.json
```diff
@@ -859,27 +1035,111 @@
...
+            "title": "Root Type for ErrorBody",
+            "description": "The root of the ErrorBody type's schema.",
```
ppatierno: `Error`

src/main/resources/openapiv2.json
```diff
@@ -859,27 +1035,111 @@
...
+            "description": "The `topics` endpoints allow you to send messages to a specified topic or topic partition. You can include message keys in requests."
+        },
+        {
+            "name": "Seek",
+            "description": "`Seek` related operations enable a consumer to begin receiving messages from a given offset position."
+        },
+        {
+            "name": "Producer",
+            "description": "The `producer` endpoints allow you to send records to a specified topic or topic partition."
         }
```
ppatierno: Regarding all the tags descriptions I would use the same form of the "Seek" one.
The tags are a way of grouping things. So the "Seek" description says clearly "... related operations ..." so it describe that it's grouping operation related to seek.
I would use the same for the other tags. Saying "The consumers endpoints ..." for example seems to me that we are describing an endpoint and not a tag. @PaulRMellor help on this please.

src/main/resources/openapiv2.json
stanlyDoge: `/topics/{topicname}/partitions/{partitionid}` uses it as response body.

src/main/resources/openapiv2.json
```diff
@@ -81,14 +103,23 @@
                 "tags": [
                     "Consumers"
                 ],
-                "description": "Deletes a consumer instance from your Kafka cluster.",
+                "description": "Deletes a consumer instance from your Kafka Bridge instance.",
```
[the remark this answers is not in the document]
PaulRMellor: Deletes a specified consumer instance from the Kafka Bridge instance that includes it.

src/main/resources/openapiv2.json
ppatierno: right! Clearly I have bad search skills on github :-) Btw I don't see any difference from ProducerRecordList even because you are defining the "records" field as a ProducerRecordList so why another type?

src/main/resources/openapiv2.json
```diff
@@ -159,6 +200,7 @@
         "/consumers/{groupid}/instances/{name}/positions/end": {
             "post": {
                 "tags": [
+                    "Seek",
                     "Consumers"
                 ],
                 "description": "Configures a subscribed consumer to seek (and subsequently read from) the last offset in one or more given topic partitions. This is the default behavior.",
```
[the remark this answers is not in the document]
PaulRMellor: Configures a subscribed consumer to seek (and subsequently read from) the offset at the end of one or more given topic partitions. This is the default behavior.

src/main/resources/openapiv2.json
PaulRMellor: Subscribes a consumer to one or more topics. You can describe the list of topics to which the consumer will subscribe in a list (of `Topics` type) or as a `topic_pattern` field.

src/main/resources/openapiv2.json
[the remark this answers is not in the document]
ppatierno: @PaulRMellor I would be more explicit. The request must to be send etc etc ... because in the way you are using it seems to me that I can send the DELETE request to any bridge instance and it will take care to delete the consumer instance even on another bridge instance (something like an internal communication between bridges).

src/main/resources/openapiv2.json
PaulRMellor: Retrieves records for a subscribed consumer from the Kafka Bridge instance that holds it, including message values, topics, and partitions.

src/main/resources/openapiv2.json
PaulRMellor: Consumer operations to create consumers in your Kafka cluster and perform common actions, such as subscribing to topics, retrieving processed records, and committing offsets.

Producer operations to send records to a specified topic or topic partition.

Seek operations that enable a consumer to begin receiving messages from a given offset position.

Topic operations to send messages to a specified topic or topic partition. You can include message keys in requests.

laidan6000:
I suggest combining the **Topics** and **Producers** sections into one because they describe the same family of operations and the same resource. Personally, I would name it "Topics" but I don't mind either way. I think having two sections may confuse users.

ppatierno:
@laidan6000 that's a good point but "Topics" it too generic because it could be applied to consumers operatin as well, not just producers. I would stick with both for now because we are going to have even admin operations in the future that will be related to topics.

ppatierno:
@stanlyDoge I approved the PR but you have to fix the conflicts. Just override whatever is in the master because your openapi.json is now the source of truth

laidan6000:
Please fix the typo in the GET request: ""the the Kafka Bridge instance".

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
-          "name": "topicname",
-          "in": "path",
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+    "openapi": "3.0.0",
+    "info": {
+        "title": "Strimzi HTTP Bridge for Apache Kafka",
+        "description": "The Strimzi HTTP Bridge for Apache Kafka provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol. The HTTP Bridge can be deployed on bare metal or in a container running on Kubernetes or OpenShift platform.",
```
tombentley: How the bridge might be deployed is completely orthogonal to the API it provides, which is what this json is describing. So I suggest we drop the final sentence.

```suggestion
        "description": "The Strimzi HTTP Bridge for Apache Kafka provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol.",
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
-          "schema": {
-            "type": "integer"
-          }
-        }
-      ]
+        "license": {
+            "name": "Apache 2.0",
+            "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
+        },
+        "version": "1.0.0"
```
tombentley: Do we consider this API to be as version 1.0.0? We're defined _all_ the endpoints and we're sure they're all correct, even though this will be the initial release?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
-            "type": "string"
-          }
+    ],
+    "paths": {
+        "/consumers/{groupid}": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Creates a consumer instance in a given consumer group. You can optionally specify a consumer name and supported configuration options.",
```
```suggestion
                "description": "Creates a consumer instance in the given consumer group. You can optionally specify a consumer name and supported configuration options.",
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/Error"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": {
+                                            "error_code": 422,
+                                            "message": "One or more consumer configuration options have invalid values."
```
tombentley: Does the message really not include the names of the problematic option(s)? How do we expect people to debug their code?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}": {
+            "delete": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Deletes a specified consumer instance. The operation MUST be performed on the the Kafka Bridge instance that contains the consumer.",
```
```suggestion
                "description": "Deletes a specified consumer instance. The operation MUST be performed on the Kafka Bridge instance that contains the consumer.",
```

"The operation MUST be performed on the Kafka Bridge instance that contains the consumer." I'm not sure users would really understand what that's supposed to mean. What you're trying to say is that if multiple bridge instances are deployed behind a load balancer then a delete request must be routed to the same instance as the preceeding create request. So why not just say that?

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
+                    }
+                },
+                {
+                    "name": "name",
+                    "in": "path",
+                    "description": "Name of the consumer that you want to delete.",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
```
tombentley: So if the delete request requires both the name and the group id I assume that means that consumer names are unique within the scope of a consumer id, rather than within the scope of the whole bridge. You should probably say that in the docs for the create endpoint.

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}/positions/beginning": {
+            "post": {
+                "tags": [
+                    "Seek",
+                    "Consumers"
+                ],
+                "description": "Configures a subscribed consumer to seek (and subsequently read from) the first offset in one or more given topic partitions.",
```
```suggestion
                "description": "Configures a subscribed consumer to seek (and subsequently read from) the first offset in one or more of the given topic partitions.",
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
+                            "schema": {
+                                "$ref": "#/components/schemas/Partitions"
+                            }
+                        }
+                    },
+                    "required": true
+                },
+                "responses": {
+                    "204": {
+                        "description": "Seek to the beggining performed successfully."
```
```suggestion
                        "description": "Seek to the begining performed successfully."
```

src/main/resources/openapi.json
```diff
@@ -1,286 +1,1279 @@
...
+                    },
+                    "404": {
```
tombentley: What is the error if the consumer instance exists, but is not subscribed to one of the given topic partitions?

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}/positions/end": {
+            "post": {
+                "tags": [
+                    "Seek",
+                    "Consumers"
+                ],
+                "description": "Configures a subscribed consumer to seek (and subsequently read from) the offset at the end of one or more given topic partitions. This is the default behavior.",
```
```suggestion
                "description": "Configures a subscribed consumer to seek (and subsequently read from) the offset at the end of one or more of the given topic partitions. This is the default behavior.",
```
"This is the default behavior." I don't think that will make sense to a reader. Its meaning isn't apparent to me.

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                            }
+                        }
+                    },
+                    "required": true
+                },
+                "responses": {
+                    "204": {
+                        "description": "Seek to the end performed successfully."
+                    },
+                    "404": {
```
tombentley: Same comment about non-subscribed topic partitions

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}/subscription": {
+            "post": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Subscribes a consumer to one or more topics. You can describe the topics to which the consumer will subscribe in a list (of `Topics` type) or as a `topic_pattern` field.",
```
tombentley: Explain that each call _replaces_ the subscribed for the consumer.

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/Error"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": {
+                                            "error_code": 422,
+                                            "message": "The record is not valid."
```
tombentley: The parameter is a topic list, so "the topic" doesn't make sense.
```suggestion
                                            "message": "The record list contains invalid records."
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                                            "error_code": 404,
+                                            "message": "The specified topic was not found."
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    "422": {
+                        "description": "The record is not valid.",
```
```suggestion
                        "description": "The record list is not valid.",
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                                            "error_code": 404,
+                                            "message": "The specified topic partition was not found."
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    "422": {
+                        "description": "The record is not valid.",
```
```suggestion
                        "description": "The record list is not valid.",
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/Error"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": {
+                                            "error_code": 422,
+                                            "message": "The record is not valid."
```
```suggestion
                                            "message": "The record list is not valid."
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                            }
+                        }
+                    },
+                    "required": true
+                },
+                "responses": {
+                    "204": {
+                        "description": "Seek performed successfully."
+                    },
+                    "404": {
```
tombentley: What if the offsets includes topics/partitions to which the consumer is not subscribed?

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                    }
+                }
+            ]
+        },
+        "/consumers/{groupid}/instances/{name}/records": {
+            "get": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Retrieves records for a subscribed consumer, including message values, topics, and partitions. The operation MUST be performed on the the Kafka Bridge instance that contains the consumer.",
```
tombentley: Same comment about that last sentence: I don't think readers will necessarily understand what you mean.

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                                            "error_code": 404,
+                                            "message": "The specified consumer instance was not found."
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    },
+                    "406": {
+                        "description": "The format of the consumer does not match the embedded format requested in the Accept header.",
```
tombentley: It's not immediately obvious what the "format of the consumer" means, and in particular that it was defined by a previous request. Maybe this is a little clearer:
```suggestion
                        "description": "The `format` used in the consumer creation request does not match the embedded format in the Accept header of this request.",
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/Error"
+                                },
+                                "examples": {
+                                    "response": {
+                                        "value": {
+                                            "error_code": 406,
+                                            "message": "The format of the consumer does not match the embedded format requested in the Accept header."
```
```suggestion
                                            "message": "The `format` used in the consumer creation request does not match the embedded format in the Accept header of this request."
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                    "description": "Name of the subscribed consumer for which you want to retrieve records.",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
+                },
+                {
+                    "name": "timeout",
+                    "in": "query",
+                    "description": "The maximum amount of time in milliseconds that the HTTP Bridge spends retrieving records before timing out the request.",
```
```suggestion
                    "description": "The maximum amount of time, in milliseconds, that the HTTP Bridge spends retrieving records before timing out the request.",
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "type": "string"
+                    }
+                }
+            },
+            "Consumer": {
+                "title": "Consumer",
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "description": "The unique name for the consumer instance. The name is used in URLs. ",
```
tombentley: It's unique within the scope of a consume group, right? Not within the whole bridge?

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "type": "string"
+                    },
+                    "format": {
+                        "description": "The allowable message format for the consumer, which can be binary (default), avro or json. The messages are converted into a JSON format. ",
```
tombentley: We don't support avro. And imho you should use quotes ("binary") or monospace (if that's possible) if those are the names people should use in the request.

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "type": "string"
+                    },
+                    "auto.offset.reset": {
+                        "description": "Resets the offset position for the consumer. \nIf set to earliest, messages are read from the first offset. \nIf set to latest, messages are read from the latest offset.",
```
tombentley: Quotes

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "type": "string"
+                    },
+                    "auto.commit.enable": {
+                        "description": "If set to true, message offsets are committed automatically for the consumer. If set to false, message offsets must be committed manually.",
```
tombentley: Quotes

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
+                        "type": "string"
+                    },
+                    "fetch.min.bytes": {
+                        "description": "Sets the minimum ammount of data in bytes for the consumer to receive. The broker waits until the data to send exceeds this amount.",
+                        "type": "string"
+                    },
+                    "consumer.request.timeout.ms": {
+                        "description": "Sets the maximum amount of time in milliseconds (ms) for the consumer to wait for messages for a request. If the timeout period is reached without a response, an error is returned.",
```
```suggestion
                        "description": "Sets the maximum amount of time, in milliseconds, for the consumer to wait for messages for a request. If the timeout period is reached without a response, an error is returned.",
```

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
-            "type": "string"
-          }
+            "name": "Producer",
+            "description": "Producer operations to send records to a specified topic or topic partition."
         }
-      ]
+    ],
+    "externalDocs": {
+        "description": "Find out more about Swagger",
+        "url": "http://swagger.io"
```
tombentley: Is this right, or should it point to Strimzi's docs for the bridge?

src/main/resources/openapi.json
```diff
@@ -1,314 +1,1279 @@
...
-          "required": true,
-          "schema": {
-            "type": "string"
-          }
+        "/consumers/{groupid}/instances/{name}": {
+            "delete": {
+                "tags": [
+                    "Consumers"
+                ],
+                "description": "Deletes a specified consumer instance. The operation MUST be performed on the Kafka Bridge instance that contains the consumer.",
```
```suggestion
                "description": "Deletes a specified consumer instance. The operation MUST be performed on the Kafka Bridge instance that contains the consumer.",
```

"The operation MUST be performed on the Kafka Bridge instance that contains the consumer." I'm not sure users would really understand what that's supposed to mean. What you're trying to say is that if multiple bridge instances are deployed behind a load balancer then a delete request must be routed to the same instance as the preceeding create request. So why not just say that? Do you want to mention the `base_uri` which is otherwise mentioned in the docs for the type but the use of which is not really explained anywhere.

src/main/resources/openapi.json
ppatierno: Good point I would say "0.1.0"

src/main/resources/openapi.json
ppatierno: It's just an example, we will have different messages in order to give more details to the user.

src/main/resources/openapi.json
ppatierno: Tbh the sentence is pretty clear to me. Btw your point about load balancer is ok. So we could use "routed" instead of "performed" :-)

src/main/resources/openapi.json
ppatierno: ?

src/main/resources/openapi.json
ppatierno: ?

src/main/resources/openapi.json
ppatierno: yeah it should be a link to the Strimzi website where we have doc just for the bridge.

src/main/resources/openapi.json
tombentley:
```suggestion
                        "description": "If set to 'true', message offsets are committed automatically for the consumer. If set to 'false', message offsets must be committed manually.",
```

src/main/resources/openapi.json
tombentley:
```suggestion
                        "description": "Resets the offset position for the consumer. \nIf set to 'earliest', messages are read from the first offset. \nIf set to 'latest', messages are read from the latest offset.",
```

src/main/resources/openapi.json
PaulRMellor: Deletes a specified consumer instance. The operation MUST be routed to the Kafka Bridge instance that contains the consumer.

src/main/resources/openapi.json
PaulRMellor: The name is unique within the scope of the consumer group.

src/main/resources/openapi.json
PaulRMellor: I removed the default sentence.

src/main/resources/openapi.json
PaulRMellor: Don't we want the same message as the previous 422: The record list contains invalid records?

tombentley: Yes, consistent. Thanks @PaulRMellor.

src/main/resources/openapi.json
PaulRMellor: Retrieves records for a subscribed consumer, including message values, topics, and partitions. The operation MUST be routed to the Kafka Bridge instance that contains the consumer.

src/main/resources/openapi.json
PaulRMellor: The unique name for the consumer instance. The name is unique within the scope of the consumer group. The name is used in URLs.

src/main/resources/openapi.json
PaulRMellor: ...which can be `binary` (default) or `json`.

src/main/resources/openapi.json
PaulRMellor: externalDocs:
    description: Find out more about the Strimzi HTTP Bridge
    url: 'https://strimzi.io/documentation/'

src/main/resources/openapi.json
ppatierno: I am going to open an issue for that. We should test this scenario. fyi @PaulRMellor

src/main/resources/openapi.json
ppatierno: I think that we should explain what `base_uri` is on the `createConsumer` operation where we have just an example of the response. @PaulRMellor

src/main/resources/openapi.json
ppatierno: I opened #195

src/main/resources/openapi.json
ppatierno: I opened #195

src/main/resources/openapi.json
ppatierno: Opened #195

ppatierno:
@tombentley I think that @PaulRMellor addressed all the comments. Related to the seek errors, I opened an issue (#195) as replied to your questions. I think we can address it later because I want to test what happens at Kafka native client level. The last thing is about describing what the `base_uri` is. Do you think you have more or after that addition we can merge this one?

PR #167, opened 2019-05-13
Handling sticky sessions

stanlyDoge:
https://github.com/strimzi/strimzi-kafka-bridge/pull/167


stanlyDoge:
According to [this](https://kubernetes.io/docs/concepts/services-networking/service/) the stickiness is time-limited. The max time for the sticky session is one day. Do we need to look for another solution?

Issue #170, opened 2019-05-14
Move timeout on HTTP GET /records as query parameter


ppatierno:
Currently, the `timeout` for the HTTP GET on the `/records` endpoint is specified in the HTTP header. We have to move it as query parameter so, for example:

`GET /consumers/my-group/instances/my-consumer/records?timeout=5000`

PR #176, opened 2019-05-15
Move timeout from header to query parameters

stanlyDoge:
https://github.com/strimzi/strimzi-kafka-bridge/issues/170

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
@@ -500,8 +500,7 @@ public class HttpBridgeTest extends KafkaClusterTestBase {
         // consume records
         Async consumeAsync = context.async();
 
-        client.get(BRIDGE_PORT, BRIDGE_HOST, baseUri + "/records")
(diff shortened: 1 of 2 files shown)


src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -500,8 +500,7 @@ public void receiveSimpleMessage(TestContext context) {
         // consume records
```
ppatierno: I would not change every single test in sending with the timeout addition even because we have to test what happens when this param is not specify. I would add a new test for this use case. I know there is a lot of common code in the tests and we have an issue for that in order to factor out common methods.

stanlyDoge: I just changed putting it to the header to putting it into the query. Are you proposing removing timeout from all the tests?

ppatierno:
Timeout wasn't in all the tests right?

stanlyDoge:
It was in all the tests containing `records` endpoint.

ppatierno:
Damn you are right! :)

Issue #171, opened 2019-05-14
Add max_bytes on HTTP GET /records as query parameter


ppatierno:
Add the maximum number of bytes of unencoded keys and values that should be included in the response.

PR #177, opened 2019-05-15
Add max_bytes parameter on /records endpoint

stanlyDoge:
https://github.com/strimzi/strimzi-kafka-bridge/issues/171

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -147,13 +147,13 @@ public class HttpSinkBridgeEndpoint<V, K> extends SinkBridgeEndpoint<V, K> {
                     if (records.succeeded()) {
                         Buffer buffer = (Buffer) messageConverter.toMessages(records.result());
                         if (buffer.getBytes().length > this.maxBytes) {
-                            routingContext.response().setStatusMessage("Response is too large").setStatusCode(ErrorCodeEnum.UNPROCESSABLE_ENTITY.getValue()).end();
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -126,10 +126,18 @@ public void handle(Endpoint<?> endpoint) {
...
+                if (routingContext.request().getParam("max_bytes") != null) {
+                    this.maxBytes = Long.parseLong(routingContext.request().getParam("max_bytes"));
+                }
+
                 this.consume(records -> {
                     if (records.succeeded()) {
                         Buffer buffer = (Buffer) messageConverter.toMessages(records.result());
-                        sendConsumerRecordsResponse(routingContext.response(), buffer);
+                        if (buffer.getBytes().length > this.maxBytes) {
+                            routingContext.response().setStatusMessage("Response is too large").setStatusCode(ErrorCodeEnum.UNPROCESSABLE_ENTITY.getValue()).end();
```
ppatierno: We should try to put the sending logic in the sendConsumerRecordsResponse refactoring it if needed with more parameters. I noticed that we should do that for other use cases.

stanlyDoge: But that is not matter of this PR, right?

ppatierno:
Well in this case you need a way to reply to the client and it is just a little refactoring for using a method. But if you prefer we can do with another PR. Whatever you want :)

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
stanlyDoge: Ok, I've done it. It was just a little change. I thought the refactoring would influence more of codebase.

PR #178, opened 2019-05-15
Added support for assignment on multiple partitions

### LINKED ISSUE #124: Support for partitions assignment on a dedicate HTTP endpoint
Currently, if the client wants to be assigned a specific partition (so assignment against subscription) it has to do the same HTTP request as subscribe but specifying partition and offset in the JSON payload.
Because assignment has a quite different meaning in Kafka, it's better having a specific endpoint for that, for example doing an HTTP POST on:

```
/consumers/<group_id>/instances/<name>/assignments
```

The JSON payload could contain a list of partitions with topic and related partition to be assigned.

Of course, error handling should be done properly with related HTTP error status codes 4xx/5xx

ppatierno:
This PR fixes #124

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -20,6 +20,7 @@ import io.strimzi.kafka.bridge.config.BridgeConfig;
 import io.strimzi.kafka.bridge.config.KafkaConfig;
 import io.strimzi.kafka.bridge.tracker.OffsetTracker;
 import io.vertx.core.AsyncResult;
+import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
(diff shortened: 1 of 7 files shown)


src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -225,61 +226,89 @@ protected void assign(boolean shouldAttachHandler) {
...
-                        if (seekResult.failed()) {
+                    this.consumer.assign(Collections.singleton(topicPartition), assignResult -> {
+
+                        this.handleAssign(assignResult);
+                        if (assignResult.failed()) {
                             return;
                         }
-                        partitionsAssigned(Collections.singleton(topicPartition));
+
+                        log.debug("Assigned to {} partition {} on topic {}", topicSubscription.getTopic(), topicSubscription.getPartition());
```
stanlyDoge: I think the topic and partition are switched here. Also there are three "{}", what is the third parameter to print?

Issue #119, opened 2019-03-13
Support HTTP DELETE consumer for unsubscribe topics 


ppatierno:
Currently, it's not possible for a consumer to unsubscribe to the subscribed topics.
It could be implemented as an HTTP DELETE on the URL:

```
/consumers/<group_id>/instances/<name>/subscription
```

We should also handle errors in the proper way.

PR #179, opened 2019-05-20
Support topic unsubscription

stanlyDoge:
https://github.com/strimzi/strimzi-kafka-bridge/issues/119

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -174,10 +174,8 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
 
     /**
      * Unubscribe all the topics which the consumer currently subscribes
-     *
-     * @param shouldAttachHandler if the handler for getting messages should be set up
      */
-    protected void unsubscribe(boolean shouldAttachHandler) {
+    protected void unsubscribe() {
 
         log.info("Unsubscribe from topics {}", this.topicSubscriptions);
         topicSubscriptions.clear();
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -67,12 +67,15 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
     protected long maxBytes = Long.MAX_VALUE;
 
     private boolean shouldAttachSubscriberHandler;
+    private boolean shouldAttachUnsubscriberHandler;
 
     // handlers called when partitions are revoked/assigned on rebalancing
     private Handler<Set<TopicPartition>> partitionsRevokedHandler;
     private Handler<Set<TopicPartition>> partitionsAssignedHandler;
     // handler called after a topic subscription request
     private Handler<AsyncResult<Void>> subscribeHandler;
+    // handler called after an unsubscription request
+    private Handler<AsyncResult<Void>> unsubscribeHandler;
     // handler called after a request for a specific partition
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -78,12 +78,15 @@
     protected long maxBytes = Long.MAX_VALUE;
 
     private boolean shouldAttachSubscriberHandler;
+    private boolean shouldAttachUnsubscriberHandler;
```
ppatierno: this doesn't make sense. The `shouldAttachSubscriberHandler` (which maybe we can rename to `shoudAttacheRecordsHandler` but not do that in this PR!) is there just to avoid or not attaching the handler on the Kafka consumer for getting messages automatically. This is the behavior we have with AMQP but for HTTP we avoid to attach the handler because the client asks for more messages sending more HTTP poll requests. On unsubscribe we don't need something like this.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -188,6 +188,17 @@ public void handle(Endpoint<?> endpoint) {
                     sendConsumerCommitOffsetResponse(routingContext.response(), status.succeeded());
                 });
                 break;
+
+            case UNSUBSCRIBE:
+
+                this.setUnsubscribeHandler(unsubscribeResult -> {
+                    if (unsubscribeResult.succeeded()) {
+                        sendConsumerUnubscriptionResponse(routingContext.response());
+                    }
```
tombentley: So if there's an error we don't send any response?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -78,12 +78,15 @@
...
 
     // handlers called when partitions are revoked/assigned on rebalancing
     private Handler<Set<TopicPartition>> partitionsRevokedHandler;
     private Handler<Set<TopicPartition>> partitionsAssignedHandler;
     // handler called after a topic subscription request
     private Handler<AsyncResult<Void>> subscribeHandler;
+    // handler called after an unsubscription request
+    private Handler<AsyncResult<Void>> unsubscribeHandler;
```
tombentley: To be honest all the changes in this class seem quite convoluted. The reason I think that is likely because I'm new to this code, but I don't really understand why a field is needed at all. AFAIK we only call unsubscribe on the consumer when the bridge receives an unsubscribe API call, so why is a field needed to store a reference to the unsubscribe handler after the response has been returned?

ppatierno: You are absolutely right if we think just about HTTP bridge. This class is common to the AMQP layer and in that case, I needed fields in the past (maybe I could be wrong today). We are just following the same path. I should check how unsubscription works in AMQP.
Btw we could do this kind of things later because it doesn't hurt the bridge from a functional perspective. Right?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -181,6 +183,18 @@ protected void subscribe(boolean shouldAttachHandler) {
         this.consumer.subscribe(topics, this::subscribeHandler);
     }
 
+    /**
+     * Unubscribe all the topics which the consumer currently subscribes
+     *
+     * @param shouldAttachHandler if the handler for getting messages should be set up
+     */
+    protected void unsubscribe(boolean shouldAttachHandler) {
```
ppatierno: so the `shouldAttachHandler` is useless, right? remove from the Javadoc as well.

stanlyDoge: Good catch.

Issue #180, opened 2019-05-20
Factor out a common method for sending HTTP response


ppatierno:
The `HttpSinkBridgeEndpoint` is polluting of `sendXXX` methods for sending the HTTP response to the client.
It would be better having just one method with parameters (errcode, message, body, ...).

Issue #169, opened 2019-05-13
Factor out common actions in the HTTP bridge tests


ppatierno:
The HTTP bridge tests have a lot of common code about, for example, consumer creation, subscription and so on.
It would be useful to factor out methods for those actions to use in multiple tests.

PR #182, opened 2019-05-21
Adding support for seek operations

### LINKED ISSUE #123: Support for consumer seeking to begin/end/position of a partition
It should be possible for an HTTP client to seek at specific position in a partition begin/end/position.
An HTTP POST could be made to the current endpoint:

```
/consumers/<group_id>/instances/<name>/positions
```
providing an offsets list containing topic, partition and offset.
For seeking at begin/end the /beginning and /end could be added to the above URL and then providing a partitions list with topic and partition.

Of course, error on seeking should be handled properly with HTTP status codes 4xx/5xx.

ppatierno:
This PR fixes #123

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -652,4 +652,28 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
             }
         });
     }
+
+    protected void seek(TopicPartition topicPartition, long offset, Handler<AsyncResult<Void>> seekHandler) {
+        this.consumer.seek(topicPartition, offset, result -> {
+            if (seekHandler != null) {
+                seekHandler.handle(result);
+            }
+        });
+    }
+
+    protected void seekToBeginning(Set<TopicPartition> topicPartitionSet, Handler<AsyncResult<Void>> seekHandler) {
+        this.consumer.seekToBeginning(topicPartitionSet, result -> {
+            if (seekHandler != null) {
+                seekHandler.handle(result);
+            }
+        });
+    }
+
+    protected void seekToEnd(Set<TopicPartition> topicPartitionSet, Handler<AsyncResult<Void>> seekHandler) {
+        this.consumer.seekToEnd(topicPartitionSet, result -> {
+            if (seekHandler != null) {
+                seekHandler.handle(result);
+            }
+        });
+    }
 }
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -106,53 +106,59 @@ public class HttpSinkBridgeEndpoint<V, K> extends SinkBridgeEndpoint<V, K> {
                 break;
 
             case SEEK:
-
-                JsonArray seekOffsetsList = bodyAsJson.getJsonArray("offsets");
-
-                List<Future> seekHandlers = new ArrayList<>(seekOffsetsList.size());
-                for (int i = 0; i < seekOffsetsList.size(); i++) {
-                    TopicPartition topicPartition = new TopicPartition(seekOffsetsList.getJsonObject(i));
-                    long offset = seekOffsetsList.getJsonObject(i).getLong("offset");
-                    Future<Void> fut = Future.future();
-                    seekHandlers.add(fut);
-                    this.seek(topicPartition, offset, fut.completer());
-                }
-
-                CompositeFuture.join(seekHandlers).setHandler(done -> {
-                    if (done.succeeded()) {
-                        sendSeekResponse(routingContext.response(), ErrorCodeEnum.NO_CONTENT);
-                    } else {
-                        sendSeekResponse(routingContext.response(), ErrorCodeEnum.INTERNAL_SERVER_ERROR);
-                    }
-                });
+                doSeek(bodyAsJson);
                 break;
 
             case SEEK_TO_BEGINNING:
             case SEEK_TO_END:
-
-                JsonArray seekPartitionsList = bodyAsJson.getJsonArray("partitions");
-
-                Set<TopicPartition> set = seekPartitionsList.stream()
-                        .map(JsonObject.class::cast)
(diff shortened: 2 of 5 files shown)


src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -652,4 +652,28 @@ protected void commit(Map<TopicPartition, io.vertx.kafka.client.consumer.OffsetA
             }
         });
     }
+
+    protected void seek(TopicPartition topicPartition, long offset, Handler<AsyncResult<Void>> seekHandler) {
+        this.consumer.seek(topicPartition, offset, result -> {
+            if (seekHandler != null) {
```
tombentley: Why would this method be called with a null `seekHandler`? Unlike the other methods in `SinkBridgeEndpoint` the parameter is not used to set a field.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -63,7 +69,7 @@ public void open() {
     }
 
     @Override
-    @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:NPathComplexity"})
+    @SuppressWarnings({"checkstyle:CyclomaticComplexity", "checkstyle:NPathComplexity", "checkstyle:MethodLength"})
```
tombentley: Surely the correct response to this warning is not to add a suppression but to refactor the method? And I don't think that refactoring is difficult: The body of each case statement would appear to be easily extracted as a method.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -188,6 +194,53 @@ public void handle(Endpoint<?> endpoint) {
...
+
+            case SEEK_TO_BEGINNING:
+            case SEEK_TO_END:
+
+                JsonArray seekPartitionsList = bodyAsJson.getJsonArray("partitions");
+
+                Set<TopicPartition> set = seekPartitionsList.stream()
+                        .map(JsonObject.class::cast)
+                        .map(json -> new TopicPartition(json.getString("topic"), json.getInteger("partition")))
+                        .collect(Collectors.toSet());
```
tombentley: So the API contract on duplicates is last-in-wins?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -188,6 +194,53 @@ public void handle(Endpoint<?> endpoint) {
...
+
+                Handler<AsyncResult<Void>> seekHandler = done -> {
```
tombentley: This is the same handler as for `SEEK` above, so why not factor out a common class for it?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -1636,4 +1639,400 @@ public void doNotRespondTooLongMessage(TestContext context) {
...
+
+        String name = "my-kafka-consumer";
+        String groupId = "my-group";
+
+        String baseUri = "http://" + BRIDGE_HOST + ":" + BRIDGE_PORT + "/consumers/" + groupId + "/instances/" + name;
+
+        JsonObject json = new JsonObject();
+        json.put("name", name);
+
+        client.post(BRIDGE_PORT, BRIDGE_HOST, "/consumers/" + groupId)
```
tombentley: You have all these repeated blocks of code for setting up the bridge so you can test it, _surely_ you can factor out a method for each HTTP request you're making?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: Totally agree. I was just going for the fast way but I will make this change which was already in my mind.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
ppatierno: We already have #169 for this :-)

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: common class or common method?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: For seek beginning/end you are not specifying an offset but just topic + partition on which you want to seek at beginning or end (remember that the beginning and end endpoint are different).
So if you write the same topic + partition in the JSON you are just duplicating the same information, there is no last-in-wins. Or I miss understood the question?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
tombentley: A common method. (It will result in a common anonymous class anyway)

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
tombentley: Excellent point sir!

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: Why an anonymous class, I don't see what you mean. I can extract a method with `done` as a param.

ppatierno: I have just factored out seek(s) methods for now. Related to the comment on seek handler, I have an issue about refactoring around sending responses to the client. I will revisit this point while addressing that issue (#180)

PR #194, opened 2019-05-24
Fix HTTP sink endpoint removing on connection close

### LINKED ISSUE #189: Wrong map key is used to remove HTTP sink endpoint on close
If for any reason, the close happens on the sink and the related handler is called, the wrong key is used to remove the HTTP sink from the related collection.

```java
sink.closeHandler(s -> {
            httpBridgeContext.getHttpSinkEndpoints().remove(sink);
        });
```

The handler or the sink itself should expose somehow the consumer id which is used as key for that collection.

ppatierno:
This PR fixes #189 

Added sink and source name
Fixed bug on removing HTTP sink endpoint (based on name/consumer-id)

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -304,7 +304,7 @@ public class HttpSinkBridgeEndpoint<K, V> extends SinkBridgeEndpoint<K, V> {
 
                 JsonObject json = bodyAsJson;
                 // if no name, a random one is assigned
-                this.name = json.getString("name", "kafka-bridge-consumer-" + UUID.randomUUID().toString());
+                this.name = json.getString("name", "kafka-bridge-consumer-" + UUID.randomUUID());
 
                 if (this.httpBridgeContext.getHttpSinkEndpoints().containsKey(this.name)) {
                     routingContext.response().setStatusMessage("Consumer instance with the specified name already exists.")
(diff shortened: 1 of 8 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -304,10 +304,9 @@ public void handle(Endpoint<?> endpoint, Handler<?> handler) {
 
                 JsonObject json = bodyAsJson;
                 // if no name, a random one is assigned
-                String consumerInstanceId = json.getString("name",
-                        "kafka-bridge-consumer-" + UUID.randomUUID().toString());
+                this.name = json.getString("name", "kafka-bridge-consumer-" + UUID.randomUUID().toString());
```
tombentley: The string concat will invoke toString anyway

```suggestion
                this.name = json.getString("name", "kafka-bridge-consumer-" + UUID.randomUUID());
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -41,6 +42,12 @@ public HttpSourceBridgeEndpoint(Vertx vertx, HttpBridgeConfig bridgeConfigProper
         this.httpBridgeContext = httpBridgeContext;
     }
 
+    @Override
+    public void open() {
+        this.name = "kafka-bridge-producer-" + UUID.randomUUID().toString();
```
tombentley:
```suggestion
        this.name = "kafka-bridge-producer-" + UUID.randomUUID();
```

PR #201, opened 2019-05-27
Use properties file

stanlyDoge:


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +33,73 @@
...
+
+        ConfigRetriever retriever = ConfigRetriever.create(vertx, options);
+        retriever.getConfig(ar -> {
+            AmqpConfig amqpConfig = new AmqpConfig(ar.result().getBoolean("amqp.enabled", AmqpConfig.DEFAULT_AMQP_ENABLED),
+                    AmqpMode.from(ar.result().getString("amqp.mode", AmqpConfig.DEFAULT_AMQP_MODE)),
+                    ar.result().getInteger("amqp.flowCredit", AmqpConfig.DEFAULT_FLOW_CREDIT),
+                    ar.result().getString("amqp.host", AmqpConfig.DEFAULT_HOST),
+                    ar.result().getInteger("amqp.port", AmqpConfig.DEFAULT_PORT),
+                    ar.result().getString("amqp.messageConverter", AmqpConfig.DEFAULT_MESSAGE_CONVERTER),
+                    ar.result().getString("amqp.certDir", AmqpConfig.DEFAULT_CERT_DIR));
```
ppatierno: If we are going to remove the env var, maybe you could avoid this hard-coded properties and use the related fields in `AmqpConfig` class which today are env vars but changing to these properties.
I.e. 

`private static final String AMQP_ENABLED = "AMQP_ENABLED";`

in

`public static final String AMQP_ENABLED = "amqp.enabled";`

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +33,73 @@
...
+
+            HttpConfig httpConfig = new HttpConfig(ar.result().getBoolean("http.enabled", HttpConfig.DEFAULT_HTTP_ENABLED),
+                    ar.result().getString("http.host", HttpConfig.DEFAULT_HOST),
+                    ar.result().getInteger("http.port", HttpConfig.DEFAULT_PORT));
```
ppatierno: ditto as above

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +33,73 @@
...
+
+            KafkaConsumerConfig kafkaConsumerConfig = new KafkaConsumerConfig(ar.result().getString("kafka.consumer.autoOffsetReset", KafkaConsumerConfig.DEFAULT_AUTO_OFFSET_RESET));
+            KafkaProducerConfig kafkaProducerConfig = new KafkaProducerConfig(Integer.toString(ar.result().getInteger("kafka.producer.acks", Integer.parseInt(KafkaProducerConfig.DEFAULT_ACKS))));
+
+            KafkaConfig kafkaConfig = new KafkaConfig(ar.result().getString("kafka.bootstrapServers", KafkaConfig.DEFAULT_BOOTSTRAP_SERVERS),
+                    kafkaProducerConfig, kafkaConsumerConfig);
```
ppatierno: ditto as above

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +33,73 @@
...
+    private static String getFilePath(String arg) {
+        String[] confFileArg = arg.split("=");
+        String path = "";
+        if (confFileArg[0].equals("--config-file")) {
+            if (confFileArg[1].startsWith("/")) {
+                // absolute path
+                path = confFileArg[1];
+            } else {
+                // relative path
+                path = System.getProperty("user.dir") + "/" + confFileArg[1];
```
scholzj: Should be Windows friendly (use `File.separator`).

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +33,73 @@
...
-                    log.debug("HTTP verticle instance deployed [{}]", done.result());
-                } else {
-                    log.debug("Failed to deploy HTTP verticle instance", done.cause());
-                }
-            });
+    private static String getFilePath(String arg) {
+        String[] confFileArg = arg.split("=");
+        String path = "";
+        if (confFileArg[0].equals("--config-file")) {
+            if (confFileArg[1].startsWith("/")) {
```
scholzj: Should be Windows friendly. Why do you actually need to distinguish between the absolute and relative path?

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +33,73 @@
...
-            HttpBridge httpBridge = new HttpBridge(httpBridgeConfig);
-            vertx.deployVerticle(httpBridge, done -> {
-                if (done.succeeded()) {
-                    log.debug("HTTP verticle instance deployed [{}]", done.result());
-                } else {
-                    log.debug("Failed to deploy HTTP verticle instance", done.cause());
-                }
-            });
+    private static String getFilePath(String arg) {
+        String[] confFileArg = arg.split("=");
```
scholzj: It would be good if we can handle both `--config-file=xxx` and `--config-file xxx`

src/main/java/io/strimzi/kafka/bridge/amqp/AmqpConfig.java
```diff
@@ -20,13 +20,13 @@
     private static final String AMQP_MESSAGE_CONVERTER = "AMQP_MESSAGE_CONVERTER";
```
scholzj: Do we still need these? Are these the env var definitions?

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
```diff
@@ -16,9 +16,9 @@
     private static final String HTTP_HOST = "HTTP_HOST";
```
scholzj: Do we still need these? Are these the env var definitions?

src/main/java/io/strimzi/kafka/bridge/amqp/AmqpConfig.java
ppatierno: We can use it as "amqp.message.converter" instead of the hard-coded one and making it public.

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
ppatierno: As above, "http.host"

src/main/java/io/strimzi/kafka/bridge/Application.java
stanlyDoge: It was just my intention to make it more robust :(

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
stanlyDoge: These are now renamed (to priperties-like format) used as key for getting these properties.

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -23,28 +35,77 @@
...
+        String path = "";
+        if (confFileArg[0].equals("--config-file")) {
+            if (confFileArg[1].startsWith(File.separator)) {
```
scholzj: I guess that this will not work ... absolute path on Windows will start with something like `N:\`.

PR #204, opened 2019-05-28
Updated OpenAPI with latest descriptions fixes

ppatierno:


src/main/resources/openapi.json
```diff
@@ -24,7 +24,7 @@
                 "tags": [
                     "Consumers"
                 ],
-                "description": "Creates a consumer instance in the given consumer group. You can optionally specify a consumer name and supported configuration options. It resturns a base URI used to construct URIs for subsequent requests against this consumer instance.",
+                "description": "Creates a consumer instance in the given consumer group. You can optionally specify a consumer name and supported configuration options. It returns a base URI used to construct URIs for subsequent requests against this consumer instance.",
```
```suggestion
                "description": "Creates a consumer instance in the given consumer group. You can optionally specify a consumer name and supported configuration options. It returns a base URI which should be used to construct URIs for subsequent requests against this consumer instance.",
```

src/main/resources/openapi.json
```diff
@@ -599,7 +599,7 @@
                 "tags": [
                     "Consumers"
                 ],
-                "description": "Retrieves records for a subscribed consumer, including message values, topics, and partitions. The operation MUST be routed to the Kafka Bridge instance that contains the consumer.",
+                "description": "Retrieves records for a subscribed consumer, including message values, topics, and partitions. The operation MUST be routed through the base URI of the consumer to the Kafka Bridge instance that contains the consumer.",
```
tombentley: I don't think readers would understand what "routed through the base URI of the consumer" is supposed to mean. 
In any case it would be more accurate to call it a URL.

I won't duplicate this suggestion everywhere, because I expect some to-and-fro over the exact form of words, but I think something like this is a lot less vague:

```suggestion
                "description": "Retrieves records for a subscribed consumer, including message values, topics, and partitions. The request for this operation MUST use the base URL (including the host and port) returned in the response from the `POST` request to `/consumers/{groupid}` which was used to create this consumer."
```

src/main/resources/openapi.json
PaulRMellor: Should be or must be?

tombentley: "must", since (depending on how it's deployed) it might not work otherwise.

src/main/resources/openapi.json
PaulRMellor: Suggestion seems clear to me. I have updated the two descriptions that include the original.

PR #206, opened 2019-05-28
Updated properties logic

stanlyDoge:


.checkstyle/suppressions.xml
```diff
@@ -8,4 +8,7 @@
 
     <!-- Note that [/\\] must be used as the path separator for cross-platform support -->
 
+    <suppress checks="NPathComplexity"
+              files="io[/\\]strimzi[/\\]kafka[/\\]bridge[/\\]Application.java"/>
+
```
ppatierno: why did you need that?

scripts/kafka_bridge_config_generator.sh
```diff
@@ -0,0 +1,83 @@
...
+        SASL_MECHANISM="SCRAM-SHA-512"
+        JAAS_SECURITY_MODULE="scram.ScramLoginModule"
+    fi
+
+    SASL_AUTH_CONFIGURATION=$(cat <<EOF
+sasl.mechanism=${SASL_MECHANISM}
+sasl.jaas.config=org.apache.kafka.common.security.${JAAS_SECURITY_MODULE} required username="${KAFKA_BRIDGE_SASL_USERNAME}" password="${PASSWORD}";
+EOF
+)
+fi
```
ppatierno: all the above configuration are needed for both producer and consumer and they have to have `kafka.consumer.` and `kafka.producer.` prefix.
Comparing with Kafka Mirror Maker configuration, there we have two different config generator scripts.

scripts/kafka_bridge_config_generator.sh
```diff
@@ -0,0 +1,83 @@
...
+# Write the config file
+cat <<EOF
+#Apache Kafka common
+kafka.bootstrapServers=${KAFKA_BRIDGE_BOOTSTRAP_SERVERS}
+
+#Apache Kafka producer
+kafka.producer.acks=${KAFKA_BRIDGE_PRODUCER_ACKS}
+
+#Apache Kafka consumer
+kafka.consumer.autoOffsetReset=${KAFKA_BRIDGE_CONSUMER_AUTO_OFFSET_RESET}
```
ppatierno: we don't need all the three above configuration parameter, we should get configuration for consumer and producer from the CRD where `consumer.config` and `producer.config` are now needed. @scholzj wdyt?

scripts/kafka_bridge_run.sh
```diff
@@ -0,0 +1,58 @@
...
+./kafka_bridge_config_generator.sh | tee /tmp/kafka-bridge.properties
+echo ""
+
+
+# Disable Kafka's GC logging (which logs to a file)...
+export GC_LOG_ENABLED="false"
+
+if [ -z "$KAFKA_LOG4J_OPTS" ]; then
+export KAFKA_LOG4J_OPTS="-Dlog4j.configuration=file:$KAFKA_HOME/custom-config/log4j.properties"
+fi
```
ppatierno: we don't have Kafka here ... pay attention at copy/paste ... it's an art ;)

scripts/kafka_bridge_run.sh
```diff
@@ -0,0 +1,58 @@
...
+    # available to a jvm using 100% of the GCroup-aware memory
+    # up to some optional DYNAMIC_HEAP_MAX
+    CALC_MAX_HEAP=$(get_heap_size ${DYNAMIC_HEAP_FRACTION} ${DYNAMIC_HEAP_MAX})
+    if [ -n "$CALC_MAX_HEAP" ]; then
+      export KAFKA_HEAP_OPTS="-Xms${CALC_MAX_HEAP} -Xmx${CALC_MAX_HEAP}"
+    fi
+fi
+
+
+. ./set_kafka_gc_options.sh
```
ppatierno: all Kafka related stuff that we don't have here.

scripts/set_kafka_gc_options.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "-verbose:gc -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+PrintGCTimeStamps"
+  fi
+}
+
+if [ "${STRIMZI_KAFKA_GC_LOG_ENABLED}" == "true" ]; then
+    export KAFKA_GC_LOG_OPTS=$(get_gc_opts)
+else
+    export KAFKA_GC_LOG_OPTS=" "
+    export GC_LOG_ENABLED="false"
+fi
```
ppatierno: I guess we don't need this, it's about Kafka ...

scripts/kafka_bridge_config_generator.sh
scholzj: I don't particularly like having two configurations ... but it sounds logical. We can also have only one set and paste it into the configuration twice with different prefixes. Kafka clients usually handle well this kind of tricks (Connect does it all the time). And I'm not sure whether there is a real need to configure the same option differently for producer and consumer. But it sould be more in sync with what we have for RHEL now.

ppatierno: After discussion related to my PR #207 , it's better having common properties like bootstrap servers, security and auth related prefixed with `kafka.` while the specific one for producer and consumer with `kafka.producer.` and `kafka.producer.`. @stanlyDoge FYI

.checkstyle/suppressions.xml
stanlyDoge: I added there a check whether the config file is specified and that increased `NPathComplexity`.

scripts/kafka_bridge_config_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+    fi
+
+    SASL_AUTH_CONFIGURATION=$(cat <<EOF
+kafka.sasl.mechanism=${SASL_MECHANISM}
+kafka.sasl.jaas.config=org.apache.kafka.common.security.${JAAS_SECURITY_MODULE} required username="${KAFKA_BRIDGE_SASL_USERNAME}" password="${PASSWORD}";
+EOF
+)
+fi
+
+content="#Common properties\n"
```
ppatierno: "Kafka common properties"

ppatierno: Did you have problem to do the same as for MM using `cat` command instead of making this big "content"?

scripts/kafka_bridge_config_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+
+content="$content\n"
+content="$content#AMQP\n"
+content="${content}amqp.enabled=$KAFKA_BRIDGE_AMQP_ENABLED\n"
+content="${content}amqp.host=$KAFKA_BRIDGE_AMQP_HOST\n"
+content="${content}amqp.port=$KAFKA_BRIDGE_AMQP_PORT\n"
+content="${content}amqp.mode=$KAFKA_BRIDGE_AMQP_MODE\n"
+content="${content}amqp.flowCredit=$KAFKA_BRIDGE_AMQP_FLOW_CREDIT\n"
+content="${content}amqp.certDir=$KAFKA_BRIDGE_AMQP_CERT_DIR\n"
+content="${content}amqp.messageConverter=$KAFKA_BRIDGE_AMQP_MESSAGE_CONVERTER\n"
```
ppatierno: I think we should avoid AMQP configuration which is something we are not going to support in Strimzi right now. Are these env vars even set in the related CO integration?

scripts/kafka_bridge_config_generator.sh
scholzj: This is super ugly, you should do it do the same as you are doing above I think.

scripts/kafka_bridge_config_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+done
+
+content="$content\n"
+content="$content#HTTP\n"
+content="${content}http.enabled=$KAFKA_BRIDGE_HTTP_ENABLED\n"
+content="${content}http.host=$KAFKA_BRIDGE_HTTP_HOST\n"
+content="${content}http.port=$KAFKA_BRIDGE_HTTP_PORT\n"
+
+content="$content\n"
+content="$content#AMQP\n"
```
scholzj: I'm fine with having the env vars and the options here is someone wants to play with it. But should we just wrap it into if and do not print them into the config file when AMQP is disabled? 

(Also, they should be right now not activated from the CO)

scripts/kafka_bridge_config_generator.sh
stanlyDoge: No, they are not set so they are empty.

scripts/kafka_bridge_config_generator.sh
stanlyDoge: I had a problem with new lines. I will take a better look :)

scripts/kafka_bridge_config_generator.sh
ppatierno: but the properties file will have a bunch of empty fields. I would void to write them is `AMQP_ENABLED` is false, the same for `HTTP_ENABLED`.

scripts/kafka_bridge_config_generator.sh
ppatierno: I said same thing above ... sorry for duplcating (but maybe enforcing) the concept :-)

PR #207, opened 2019-05-28
Refactoring on bridge configuration

### LINKED ISSUE #205: Allow consumer and producer configuration within the properties file
Instead of having a sort of "hard-coded" parameters and fields for the bridge configuration we should also allow consumer and producer configuration.
The properties file should use `kafka.consumer` and `kafka.producer` prefixes for all consumer and producer related parameters that could be overridden by the specific ones when an HTTP client connect.
The same for specific configuration related to HTTP and AMQP so using prefixes `http` and `amqp`.

ppatierno:
This PR fixes #205


ppatierno:
As agreed offline I will change it a little bit.
Common configuration parameters like bootstrap servers, security and auth related can be specified with `kafka.` prefix and they are common to both producer and consumer.
Specific configurations will have `kafka.producer.` and `kafka.consumer.` prefixes and if someone will be the same as the common, they will take precedence overriding the common ones.

PR #210, opened 2019-05-29
JSON error support and HTTP responses

### LINKED ISSUE #193: Return a JSON with error details
When the bridge returns any kind of error with HTTP status codes 4xx or 5xx it should send a JSON message in the body with some error details in the following format.

```json
{
    "error_code": 422,
    "message": "The records collection cannot be empty"
}
```

ppatierno:
This PR fixes #193 and #180


stanlyDoge:
Wouldn't it be easier just to put json body as the parameter of the `end()` method? For example:
```
private void sendConsumerDeletionResponse(HttpServerResponse response) {
        response.setStatusCode(204).end("{some json}");
}
```

ppatierno:
@stanlyDoge I have to write body only if I have a body. If I remove the `write` from the "if" condition I have to add another "if" later to call different `end` accepting a body (as Buffer) or not, so I guess the current way is better. Wdyt?

Issue #184, opened 2019-05-21
Add health endpoint for readiness and liveness on Kubernetes


ppatierno:
We need an endpoint, i.e. `/health` for readiness and liveness probes on Kubernetes.

PR #211, opened 2019-05-29
Readiness check

stanlyDoge:
This is based on https://github.com/strimzi/strimzi-kafka-bridge/pull/206 so we need to wait until it is merged.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -96,6 +96,7 @@ public class HttpBridge extends AbstractVerticle {
                 routerFactory.addHandlerByOperationId(HttpOpenApiOperations.SEEK.toString(), this::seek);
                 routerFactory.addHandlerByOperationId(HttpOpenApiOperations.SEEK_TO_BEGINNING.toString(), this::seekToBeginning);
                 routerFactory.addHandlerByOperationId(HttpOpenApiOperations.SEEK_TO_END.toString(), this::seekToEnd);
+                routerFactory.addHandlerByOperationId(HttpOpenApiOperations.READINESS_CHECK.toString(), this::readinessCheck);
 
                 routerFactory.addFailureHandlerByOperationId(HttpOpenApiOperations.SEND.toString(), this::send);
                 routerFactory.addFailureHandlerByOperationId(HttpOpenApiOperations.SEND_TO_PARTITION.toString(), this::sendToPartition);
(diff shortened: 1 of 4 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -260,6 +261,10 @@ private void seekToEnd(RoutingContext routingContext) {
         processConsumer(routingContext);
     }
 
+    private void readinessCheck(RoutingContext routingContext) {
+        routingContext.response().setStatusCode(200).end();
```
ppatierno: Now we have an `HttpUtils` class for sending response to the HTTP clients, we have to use that one. And I am using `HttpResponseStatus` class for the status codes (i.e. `OK` which is 200)

ppatierno:
@stanlyDoge  btw this PR should fix #184 so it means that we need both readiness and liveness endpoint.
I would use `/healthy` for the liveness and `/ready` for the readiness as we have in the TO.
So OpenAPI spec has to be updated accordingly.

ppatierno:
I thought more about that. The healthy and ready endpoints cannot live in the "route" we are using for the HTTP bridge even because enabling AMQP and disabling HTTP we won't have checks anymore!
The readiness/liveness thing should live outside of the verticles running AMQP and HTTP protocol heads.
The solutions I see are:

1. having a different HTTP server living at a higher level of the bridge (independent from the one used for the HTTP support), exposing `/healthy` and `/ready' endpoints (on different ports). This server will be hit by readiness/liveness probes to check the availability of the bridge in general, in both cases, it has or not AMQP and/or HTTP enabled/disabled.
2. we can add readiness/liveness endpoints to the current HTTP server used when HTTP is enabled but then when AMQP is enabled we need two probes, one hitting AMQP endpoint and the other one HTTP endpoint. Maybe it's not possible in Kubernetes, so maybe a command probe should be developed to hit both of them (or the enabled ones, thinking more protocols in the future as well, i.e. MQTT).

FYI, currently, the AMQP verticle starts an HTTP server for readiness/liveness and this was causing a problem I was facing: having both protocols enabled so two verticles running, the AMQP one was using port 8080 for readiness/liveness and the HTTP one for the actual HTTP bridging. Every HTTP client request was satisfied every two calls, because Vert.x was load balancing the HTTP traffic on port 8080 between the two servers (the one for AMQP readiness and the one for HTTP bridiging). It means that we have to remove the readiness server from the AMQP verticle anyway.

@stanlyDoge @tombentley @scholzj any thoughts? even about the above solutions or different ones you have in mind?

tombentley:
Given that we're not currently fully committed to AMQP and HTTP is the current priority I would say to do something simple for now. It's not set in stone and we can always revise it in the future.

ppatierno:
So I would go for a quick solution first. I will remove the HTTP health server from the AMQP verticle and I will provide this one at upper-level listening on a different port, let's say 8081.
I would remove the healthy and ready endpoints from the OpenAPI as well.

stanlyDoge:
Which endpoint will CO "ping" then?

ppatierno:
In my mind they are always `/healthy` and `/ready` but on port 8081. I am working on that.

ppatierno:
I am going to close this one for #222

PR #212, opened 2019-05-29
Fix JsonObject serializing/deserializing

### LINKED ISSUE #208: JSON value not encoded/decoded properly
If a producer sends a record with the following JSON in the HTTP body:

```json
{
	"records": [
		{
			"key": "key1",
			"value": { "foo": "bar", "aaa": 5 }
		}
	]
}
```

the consumer will get the following one:

```json
[
    {
        "topic": "test",
        "key": "key1",
        "value": {
            "map": {
                "foo": "bar",
                "aaa": 5
            },
            "empty": false
        },
        "partition": 0,
        "offset": 10
    }
]
```

where the "value" it's not exactly what expected.

ppatierno:
This PR fixes #208 .
We should need more unit tests for that but I am planning for a different PR for now.

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -8,7 +8,7 @@ kafka.producer.acks=1
 kafka.consumer.auto.offset.reset=earliest
 
 #AMQP related settings
-amqp.enabled=true
+#amqp.enabled=true
 amqp.flowCredit=100
 amqp.mode=SERVER
 amqp.host=0.0.0.0
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -8,7 +8,7 @@ kafka.producer.acks=1
 kafka.consumer.auto.offset.reset=earliest
 
 #AMQP related settings
-#amqp.enabled=true
+amqp.enabled=true
 amqp.flowCredit=100
 amqp.mode=SERVER
 amqp.host=0.0.0.0
--- a/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
@@ -8,8 +8,6 @@ package io.strimzi.kafka.bridge.http;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.strimzi.kafka.bridge.BridgeContentType;
 import io.strimzi.kafka.bridge.KafkaClusterTestBase;
-import io.strimzi.kafka.bridge.KafkaJsonDeserializer;
-import io.strimzi.kafka.bridge.KafkaJsonSerializer;
 import io.strimzi.kafka.bridge.amqp.AmqpConfig;
 import io.strimzi.kafka.bridge.config.KafkaConfig;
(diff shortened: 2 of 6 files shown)


src/main/resources/application.properties
```diff
@@ -8,7 +8,7 @@ kafka.producer.acks=1
 kafka.consumer.auto.offset.reset=earliest
 
 #AMQP related settings
-amqp.enabled=true
+#amqp.enabled=true
```
tombentley: Does this change really belong in this PR?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -49,6 +47,7 @@
 import java.util.stream.Collectors;
 
 @RunWith(VertxUnitRunner.class)
+@SuppressWarnings("checkstyle:JavaNCSS")
```
tombentley: I think you should be able to annotate the method, rather than the whole class.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -2770,4 +2769,148 @@ public void formatAndAcceptMismatch(TestContext context) {
...
+
+        Async async = context.async();
+
+        JsonObject sentKey = new JsonObject()
+                .put("f1", "v1")
+                .put("array", new JsonArray().add(1).add(2));
+
+        JsonObject sentValue = new JsonObject()
+                .put("array", new JsonArray().add("v1").add("v2"))
+                .put("foo", "bar").put("number", 123);
```
tombentley: TBH I would like to see nested objects tested too.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -2770,4 +2769,148 @@ public void formatAndAcceptMismatch(TestContext context) {
...
+                    JsonObject jsonResponse = response.body().getJsonObject(0);
+
+                    String kafkaTopic = jsonResponse.getString("topic");
+                    int kafkaPartition = jsonResponse.getInteger("partition");
+                    JsonObject key = jsonResponse.getJsonObject("key");
+                    JsonObject value = jsonResponse.getJsonObject("value");
+                    long offset = jsonResponse.getLong("offset");
+
+                    context.assertEquals(topic, kafkaTopic);
+                    context.assertTrue(sentValue.equals(value));
```
tombentley: `assertEquals()` would be more conventional.

src/main/resources/application.properties
ppatierno: No you are right, I was testing offline with postman and removed the AMQP stuff.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
ppatierno: Reading about the metric I think that in this case it's about the whole class which is becoming quite big even if QE people are working on having it better.

PR #220, opened 2019-05-30
Add first Makefiles

### LINKED ISSUE #213: Add a Makefile based environment for building package and Docker image
As it happens for the Strimzi core project we should set up a Makefile based environmente for building package and Docker images (even pushing to Docker Hub) in order to automate the release process as well.

scholzj:
This adds the initial Make build for the bridge and integrates it into Travis. You can do `make all` to build the Java code and build and push the Docker image. You can also use following targets:
* `make java_install`
* `make java_package`
* `make java_verify`
* `make findbugs`
* `make docker_build`
* `make docker_push`

The env. vars `DOCKER_ORG` and `DOCKER_TAG` can be used to control how the Docker image is going to be pushed.

This should close #213.


Dockerfile
```diff
@@ -1,13 +1,26 @@
...
-# copy scripts for starting the bridge
-COPY scripts/ /
+COPY scripts/ bin/
+COPY target/kafka-bridge.jar lib/
 
-EXPOSE 5672 5671 8080
+USER 1001
 
-CMD ["/run_bridge.sh"]
+CMD ["/opt/strimzi/bin/run_bridge.sh"]
```
ppatierno: is this change coordinated with Stanislav PR, he is removing the CMD line if I recall well. He is also using a different /opt subfolder. See #206

Dockerfile
```diff
@@ -1,13 +1,26 @@
...
+
+# Set JAVA_HOME env var
 ENV JAVA_HOME /usr/lib/jvm/java
 
-ADD target/kafka-bridge.jar /
+# Add strimzi user with UID 1001
+# The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1001 -g 0 strimzi
+
+ARG strimzi_version=1.0-SNAPSHOT
```
ppatierno: is this the version of the bridge? it should be 0.12.0-SNAPSHOT right now.

Dockerfile
```diff
@@ -1,13 +1,26 @@
...
+    && yum -y install java-${JAVA_VERSION}-openjdk-headless openssl \
+    && yum -y clean all
+
+# Set JAVA_HOME env var
 ENV JAVA_HOME /usr/lib/jvm/java
 
-ADD target/kafka-bridge.jar /
+# Add strimzi user with UID 1001
+# The user is in the group 0 to have access to the mounted volumes and storage
+RUN useradd -r -m -u 1001 -g 0 strimzi
```
ppatierno: do we need that now? we don't have volumes and storage here.

Dockerfile
```diff
@@ -1,13 +1,26 @@
...
+
+ARG strimzi_version=1.0-SNAPSHOT
+ENV STRIMZI_KAFKA_BRIDGE_VERSION ${strimzi_kafka_bridge_version}
```
ppatierno: if it's the bridge version, should the above `strimzi_version` have the same name? are they the same thing or I missunderstood?

Makefile
```diff
@@ -0,0 +1,38 @@
...
+GITHUB_VERSION ?= master
+RELEASE_VERSION ?= latest
+
+ifneq ($(RELEASE_VERSION),latest)
+  GITHUB_VERSION = $(RELEASE_VERSION)
+endif
+
+all: java_package docker_build docker_push
+clean: java_clean
+release: release_prepare release_version release_helm_version release_maven $(SUBDIRS) release_docu release_single_file release_pkg release_helm_repo docu_clean
```
ppatierno: we don't have helm. Not sure that all these above are existing targets following here.

Dockerfile
scholzj: Well, one of us will need to reconcile. This follows the images from main repo, so I think this is much better then just pushing stuff into root.

Dockerfile
scholzj: This is just a default. So that does not matter what is there since it will be overwritten in the build. But is had a wrong name (should be `strimzi_kafka_bridge_version`).

Dockerfile
scholzj: We should still not run the images under root since that is unsecure.

Makefile
scholzj: Yeah, that were leftovers - I cleaned them up. But I guess the release will anyway need some finishing up to package the docs etc. But that should be done later.

Dockerfile
ppatierno: @stanlyDoge ^^

Dockerfile
scholzj: We should still not run the images under root since that is unsecure.

Dockerfile
ppatierno: Makes sense

scholzj: Yeah, fixed.

Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+ifneq ($(RELEASE_VERSION),latest)
+  GITHUB_VERSION = $(RELEASE_VERSION)
+endif
+
+all: java_package docker_build docker_push
+clean: java_clean
+release: release_prepare release_maven release_package
+
+next_version:
```
ppatierno: what's this next_version  and the related `NEXT_VERSION` env var?

scholzj: This is for setting the new version in master after a release ... e.g. to bump it from 0.12.0-SNAPSHOT to 0.13.0-SNAPSHOT.

ppatierno: Thanks! I am going to merge.

PR #221, opened 2019-05-30
Fix issues found by Findbugs

scholzj:
There are several issues detected by Findbugs ... this PR fixes all of them so that we can add findbugs into the Travis build.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -294,7 +294,7 @@ public class HttpBridge extends AbstractVerticle {
         SourceBridgeEndpoint source = this.httpBridgeContext.getHttpSourceEndpoints().get(httpServerRequest.connection());
 
         if (source == null) {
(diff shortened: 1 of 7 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -294,7 +294,7 @@ private void processProducer(RoutingContext routingContext) {
         SourceBridgeEndpoint source = this.httpBridgeContext.getHttpSourceEndpoints().get(httpServerRequest.connection());
 
         if (source == null) {
-            source = new HttpSourceBridgeEndpoint<>(this.vertx, this.httpBridgeConfig, this.httpBridgeContext,
+            source = new HttpSourceBridgeEndpoint<>(this.vertx, this.httpBridgeConfig,
```
ppatierno: I left the field for simmetry with the sink endpoint, maybe used in the future but if findbugs is complaining...

scholzj: We need to create some fake use or remove it to silence it ;-).

Issue #219, opened 2019-05-30
HTTP endpoint hit every two HTTP client calls


ppatierno:
When both AMQP and HTTP endpoints are enabled, the HTTP one is hit every two calls from an HTTP client (i.e. for sending a message the HTTP POST on /topics/my-topic has to be done twice).
It seems to be related to the fact that the AMQP endpoint starts another HTTP server listening on port 8080 for readiness/liveness checks. Because it's in another verticle, the Vert.x stuff is doing round-robin balancing distributing HTTP calls to this server and the one hosting the HTTP endpoint.
We should remove the readiness/liveness from inside AMQP moving to upper level as common between AMQP and HTTP endpoints and listening on a different port.

PR #222, opened 2019-05-30
Added a unique health server for the bridge

### LINKED ISSUE #184: Add health endpoint for readiness and liveness on Kubernetes

ppatierno:
This PR fixes #184 and #219

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -32,7 +32,9 @@ public class Application {
 
     private static final Logger log = LoggerFactory.getLogger(Application.class);
 
-    private static final int HEALTH_SERVER_PORT = 8081;
+    private static final String HEALTH_SERVER_PORT = "HEALTH_SERVER_PORT";
+
+    private static final int DEFAULT_HEALTH_SERVER_PORT = 8081;
 
     public static void main(String[] args) {
         Vertx vertx = Vertx.vertx();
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -27,6 +32,8 @@
 
     private static final Logger log = LoggerFactory.getLogger(Application.class);
 
+    private static final int HEALTH_SERVER_PORT = 8081;
```
scholzj: Since the HTTP port is configurable, I think this should be configurable as well and might need to prevent conflicts with the regular port.

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -47,32 +54,75 @@ public static void main(String[] args) {
...
+     * Start an HTTP health server
+     */
+    private static void startHealthServer(Vertx vertx) {
+
+        vertx.createHttpServer()
+                .requestHandler(request -> {
+                    if (request.path().equals("/healthy")) {
+                        request.response().setStatusCode(HttpResponseStatus.OK.code()).end();
+                    } else if (request.path().equals("/ready")) {
+                        request.response().setStatusCode(HttpResponseStatus.OK.code()).end();
```
scholzj: Should this be bound to the Kafka connectivity? Maybe at least as a issue for the future? Or is that a problem from the technical perspective?

ppatierno: Yeah I didn't add a fine control just for now but I was going to open an issue to improve the check. We are on the same page.

src/main/java/io/strimzi/kafka/bridge/Application.java
ppatierno: I guess env var is enough here, so it's something that @stanlyDoge has to take into account for his CO integration.

scholzj: Yeah ... unfortunately ... this is why I suggested to not have the HTTP port configurable. Now you will need to deal with conflicts, make them both configurable etc. :-(

ppatierno: @scholzj now the health port is got from env var and it checks conflicts with AMQP and HTTP ports exiting the application.

PR #225, opened 2019-05-31
Junit5 migration

Frawless:
This PR introduce junit  migration from version 4 to 5. I changed some values for checkstyle. However, these values should be changed back after next test refactor.

### DIFF
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -87,13 +87,15 @@
         </module>
 
         <!-- code quality -->
-        <module name="MethodLength"/>
+        <module name="MethodLength">
+            <property name="max" value="170"/>
+        </module>
         <module name="ParameterNumber">
             <!-- default is 8 -->
         </module>
         <module name="ClassDataAbstractionCoupling">
             <!-- default is 7 -->
-            <property name="max" value="12"/>
+            <property name="max" value="15"/>
         </module>
         <module name="BooleanExpressionComplexity">
             <!-- default is 3 -->
@@ -101,7 +103,7 @@
 
         <module name="ClassFanOutComplexity">
             <!-- default is 20 -->
-            <property name="max" value="36"/>
+            <property name="max" value="40"/>
         </module>
         <module name="CyclomaticComplexity">
             <!-- default is 10-->
--- a/.checkstyle/checkstyle.xml
+++ b/.checkstyle/checkstyle.xml
@@ -87,15 +87,13 @@
         </module>
 
         <!-- code quality -->
-        <module name="MethodLength">
-            <property name="max" value="170"/>
-        </module>
+        <module name="MethodLength"/>
         <module name="ParameterNumber">
             <!-- default is 8 -->
         </module>
         <module name="ClassDataAbstractionCoupling">
             <!-- default is 7 -->
-            <property name="max" value="15"/>
+            <property name="max" value="12"/>
         </module>
         <module name="BooleanExpressionComplexity">
             <!-- default is 3 -->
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -112,6 +139,38 @@
 
 	<build>
 		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
```
serrss: Do we need 2 same plugins in one POM.xml? `<artifactId>maven-surefire-plugin</artifactId>` line 144 and line 226

Frawless: Nope, going to remove it, thanks

.checkstyle/checkstyle.xml
```diff
@@ -87,21 +87,23 @@
         </module>
 
         <!-- code quality -->
-        <module name="MethodLength"/>
+        <module name="MethodLength">
```
scholzj: Can't we just suppress these with annotations directly in the code? That is much better then rewriting the rules for all methods.

Frawless: Make sense, let me change it

.checkstyle/checkstyle.xml
```diff
@@ -87,21 +87,23 @@
...
+            <property name="max" value="15"/>
         </module>
         <module name="BooleanExpressionComplexity">
             <!-- default is 3 -->
         </module>
 
         <module name="ClassFanOutComplexity">
             <!-- default is 20 -->
-            <property name="max" value="36"/>
+            <property name="max" value="40"/>
```
ppatierno: I was always trying to not changing the default metrics value to suppress locally or fixing the problems why now we are doubling these values? :-)

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -2249,53 +2196,57 @@ public void seekToBeginningAndReceive(TestContext context) {
...
         kafkaCluster.createTopic(topic, 1, 1);
 
-        Async creationAsync = context.async();
+        CompletableFuture<Boolean> produce = new CompletableFuture<>();
+        CompletableFuture<Boolean> create = new CompletableFuture<>();
+        CompletableFuture<Boolean> subscribe = new CompletableFuture<>();
+        CompletableFuture<Boolean> delete = new CompletableFuture<>();
+        CompletableFuture<Boolean> seek = new CompletableFuture<>();
+        CompletableFuture<Boolean> dummy = new CompletableFuture<>();
+        CompletableFuture<Boolean> consumeSeek = new CompletableFuture<>();
```
ppatierno: can we create them every time each one is needed?

Frawless: Sure, I was thinking about that as well.

serrss:
@scholzj Sorry, I'm not sure that I understand about what things you are talking about.

serrss:
@scholzj oh, you mean 2 plugins POM file

scholzj:
Exactly.

ppatierno:
@Frawless I don't see any further work on this after my comments. Any news?

Frawless:
@ppatierno  finishing this one https://github.com/strimzi/strimzi-kafka-operator/pull/1613, after that I will get back to this PR.

pom.xml
```diff
@@ -112,6 +139,32 @@
 
 	<build>
 		<plugins>
+			<plugin>
```
scholzj: @Frawless I do not think you should be adding this. The assembly is done on L198-215.

Frawless: @scholzj redundant assembly removed

PR #226, opened 2019-05-31
Manage permissions of the released files

scholzj:
This PR sets the permissions of the files we package in the Maven package phase. It also adds the TAR.GZ packaging which can better preserve the permissions.


scholzj:
Cherry picked into `release-0.12.x` branch.

PR #231, opened 2019-06-04
Removed base directory in the archives (zip and tar)

ppatierno:


ppatierno:
@scholzj what do you think about cherry picking this one to the 0.12.x release branch, I think it's better for having consistent zip content since the first version.

PR #232, opened 2019-06-05
Support openapi enpoint to return openAPI spec

stanlyDoge:


ppatierno:
It's interesting that the Vert.x team has opened this issue [1] around that on which we could rely for having this endpoint out of box.
I would say, let's close this one and wait for them because it's not high priority for us.
@stanlyDoge wdyt?

[1] https://github.com/vert-x3/vertx-openapi/issues/3

PR #235, opened 2019-06-06
Makefile improvements

tombentley:
This PR makes a few improvements to the `Makefile`:

* Use `.mk` extn (which is recognised by tools/editors) for included makefiles.
* Use a variable for docker command, enabling use of podman
* Use temp files to only build jars when the source has changed.

I'd eventually like to make similar changes to the strimzi-kafka-operator build, but thought I'd start here so we can discuss a simpler case.


scholzj:
What is the motivation to change something what works completely fine?

tombentley:
It "works", but the `make` build is far from an optimal and causes me much longer code-compile-debug iterations than is strictly necessary when I need to build images. Where I want to go with these changes is that I can just execute `make all` without having to keep track of what changed (because that's `make`'s job). Really what I'm aiming for is improving the `make` build in the strimzi-kafka-operator, but I wanted to start here because it's much simpler.

ppatierno:
@tombentley is this PR still in place?

scholzj:
Discussed on the community call => can be closed.

PR #241, opened 2019-06-10
Added proper validation handling for required properties

### LINKED ISSUE #239: OpenAPI validation isn't done properly on sending operations
Validation doesn't happen properly when sending a message via the HTTP protocol but the required fields (i.e. the `value` field) is not present.
It means that sending a message with body like this:

```json
{
	"records": [
		{
			"key": "key1",
			"valu": { "foo": "bar", "aaa": 5 }
		}
	]
}
```

where there is no required `value` but the wrong `valu` field, it ends on sending to Kafka a message with null value so the consumer will get:

```json
[
    {
        "topic": "test",
        "key": "key1",
        "value": null,
        "partition": 0,
        "offset": 9
    }
]
```

It turned out that it's related to set the following in the `HttpBridge`:

```java
routerFactory.addFailureHandlerByOperationId(HttpOpenApiOperations.SEND.toString(), this::send);
                routerFactory.addFailureHandlerByOperationId(HttpOpenApiOperations.SEND_TO_PARTITION.toString(), this::sendToPartition);
```

but not calling `routingContext.next` inside.
These failure handlers can be removed, I used them for testing but forgot to remove!

ppatierno:
This fixes #239

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
@@ -1529,8 +1529,8 @@ class HttpBridgeTest extends KafkaClusterTestBase {
                         assertTrue(ar.succeeded());
                         HttpResponse<JsonObject> response = ar.result();
                         HttpBridgeError error = HttpBridgeError.fromJson(response.body());
-                        assertEquals(HttpResponseStatus.BAD_REQUEST.code(), response.statusCode());
-                        assertEquals(HttpResponseStatus.BAD_REQUEST.code(), error.getCode());
+                        assertEquals(HttpResponseStatus.NOT_FOUND.code(), response.statusCode());
+                        assertEquals(HttpResponseStatus.NOT_FOUND.code(), error.getCode());
                     });
                     context.completeNow();
                 });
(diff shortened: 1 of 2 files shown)


src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -1529,8 +1529,8 @@ void invalidRequestTest(VertxTestContext context) {
                         assertTrue(ar.succeeded());
                         HttpResponse<JsonObject> response = ar.result();
                         HttpBridgeError error = HttpBridgeError.fromJson(response.body());
-                        assertEquals(HttpResponseStatus.BAD_REQUEST.code(), response.statusCode());
-                        assertEquals(HttpResponseStatus.BAD_REQUEST.code(), error.getCode());
+                        assertEquals(HttpResponseStatus.NOT_FOUND.code(), response.statusCode());
```
tombentley: Really you ought to rename this `invalidRequestTest()` method to something like `postToNonexistentEndpoint()`, because that's what it's doing. In reality there are many kinds of invalid request, but this method is just testing a single case.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3007,4 +3007,36 @@ void sendReceiveJsonMessage(VertxTestContext context) throws InterruptedExceptio
         delete.get(TEST_TIMEOUT, TimeUnit.SECONDS);
         context.completeNow();
     }
+
+    @Test
+    void sendMessageWithNoRequiredProperty(VertxTestContext context) throws Throwable {
```
tombentley:
```suggestion
    void sendMessageLackingRequiredProperty(VertxTestContext context) throws Throwable {
```

PR #242, opened 2019-06-10
Fix validation unknown properties

### LINKED ISSUE #240: OpenAPI validation isn't done for requests with not defined properties
When sending a message with some additional properties not defined by the OpenAPI definition, they are just ignored with no error returned to the HTTP client.
For example, on producing message:

```json
{
	"records": [
		{
			"key": "key1",
			"value": { "foo": "bar", "aaa": 5 },
			"foo": 10
		}
	]
}
```

the `foo` field not defined in the `ProducerRecord` OpenAPI data type will be just ignored.
We should return an error about bad request to the HTTP client instead.

ppatierno:
This PR fixes #240 but also ...

... fixes some tests due to adding the `additionalProperties: false` which makes the OpenAPI validator doing more work on validation (not done before).
... fix a "bug" in the documentation generation which was blocking Travis to be successful
... defines the `partition-id` as Integer instead of String in the related endpoint for validating it

Please note that, even if the OpenAPI JSON file seems to be different with huge parts removed, actually there are only a few small changes (it seems that apicurio likes to move things sometimes).


.gitignore
```diff
@@ -29,3 +29,6 @@ hs_err_pid*
 .vscode/
 
 .DS_Store
+
+# Auto generated doc
+**/html*/*
```
scholzj:
```suggestion
documentation/html/**
documentation/htmlnoheader/**
documentation/book/build/**
```

PR #243, opened 2019-06-11
Make TLS work again when used with StrimziCO

scholzj:
The TLs doesn't work currently when the image is used from Docker. This is because the startup scripts expect the certificates to be under `/opt/kafka` but they are under `/opt/strimzi`. So it never finds them and the Bridge will never connect. This PR changes the paths to use the right ones.


scholzj:
(this has to be picked for the 0.12.0 release branch)

PR #245, opened 2019-06-12
Separate api doc generation from main build

tombentley:
This PR separates the generation of the API documentation from the main maven build and adjusts the `Makefile` so the API docs are generated as part of the `docu_html` target. In addition the `docu_check` target will also check the generated docs, ensuring that the generated docs stay in sync with the `openapiv2.json`.


ppatierno:
@tombentley does it need to be cherry-picked to 0.12.0 branch? If yes, please do.

PR #251, opened 2019-06-14
Catching exceptions on native Kafka client creation and return proper error

### LINKED ISSUE #250: Catching exception on native Kafka client creation
If any problems happen during the creation of a native Kafka consumer or producer in the related sink/source endpoint, the bridge just returns HTTP internal server error 500 without any error message or more information. (see for example, #248).
It would be better to catch exceptions and returning more information to the HTTP client.

ppatierno:
This PR fixes #250

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -315,7 +315,9 @@ public class HttpBridge extends AbstractVerticle {
             source.handle(new HttpEndpoint(routingContext));
 
         } catch (Exception ex) {
-            source.close();
+            if (source != null) {
+                source.close();
+            }
             HttpBridgeError error = new HttpBridgeError(
                     HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
                     ex.getMessage()
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -197,6 +197,7 @@ public class HttpBridge extends AbstractVerticle {
                 httpBridgeContext.getHttpSinkEndpoints().put(consumerId.toString(), sink);
             });
         } catch (Exception ex) {
+            sink.close();
             HttpBridgeError error = new HttpBridgeError(
                     HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
                     ex.getMessage()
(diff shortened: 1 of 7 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -190,11 +190,20 @@ private void createConsumer(RoutingContext routingContext) {
...
+                httpBridgeContext.getHttpSinkEndpoints().put(consumerId.toString(), sink);
+            });
+        } catch (Exception ex) {
+            HttpBridgeError error = new HttpBridgeError(
+                    HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
+                    ex.getMessage()
+            );
+            HttpUtils.sendResponse(routingContext.response(), HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
+                    BridgeContentType.KAFKA_JSON, error.toJson().toBuffer());
+        }
```
tombentley: If the `sink.handle()` throws then do we need to clean up side-effects of the `sink.open()`? IOW can the exception result in any kind of resource leakage?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -291,17 +300,27 @@ private void processProducer(RoutingContext routingContext) {
...
+            }
+            source.handle(new HttpEndpoint(routingContext));
+
+        } catch (Exception ex) {
+            HttpBridgeError error = new HttpBridgeError(
+                    HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
+                    ex.getMessage()
+            );
+            HttpUtils.sendResponse(routingContext.response(), HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
+                    BridgeContentType.KAFKA_JSON, error.toJson().toBuffer());
```
tombentley: Same comment about possible resource leakage here.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: I took a look and there should not be any resource leakage. I was thinking about the endpoint saved and not removed from the collection so having a pending reference to a not used object (not "garbeageable" then) but it's not the case.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Ditto as above.

ppatierno:
@tombentley actually it was not so true for producers (we have two of them related to AMQP and one of them crashing on creation doesn't clean up the other). So I refactored a little bit the closing procedure and now I close sink/source on exceptions.

ppatierno:
@tombentley I wrote that one but then I deleted it. It's not so simple to test because actually, the only way to inject a crash on producer creation is to provide wrong producer configuration in the HTTP bridge properties file but the bridge is started before the test method running and we cannot change the configuration on fly. We need to think more about that I will open an issue.

ppatierno:
Cherry picked to branch `release-0.12.x`

PR #252, opened 2019-06-14
Logging about HTTP request/response and OpenAPI operations

ppatierno:


src/main/java/io/strimzi/kafka/bridge/http/HttpLogging.java
```diff
@@ -0,0 +1,28 @@
...
+    private static final Logger log = LoggerFactory.getLogger(HttpLogging.class);
+
+    public static void logRequest(HttpServerRequest request) {
+        log.info("Request: method = {}, path = {}", request.method(), request.path());
+        log.debug("Request: headers = {}", request.headers());
+    }
+
+    public static void logResponse(HttpServerResponse response) {
+        log.info("Response: statusCode = {}, message = {}", response.getStatusCode(), response.getStatusMessage());
+    }
```
tombentley: Does these methods really need to be in a separate class?  It looks like the could be moved to `HttpUtils` to me.

src/main/java/io/strimzi/kafka/bridge/http/HttpLogging.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Utility class for logging HTTP requests/responses
+ */
+public class HttpLogging {
+
+    private static final Logger log = LoggerFactory.getLogger(HttpLogging.class);
+
+    public static void logRequest(HttpServerRequest request) {
+        log.info("Request: method = {}, path = {}", request.method(), request.path());
+        log.debug("Request: headers = {}", request.headers());
```
tombentley: When the bridge is serving requests concurrently how will you correlate the request and response log messages? Is there some correlation id which the request and response share, perhaps?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -179,6 +184,7 @@ private void sendToPartition(RoutingContext routingContext) {
 
     private void createConsumer(RoutingContext routingContext) {
         this.httpBridgeContext.setOpenApiOperation(HttpOpenApiOperations.CREATE_CONSUMER);
+        log.info("OpenAPI operation = {}", this.httpBridgeContext.getOpenApiOperation());
```
tombentley: This code `this.httpBridgeContext.setOpenApiOperation()` followed by getting that same value in order to log it is not exactly easy to read.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -263,6 +270,7 @@ private void seekToEnd(RoutingContext routingContext) {
      * @param routingContext RoutingContext instance
      */
     private void processConsumer(RoutingContext routingContext) {
+        log.info("OpenAPI operation = {}", this.httpBridgeContext.getOpenApiOperation());
```
tombentley: You're relying here on getting the operation that the caller set on the context. But why not just do the logging in the caller when you're setting the operation on the context? That the setting and logging is done in the caller just like the code in the other handlers like `deleteConsumer()` and you can factor out a method to do the setting and logging. Or just do the logging in the `context.setOpenApiOperation`.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -67,6 +67,7 @@ public void handle(Endpoint<?> endpoint) {
         // TODO: it seems that getBodyAsJson raises an exception when the body is empty and not null
         try {
             bodyAsJson = routingContext.getBodyAsJson();
+            log.debug("Request: body = {}", bodyAsJson);
```
tombentley: Again, this is find for debugging when you're typically only making a single request at a time, but it needs a correlation id to be useful for analysing production logs.

src/main/java/io/strimzi/kafka/bridge/http/HttpLogging.java
ppatierno: Good point. I didn't like adding a new class as well. I will move the methods and deleting this class.

ppatierno:
@tombentley afaik there is nothing like that in HTTP 1.0. The HTTP 1.1 protocol introduced the pipelining and reading more, the correlation is done adding an additional header in the response so that the client knows what was the related request. Then HTTP 2.0 improved that with multiplexing on same connection and better correlation like AMQP.

tombentley:
@ppatierno I was referring to some mechanism in the vertx API, rather than HTTP itself. If vertx doesn't provide anything directly itself you could use the `System.identityHashCode` of the request (but you'd need to be able to log that even in methods where the request wasn't available, maybe https://vertx.io/docs/vertx-web/java/#_context_data could help with that). But I feel you need _something_, otherwise this logging will only really be useful in very limited debugging scenarios.

ppatierno:
So this is what I meant as at "higher level" in my previous comment. Thanks for pointing me to the doc showing the routing context intrinsic collection.
Btw we are correlating HTTP request/response, so we always have a request to log and the related response. What do you mean with "...where the request wasn't available,...".?

tombentley:
For example your `logResponse` is passed only the response, and `logRequest` is passed only the request, so you cannot emit anything in the log message which would allow you to identify how request and response log messages correlate.

ppatierno:
@tombentley this is why I said that I can change passing the `RoutingContext`

PR #266, opened 2019-07-01
Added check on X-Forwarded headers for building consumers base URI

### LINKED ISSUE #265: Consumer base_uri field filled with API gateway host if needed
When a new consumer is created, the `base_uri` returned in the HTTP body by the bridge contains the address to be used by the client for subsequent consumer operations.
The bridge uses the HTTP "Host" header for getting the address that the HTTP client has to call.
Using an API gateway (i.e. 3scale) in front of the bridge changes the scenario.
The HTTP client is using the API gateway address and it should be used in the future requests but because the bridge isn't aware that the request comes from an API gateway, just use the "Host" header and .... the client gets the bridge address instead of the API gateway one in the `base_uri` field.

For example, client sends request to:

```
http://gateway.192.168.1.9.nip.io:8080/consumers/my-group
```

but the `base_uri` in the JSON response is: 

```
http://bridge.192.168.1.9.nip.io:8083/consumers/my-group/instances/consumer1
```

On solution could be having the API gateway adding a custom header in each request like `API_GATEWAY_HOST` so that the bridge can use this one instead of the default "Host" as base address in the `base_uri`.

@tombentley @scholzj @kyguy @stanlyDoge any thoughts?

ppatierno:
This PR fixes #265

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -36,10 +36,14 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class HttpSinkBridgeEndpoint<K, V> extends SinkBridgeEndpoint<K, V> {
+
+    Pattern hostPattern = Pattern.compile("host=([^;]+)");
+    Pattern protoPattern = Pattern.compile("proto=([^;]+)");
 
     private MessageConverter<K, V, Buffer, Buffer> messageConverter;
 
@@ -453,15 +457,40 @@ public class HttpSinkBridgeEndpoint<K, V> extends SinkBridgeEndpoint<K, V> {
     private String buildRequestUri(RoutingContext routingContext) {
         String baseUri = routingContext.request().absoluteURI();
 
-        String xForwardedHost = routingContext.request().getHeader("x-forwarded-host");
-        String xForwardedProto = routingContext.request().getHeader("x-forwarded-proto");
-        String xForwardedPort = routingContext.request().getHeader("x-forwarded-port");
-        if (xForwardedHost != null && !xForwardedHost.isEmpty() &&
-            xForwardedProto != null && !xForwardedProto.isEmpty() &&
-            xForwardedPort != null && !xForwardedPort.isEmpty()) {
-            baseUri = String.format("%s://%s:%s", xForwardedProto, xForwardedHost, xForwardedPort) +
-                    routingContext.request().path();
+        String xForwarded = routingContext.request().getHeader("x-forwarded");
+        if (xForwarded != null && !xForwarded.isEmpty()) {
+            Matcher hostMatcher = hostPattern.matcher(xForwarded);
+            Matcher protoMatcher = protoPattern.matcher(xForwarded);
+            if (hostMatcher.find() && !hostMatcher.group(1).isEmpty() &&
+                protoMatcher.find() && !protoMatcher.group(1).isEmpty()) {
+
+                baseUri = this.buildBaseUri(routingContext, hostMatcher.group(1), protoMatcher.group(1));
+            }
+        } else {
+            String xForwardedHost = routingContext.request().getHeader("x-forwarded-host");
+            String xForwardedProto = routingContext.request().getHeader("x-forwarded-proto");
+            if (xForwardedHost != null && !xForwardedHost.isEmpty() &&
+                xForwardedProto != null && !xForwardedProto.isEmpty()) {
+
+                baseUri = this.buildBaseUri(routingContext, xForwardedHost, xForwardedProto);
+            }
         }
         return baseUri;
+    }
+
+    private String buildBaseUri(RoutingContext routingContext, String host, String proto) {
+        String[] hostSplit = host.split(":");
+        if (hostSplit.length == 1) {
+            int port;
+            if (proto.equals("http")) {
+                port = 80;
+            } else if (proto.equals("https")) {
+                port = 443;
+            } else {
+                throw new IllegalArgumentException(proto + " is not a valid schema/proto");
+            }
+            return String.format("%s://%s", proto, host + ":" + port) + routingContext.request().path();
+        }
+        return String.format("%s://%s", proto, host) + routingContext.request().path();
     }
 }
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -462,15 +462,17 @@ public class HttpSinkBridgeEndpoint<K, V> extends SinkBridgeEndpoint<K, V> {
             Matcher hostMatcher = forwardedHostPattern.matcher(forwarded);
             Matcher protoMatcher = forwardedProtoPattern.matcher(forwarded);
             if (hostMatcher.find() && protoMatcher.find()) {
-                log.debug("Getting base URI from forwarded: {} HTTP header", forwarded);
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -442,4 +442,26 @@ private boolean checkAcceptedBody(String accept) {
...
+     * Build the request URI for the future consumer requests
+     *
+     * @param routingContext context of the current HTTP request
+     * @return the request URI for the future consumer requests
+     */
+    @SuppressWarnings({"checkstyle:BooleanExpressionComplexity"})
+    private String buildRequestUri(RoutingContext routingContext) {
+        String baseUri = routingContext.request().absoluteURI();
+
+        String xForwardedHost = routingContext.request().getHeader("x-forwarded-host");
```
tombentley: No support for `Forwarded`?

ppatierno: Well, do we really need that? We have to parse for "host" and "proto" inside it and anyway the proxy has to add `X-Forwarded-Port` which is something not defined. What's the advantage of supporting `Forwarded` + `X-Forwarded-Port` instead of having one more header?

tombentley: If you read the documentation carefully you'll see that the port would be included in the `host`, since the `Forwarded.host` is defined to be the value of the `Host` header observed by the proxy, and `Host` is defined to include both hostname and port.

ppatierno: You are right but the TCP port is marked as optional. If not given it implies the default for the service requested, so we should default to 80 for HTTP and 443 for HTTPS.
Anyway, we need some more process on the "host" we get.

tombentley: It just seems strange to me that we're adding support for non-standard `X-` headers without adding support for the standardized header which is intended to replace the use of the non-standard versions. RFC7239 isn't even particularly new, it's half a decade old.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -36,11 +36,15 @@
...
 import java.util.Set;
 import java.util.UUID;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class HttpSinkBridgeEndpoint<K, V> extends SinkBridgeEndpoint<K, V> {
 
+    Pattern hostPattern = Pattern.compile("host=([^;]+)");
+    Pattern protoPattern = Pattern.compile("proto=([^;]+)");
```
tombentley: I think these should be case insensitive matches, if you look at the production for HTTP token in RFC 7230.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -442,4 +446,50 @@ private boolean checkAcceptedBody(String accept) {
...
+     * @return the request URI for the future consumer requests
+     */
+    private String buildRequestUri(RoutingContext routingContext) {
+        String baseUri = routingContext.request().absoluteURI();
+
+        String xForwarded = routingContext.request().getHeader("x-forwarded");
+        if (xForwarded != null && !xForwarded.isEmpty()) {
+            Matcher hostMatcher = hostPattern.matcher(xForwarded);
+            Matcher protoMatcher = protoPattern.matcher(xForwarded);
+            if (hostMatcher.find() && !hostMatcher.group(1).isEmpty() &&
```
tombentley: It can't be empty because you used `+` not `*`.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -442,4 +446,50 @@ private boolean checkAcceptedBody(String accept) {
...
+    /**
+     * Build the request URI for the future consumer requests
+     *
+     * @param routingContext context of the current HTTP request
+     * @return the request URI for the future consumer requests
+     */
+    private String buildRequestUri(RoutingContext routingContext) {
+        String baseUri = routingContext.request().absoluteURI();
+
+        String xForwarded = routingContext.request().getHeader("x-forwarded");
```
tombentley: It's just forwarded, right?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -442,4 +446,50 @@ private boolean checkAcceptedBody(String accept) {
...
+                xForwardedProto != null && !xForwardedProto.isEmpty()) {
+
+                baseUri = this.buildBaseUri(routingContext, xForwardedHost, xForwardedProto);
+            }
+        }
+        return baseUri;
+    }
+
+    private String buildBaseUri(RoutingContext routingContext, String host, String proto) {
+        String[] hostSplit = host.split(":");
```
tombentley: What if host is an ipv6 address, e.g. the url starts `http://[1080:0:0:0:8:800:200C:417A]:8080/`? 

I suppose `host.matches("^.*:[0-9]+$")` is probably sufficient, but you should probably factor out a constant `Pattern`.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -442,4 +446,50 @@ private boolean checkAcceptedBody(String accept) {
...
+
+        String xForwarded = routingContext.request().getHeader("x-forwarded");
+        if (xForwarded != null && !xForwarded.isEmpty()) {
+            Matcher hostMatcher = hostPattern.matcher(xForwarded);
+            Matcher protoMatcher = protoPattern.matcher(xForwarded);
+            if (hostMatcher.find() && !hostMatcher.group(1).isEmpty() &&
+                protoMatcher.find() && !protoMatcher.group(1).isEmpty()) {
+
+                baseUri = this.buildBaseUri(routingContext, hostMatcher.group(1), protoMatcher.group(1));
+            }
```
tombentley: Surely we should have an `else { log }` here, to aid debugging.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: good spot ;)

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: damn, confused by the others starting with `X-`

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -442,4 +447,55 @@ private boolean checkAcceptedBody(String accept) {
...
+        String baseUri = null;
+
+        String forwarded = routingContext.request().getHeader("forwarded");
+        if (forwarded != null && !forwarded.isEmpty()) {
+            Matcher hostMatcher = forwardedHostPattern.matcher(forwarded);
+            Matcher protoMatcher = forwardedProtoPattern.matcher(forwarded);
+            if (hostMatcher.find() && protoMatcher.find()) {
+                log.debug("Getting base URI from forwarded: {} HTTP header", forwarded);
+                baseUri = this.buildBaseUri(routingContext, hostMatcher.group(1), protoMatcher.group(1));
+            }
```
tombentley:
```suggestion
            } else {
                log.debug("Forwarded header '{}' lacked 'host' and/or 'proto' pair; ignoring header", forwarded);
            }
```

PR #275, opened 2019-07-22
Moving the "healthy" and "ready" endpoints as part of the OpenAPI spec

### LINKED ISSUE #264: Moving the "healthy" and "ready" endpoints as part of the OpenAPI spec
Currently, the bridge starts a "health" server on port 8081 by default (configurable via `HEALTH_SERVER_PORT` env var) replying to HTTP requests on `/healthy` and `/ready` endpoints as for liveness and readiness probes on Kubernetes.
This choice was made because we need such support even for AMQP bridging and not just HTTP.
Thinking more I am proposing to:

* start this "health" server only if only AMQP bridging is enabled so that it can use the default 8080 port
* exposing the "health" endpoints in the OpenAPI spec

In this case, the "healthy" is always tested on a standard HTTP 8080 port.
When only AMQP is running (or any other protocol i.e. MQTT which is not HTTP), the "health" server is in charge to reply liveness and readiness requests.
When HTTP is running, it's in charge to expose these endpoints (even if AMQP is running at the same time).

Nothing changes from users/client/Kubernetes perspective because they can always test `/healthy` and `/ready` endpoints on HTTP port 8080.

Any thoughts @tombentley @stanlyDoge @scholzj @kyguy ?

ppatierno:
This PR fixes #264.
Other than that it factors out a class for doing health checks relying on the status of enabled protocols (bridges). Currently, there is not so much logic about when a protocol bridge is ready/healthy (just reusing a `isReady` member that was already there). More complex logic could be added in the future with another PR.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
+++ b/src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
@@ -1,0 +1,17 @@
+package io.strimzi.kafka.bridge;
+
+/**
+ * Represents a component which exposes information about its health
+ */
+public interface HealthCheckable {
+
+    /**
+     * @return if it's healthy answering to requests
+     */
+    boolean isHealthy();
+
+    /**
+     * @return if it's ready to start answering to requests
+     */
+    boolean isReady();
+}
--- a/src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
+++ b/src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright 2019, Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
 package io.strimzi.kafka.bridge;
 
 /**
--- a/src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
+++ b/src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
@@ -13,7 +13,7 @@ public interface HealthCheckable {
     /**
      * @return if it's healthy answering to requests
      */
-    boolean isHealthy();
+    boolean isAlive();
 
     /**
      * @return if it's ready to start answering to requests
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -341,8 +341,8 @@ public class HttpBridge extends AbstractVerticle implements HealthCheckable {
     }
 
     private void healthy(RoutingContext routingContext) {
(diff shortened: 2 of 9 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -335,6 +340,16 @@ private void processProducer(RoutingContext routingContext) {
         }
     }
 
+    private void healthy(RoutingContext routingContext) {
+        HttpResponseStatus httpResponseStatus = this.healthChecker.isHealthy() ? HttpResponseStatus.OK : HttpResponseStatus.INTERNAL_SERVER_ERROR;
```
tombentley: Is 500 really the best choice? 500 does sort of make sense for healthy, but for readiness 500 is a little ambiguous and 404 would be a better fit, no?

src/main/resources/openapi.json
```diff
@@ -1022,6 +1022,28 @@
...
+        },
+        "/healthy": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "The bridge is healthy"
+                    }
+                },
+                "operationId": "healthy",
+                "description": "Check if the bridge is healthy and can accept requests."
```
```suggestion
                "description": "Check if the bridge is running. This does not necessarily imply that it is ready to accept requests."
```

src/main/resources/openapiv2.json
```diff
@@ -854,6 +854,28 @@
...
+    },
+    "/healthy": {
+      "get": {
+        "responses": {
+          "200": {
+            "description": "The bridge is healthy"
+          }
+        },
+        "operationId": "healthy",
+        "description": "Check if the bridge is healthy and can accept requests."
```
tombentley:
```suggestion
        "description": "Check if the bridge is running. This does not necessarily imply that it is ready to accept requests."
```

src/main/java/io/strimzi/kafka/bridge/HealthCheckable.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Represents a component which exposes information about its health
+ */
+public interface HealthCheckable {
+
+    /**
+     * @return if it's healthy answering to requests
+     */
+    boolean isHealthy();
```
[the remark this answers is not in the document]
ppatierno: I am fine with that I was just using the same name as the endpoint. I will change, agree.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: I put 500 but I was not so happy about that. The 404 is returned when I resource doesn't exist, so it could fit well but it is also returned when a GET is done on a not existing endpoint. For example, for the HTTP bridge if you do a GET on the root "/" it returns 404 because there is no Vert.x route handling that (no endpoint exposed on the OpenAPI spec). So could it make confusion?

tombentley: Yeah, I wondered about that, but I don't think it really makes a difference. I don't think there's any diagnostic value in having them be different.

PR #276, opened 2019-07-22
Deleted Kubernetes deployment YAMLs

ppatierno:
On Kubernetes, the bridge is deployed via the Strimzi operator which is the preferred way so we can delete these not up to date YAML files.


tombentley:
I'm not sure I agree about this. While we want people to use the kafka operator to deploy the bridge, by removing these resources we're making life harder for anyone who doesn't, or who just wants to experiment with the bridge. AFAICS there's very little cost in maintaining these resources, so I feel we should keep them, even if we decide not to document this style of deployment in order to push the operator way.

scholzj:
While not necessarily disagreeing that the files might be useful for someone, when I look at the content it seems that there is lot more maintenance effort needed than we gave it:
* Nobody updated it for the HTTP Bridge
* It is not plugged into the release process
* Nobody noticed / complained

ppatierno:
> Nobody updated it for the HTTP Bridge

> Nobody noticed / complained

These are the reasons why I deleted the files.

> It is not plugged into the release process

what do you mean by that? that we should copy this files in the generated zip files?

scholzj:
> what do you mean by that? that we should copy this files in the generated zip files?

Well, if nothing else it needs to update the tags of the Docker images from `:latest` to `:x.y.z` in the release tag. Possibly we should also upload the modified files to the GitHub release page.

it's not like it would some super huge effort. But today these are really stale files which were not used or updated for a long time.

ppatierno:
I can close this PR but it doesn't sound high priority to me right now maintaining these files and including them in the release process.

PR #277, opened 2019-07-23
Exposing openapi endpoint for getting OpenAPI v2 spec in JSON format

ppatierno:
This PR exposes the OpenAPI v2 specification in JSON format on the `/openapi` endpoint.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
+++ b/src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
@@ -16,4 +16,5 @@ public class BridgeContentType {
     public static final String KAFKA_JSON_BINARY = "application/vnd.kafka.binary.v2+json";
     // JSON encoding
     public static final String KAFKA_JSON = "application/vnd.kafka.v2+json";
+    public static final String JSON = "application/json";
 }
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -15,6 +15,7 @@ import io.strimzi.kafka.bridge.SourceBridgeEndpoint;
 import io.strimzi.kafka.bridge.http.model.HttpBridgeError;
 import io.vertx.core.AbstractVerticle;
 import io.vertx.core.Future;
+import io.vertx.core.file.FileSystem;
 import io.vertx.core.http.HttpConnection;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerOptions;
@@ -24,6 +25,7 @@ import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.api.contract.openapi3.OpenAPI3RouterFactory;
 import io.vertx.ext.web.api.validation.ValidationException;
+
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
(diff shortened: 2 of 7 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -350,6 +353,21 @@ private void ready(RoutingContext routingContext) {
         HttpUtils.sendResponse(routingContext, httpResponseStatus.code(), null, null);
     }
 
+    private void openapi(RoutingContext routingContext) {
+        FileSystem fileSystem = vertx.fileSystem();
+        fileSystem.readFile("openapiv2.json", readFile -> {
+            if (readFile.succeeded()) {
+                HttpUtils.sendResponse(routingContext, HttpResponseStatus.OK.code(), BridgeContentType.JSON, readFile.result());
```
tombentley: If you are going to access the json as a file (rather than classpath) then it would be simpler and more efficient to use just `HttpServerResponse.sendFile()`.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
```diff
@@ -16,4 +16,5 @@
     public static final String KAFKA_JSON_BINARY = "application/vnd.kafka.binary.v2+json";
     // JSON encoding
     public static final String KAFKA_JSON = "application/vnd.kafka.v2+json";
+    public static final String JSON = "application/json";
```
tombentley: `BridgeContentType` is supposedly about "bridge supported content types" (according to its Javadoc), but this content type cannot be used as the content type for a message, so I don't think this content type really belongs here. Alternatively if `BridgeContentType` is really just some constants for content types then it should be renamed to `ContentType`.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -350,6 +353,21 @@ private void ready(RoutingContext routingContext) {
         HttpUtils.sendResponse(routingContext, httpResponseStatus.code(), null, null);
     }
 
+    private void openapi(RoutingContext routingContext) {
+        FileSystem fileSystem = vertx.fileSystem();
+        fileSystem.readFile("openapiv2.json", readFile -> {
```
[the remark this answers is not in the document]
ppatierno: Vert.x provide the `FileResolver` class used by the `FileSystem`. It copies files from classpath in a cacheDir. The option is enabled by default but can be disabled via `VertxOptions` class.

```
vo.getFileSystemOptions().setClassPathResolvingEnabled(false);
```

It's also possible to disable caching. 
In our case we are using default options and in the bridge run script we already set `-Dvertx.cacheDirBase=/tmp`.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Yeah if we agree to not use classpath (see comment above) then it makes sense.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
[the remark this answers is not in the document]
tombentley: Ah, I didn't know that. Thank you!

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
ppatierno: From the beginning, I decided to use `Bridge` prefix because these content types were strictly related to the contents that the bridge is able to handle. Actually, clients should be able to handle `application/json` as well together with `application/vnd.kafka.json.v2+json` and `application/vnd.kafka.v2+json`. So actually this is a bridge supported content type.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tombentley: I'm fine with the caching described above, so using `sendFile()` seems like the right way to handle this.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
tombentley: Fair enough.

Issue #273, opened 2019-07-05
Generating random consumer names based on bridge id


ppatierno:
When, on consumer creation, a "name" is not specified, it's created by:

```java
"kafka-bridge-consumer-" + UUID.randomUUID()
```

In this way during debugging or log analysis is not so simple to correlate consumers with related bridges so it could be useful to add an `id` configuration parameter in the property file used as prefix instead of the hardcoded "kafka-bridge-consumer-" string. So using something like:

```java
bridgeId + UUID.randomUUID()
```

PR #279, opened 2019-07-24
Consumer name is prefixed by Bridge ID when available

stanlyDoge:
https://github.com/strimzi/strimzi-kafka-bridge/issues/273

`name` neither `bridge.id` set.
```
{
    "instance_id": "kafka-bridge-consumer-a613d84b-4842-4a91-a5e3-ffe251b285e6",
    "base_uri": "http://localhost:8080/consumers/my-group/instances/kafka-bridge-consumer-a613d84b-4842-4a91-a5e3-ffe251b285e6"
}
```

`name` not set, `bridge.id` set to "my-bridge"
```
{
    "instance_id": "my-bridge-aafec785-0dab-4a49-830f-5339b664a2f5",
    "base_uri": "http://localhost:8080/consumers/my-group/instances/my-bridge-aafec785-0dab-4a49-830f-5339b664a2f5"
}
```

### DIFF
--- a/bin/docker/kafka_bridge_config_generator.sh
+++ b/bin/docker/kafka_bridge_config_generator.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+BRIDGE_PROPERTIES=$(cat <<-EOF
+#Bridge configuration
+bridge.id=${KAFKA_BRIDGE_ID}
+EOF
+)
+
 SECURITY_PROTOCOL=PLAINTEXT
 
 if [ "$KAFKA_BRIDGE_TLS" = "true" ]; then
@@ -105,6 +111,8 @@ EOF
 
 # if http/amqp is disabled, do not print its configuration
 PROPERTIES=$(cat <<EOF
+BRIDGE_PROPERTIES
+
 $KAFKA_PROPERTIES
 
 $PRODUCER_PROPERTIES
--- a/bin/docker/kafka_bridge_config_generator.sh
+++ b/bin/docker/kafka_bridge_config_generator.sh
@@ -111,7 +111,7 @@ EOF
 
 # if http/amqp is disabled, do not print its configuration
 PROPERTIES=$(cat <<EOF
-BRIDGE_PROPERTIES
+$BRIDGE_PROPERTIES
 
 $KAFKA_PROPERTIES
 
--- a/config/application.properties
+++ b/config/application.properties
@@ -1,3 +1,6 @@
+#Bridge related settings
+bridge.id=my-bridge
+
 #Apache Kafka common
 kafka.bootstrap.servers=localhost:9092
 
--- a/src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
+++ b/src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
@@ -88,6 +88,10 @@ public class BridgeConfig extends AbstractConfig {
     }
 
     public String getBridgeID() {
-        return config.get(BridgeConfig.BRIDGE_ID) == null ? null : config.get(BridgeConfig.BRIDGE_ID).toString();
+        if (config.get(BridgeConfig.BRIDGE_ID) == null) {
+            return null;
+        } else {
+            return config.get(BridgeConfig.BRIDGE_ID).toString();
+        }
     }
 }
--- a/src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
+++ b/src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
@@ -18,6 +18,8 @@ public class BridgeConfig extends AbstractConfig {
 
     public static final String BRIDGE_CONFIG_PREFIX = "bridge.";
 
+    public static final String BRIDGE_ID = BRIDGE_CONFIG_PREFIX + "id";
+
     private KafkaConfig kafkaConfig;
     private AmqpConfig amqpConfig;
     private HttpConfig httpConfig;
@@ -83,5 +85,9 @@ public class BridgeConfig extends AbstractConfig {
                 ",amqpConfig=" + this.amqpConfig +
                 ",httpConfig=" + this.httpConfig +
                 ")";
+    }
+
+    public String getBridgeID() {
+        return config.get(BridgeConfig.BRIDGE_ID) == null ? null : config.get(BridgeConfig.BRIDGE_ID).toString();
     }
 }
--- a/src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
+++ b/src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
@@ -24,7 +24,7 @@ public class BridgeConfig extends AbstractConfig {
 
     /**
      * Constructor
-     * 
+     *
      * @param config bridge common configuration parameters map
      * @param kafkaConfig Kafka related configuration
      * @param amqpConfig AMQP endpoint related configuration
@@ -71,7 +71,7 @@ public class BridgeConfig extends AbstractConfig {
 
         return new BridgeConfig(map.entrySet().stream()
                 .filter(e -> e.getKey().startsWith(BridgeConfig.BRIDGE_CONFIG_PREFIX))
-                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)), 
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)),
                 kafkaConfig, amqpConfig, httpConfig);
     }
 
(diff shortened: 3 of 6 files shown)


bin/docker/kafka_bridge_config_generator.sh
```diff
@@ -103,13 +103,21 @@ amqp.messageConverter=${KAFKA_BRIDGE_AMQP_MESSAGE_CONVERTER}
...
+
 # if http/amqp is disabled, do not print its configuration
 PROPERTIES=$(cat <<EOF
 $KAFKA_PROPERTIES
 
 $PRODUCER_PROPERTIES
 
 $CONSUMER_PROPERTIES
+
+$BRIDGE_PROPERTIES
```
ppatierno: just a nit, maybe it's better having it at the top of the file

config/application.properties
```diff
@@ -19,4 +19,7 @@ amqp.messageConverter=io.strimzi.kafka.bridge.amqp.converter.AmqpDefaultMessageC
 #HTTP related settings
 http.enabled=true
 http.host=0.0.0.0
-http.port=8080
\ No newline at end of file
+http.port=8080
+
+#Bridge related settings
+bridge.id=my-bridge
```
ppatierno: ditto as above

src/main/java/io/strimzi/kafka/bridge/bridge/BridgeCommonConfig.java
```diff
@@ -0,0 +1,57 @@
...
+
+import io.strimzi.kafka.bridge.config.AbstractConfig;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Bridge related configuration
+ */
+public class BridgeCommonConfig extends AbstractConfig {
```
ppatierno: Why having a new class for that, isn't `BridgeConfig` enough for containing this info.
It has the Kafka config (with producer and consumer) and `T` endpoint config (so AMQP or HTTP).
It could have info like `id` which are bridge wide.

bin/docker/kafka_bridge_config_generator.sh
```diff
@@ -105,6 +111,8 @@ EOF
 
 # if http/amqp is disabled, do not print its configuration
 PROPERTIES=$(cat <<EOF
+BRIDGE_PROPERTIES
```
tombentley:
```suggestion
$BRIDGE_PROPERTIES
```

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -84,4 +86,8 @@ public String toString() {
                 ",httpConfig=" + this.httpConfig +
                 ")";
     }
+
+    public String getBridgeID() {
+        return config.get(BridgeConfig.BRIDGE_ID) == null ? null : config.get(BridgeConfig.BRIDGE_ID).toString();
```
tombentley: Using the ternary expression forces you into looking up the same value twice and in this case is harder to read. So better to use an if statement in this case.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -371,7 +373,7 @@ public void handle(Endpoint<?> endpoint, Handler<?> handler) {
         try {
             bodyAsJson = routingContext.getBodyAsJson();
         } catch (DecodeException ex) {
-            
+
```
tombentley: If you're going to ignore an exception you should at least add a comment about why that is the correct thing to do.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -84,4 +86,8 @@ public String toString() {
                 ",httpConfig=" + this.httpConfig +
                 ")";
     }
+
+    public String getBridgeID() {
```
tombentley: Why are we calling it "bridge id" when the corresponding key in the consumer JSON and in the `BridgeEndpoint` is called "name"? And the configured value is more of a prefix than a complete id/name, so the naming ought to reflect that, no?

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -43,7 +43,7 @@ public HttpSourceBridgeEndpoint(Vertx vertx, BridgeConfig bridgeConfig,
 
     @Override
     public void open() {
-        this.name = "kafka-bridge-producer-" + UUID.randomUUID();
+        this.name = this.bridgeConfig.getBridgeID() == null ? "kafka-bridge-producer-" + UUID.randomUUID() : this.bridgeConfig.getBridgeID() + "-" + UUID.randomUUID();
```
tombentley: Does anything actually use the name for the producer case?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
ppatierno: The "name" is for the endpoint which ends up in the consumer name. Actually, it's a bridge id/name but used as prefix for consumer name creation.

tombentley: So it's just "identifying" a consumer, not the bridge itself?

ppatierno: it identifies the bridge but is used as prefix for consumer names.

bin/docker/kafka_bridge_config_generator.sh
stanlyDoge: Good catch!

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
stanlyDoge: It seems I just removed some white spaces there. The code is from @ppatierno

ppatierno: The comment is just before the `try`

tombentley: How do you know that the body is empty (and thus that it's OK to ignore the exception, as you say)? Shouldn't the client not send a body, rather than sending an empty one?

ppatierno: I noticed that long time ago I guess that by empty I meant no body at all.

ppatierno:
@tombentley any more comments on this?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
tombentley: But that's kind of my point: How does this "identify the bridge"? AFAICS it is _only_ used as a prefix for consumer names. Thus is only indirectly identified the bridge and primarily identifies the consumers. Therefore calling it `consumerIdPrefix` is a lot closer to the truth of what this is used for (and thus more easily understood by users). Do you have plans for being able to identify the bridge in some other way, which will make the "bridge id" more meaningful?

ppatierno: I don't have right now, but it could be used in the future in that way. If we call it "consumerIdPrefix" today we are going to do more changes in the future if the bridge id could be useful.
What's the problem saying that it's a bridge identifier (which will be the `metadata.name` from the K8S resource) and it's also used as prefix for consumer names. What's wrong with that? :-)

bin/docker/kafka_bridge_config_generator.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+BRIDGE_PROPERTIES=$(cat <<-EOF
+#Bridge configuration
+bridge.id=${KAFKA_BRIDGE_ID}
```
ppatierno: one more comment, from where the `KAFKA_BRIDGE_ID` comes? do you need to do additional work in the cluster operator for providing it from the `metadata.name` of the `KafkaBridge` resource?

stanlyDoge: Yes, we will need to add this in the `KafkaBridgeCluster`.

ppatierno: Open an issue there and reference this PR please.

stanlyDoge: Addressed in https://github.com/strimzi/strimzi-kafka-operator/pull/1875

ppatierno:
I think we need at least a couple of tests for this.
When the bridgeId is set ... checking that on consumer creation, if the name is not supplied, then in the creation response the "instanceId" starts with the bridgeId (we cannot check the entire name because there is a UUID in there).
The other test is about not setting bridgeId so the consumer name (if not provided on creation) has to start with "kafka-bridge-consumer-".
Let's merge this one, I will open an issue for the tests.

ppatierno:
@stanlyDoge I opened #295 can you work on that?

PR #280, opened 2019-07-24
Refactoring to centralize all operations inside one handle method

ppatierno:
This PR just move all the operations handling in one handle method instead of having one for the consumer creation.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -21,6 +21,7 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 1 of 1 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -368,59 +364,54 @@ private void addConfigParameter(String key, String value, Properties props) {
     @Override
     public void handle(Endpoint<?> endpoint, Handler<?> handler) {
         RoutingContext routingContext = (RoutingContext) endpoint.get();
-        JsonObject bodyAsJson = routingContext.getBodyAsJson();
+        JsonObject bodyAsJson = null;
+        // TODO: it seems that getBodyAsJson raises an exception when the body is empty and not null
+        try {
+            bodyAsJson = routingContext.getBodyAsJson();
+        } catch (Exception ex) {
```
scholzj: Can we use some more specific exception here instead of silently catching everything?

PR #285, opened 2019-07-26
Added logging documentation

ppatierno:


laidan6000:
Did you mean to position this module within the _Strimzi Kafka Bridge Cluster Configuration_ section? If so, please add it to `assembly-deployment-configuration-kafka-bridge.adoc` .

Please use a more descriptive title, such as "Configuring logging for the Kafka Bridge".

ppatierno:
For sure this logging chapter is not part of the API reference but at same time it doesn't make sense to have it in the Strimzi documentation. Remember that people can use the bridge as standalone and in that case the logging documentation makes sense out of Strimzi (so out of how the bridge was deployed using Strimzi or not). So if this book is just about API, I think we should have another book for not API related bridge configuration?

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
+= Logging
+
+This section provides information on loggers and how to configure log levels.
```
laidan6000: Suggest removing this line.

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+The bridge allows to set a different log level for each of the operations defined by the related OpenAPI specification.
```
laidan6000:
```suggestion
The Kafka bridge allows you to set a different log level for each operation that is defined by the related OpenAPI specification. 
```

laidan6000: Insert a blank line (new paragraph).

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+Each operation matches a corresponding API endpoint on which the bridge gets requests from the HTTP clients.
```
laidan6000:
```suggestion
Each operation has a corresponding API endpoint through which the bridge receives requests from HTTP clients.
```

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+Changing the log level on each of them allows to have a more or less fine grained logging information about the incoming HTTP requests and the outgoing HTTP responses.
```
laidan6000:
```suggestion
You can change the log level on each endpoint to produce more or less fine-grained logging information about the incoming and outgoing HTTP requests.
```

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+The loggers are defined inside the `log4j.properties` file which by default has the following configuration for `healthy` and `ready` endpoints.
```
laidan6000:
```suggestion
Loggers are defined in the `log4j.properties` file, which has the following default configuration for `healthy` and `ready` endpoints:
```

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+```
+log4j.logger.http.openapi.operation.healthy=WARN, out
+log4j.additivity.http.openapi.operation.healthy=false
+log4j.logger.http.openapi.operation.ready=WARN, out
+log4j.additivity.http.openapi.operation.ready=false
+```
+
+All the other operations are by default at logging `INFO` level.
```
laidan6000:
```suggestion
The log level of all other operations is set to `INFO` by default.
```

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+The format for each logger is the following.
```
laidan6000:
```suggestion
Loggers are formatted as follows:
```

documentation/book/logging.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+```
+log4j.logger.http.openapi.operation.<operation-id>
+```
+
+Where `<operation-id>` is the identifier of the specific operation.
+Following the list of the operations defined by the OpenAPI specification.
```
laidan6000:
```suggestion
Following is the list of operations defined by the OpenAPI specification:
```

scholzj:
> For sure this logging chapter is not part of the API reference but at same time it doesn't make sense to have it in the Strimzi documentation. Remember that people can use the bridge as standalone and in that case the logging documentation makes sense out of Strimzi (so out of how the bridge was deployed using Strimzi or not). So if this book is just about API, I think we should have another book for not API related bridge configuration?

@ppatierno If this was on me ... yes exactly, I think it would be better to have separate book for the regular docs (in this repo, but separate). Especially if @laidan6000 want to link to the API docs from downstream, it would be better to have the API docs separate and not duplicate other content they will have downstream.

ppatierno:
Because for now, we have this kind of content (not API related) just on the Strimzi doc, I am going to close this PR using the same logging content to open a PR there.
Anyway we have to discuss further about having that content as new book here.

PR #289, opened 2019-07-29
ST: Factored out general methods for ST

serrss:
This is a draft PR. Work on it still in progress. 

1. Added Services as test resources to have code with ST in more readable form.
2. Created general methods for ST
3. Split out ST for Consumer and Topics

After this PR we will have DSL for Kafka Bridge ST what will make them more readable.
```
        consumerService(client)
            .groupId(groupId)
            .headers(headers)
            .jsonBody(consumerJson)
            .createConsumer(context);
```


src/test/java/io/strimzi/kafka/bridge/http/service/BaseService.java
```diff
@@ -0,0 +1,19 @@
+package io.strimzi.kafka.bridge.http.service;
+
+import io.vertx.ext.web.client.WebClient;
+
+
+public class BaseService <B extends BaseService> {
```
ppatierno: what's the purpose of this generic type B here? The same question for `ConsumerService` and `TopicService`, it seems you don't use these generic types (B, C, and T) in the tests.

src/test/java/io/strimzi/kafka/bridge/http/TopicsTest.java
```diff
@@ -0,0 +1,518 @@
...
+                    assertEquals("Validation error on: body.records[0] - $.records[0].value: is missing but it is required",
+                            error.getMessage());
+                });
+
+        topicService(client)
+            .headers(defaultHeaders(jsonObject))
+            .topicName(topic)
+            .jsonBody(jsonObject)
+            .sendRecord(context, handler);
+        context.completeNow();
```
ppatierno: the context should not be completed here but in the handler as before. It's an async test.
As you did now, the `completeNow` will be called before the handler is executed.
Or am I missing something?

src/test/java/io/strimzi/kafka/bridge/http/service/BaseService.java
serrss: oh, I forget to remove it from `BaseService`. We no need to have it. But as for `ConsumerService` and `TopicService` I use generics to have DSL for configuring and sending the requests

src/test/java/io/strimzi/kafka/bridge/http/TopicsTest.java
serrss: You are right, we need to have it somewhere in a block of `context.verify()`

PR #290, opened 2019-07-29
Consumer timeout

stanlyDoge:


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -84,6 +90,23 @@ private void bindHttpServer(Future<Void> startFuture) {
...
+            log.debug("Checking for stale customers {}", timestampMap.size());
+            Iterator it = timestampMap.entrySet().iterator();
+            while (it.hasNext()) {
+                Map.Entry<String, Long> item = (Map.Entry) it.next();
+                if (item.getValue() + this.bridgeConfig.getHttpConfig().getConsumerTimeout() < System.currentTimeMillis()) {
+                    final SinkBridgeEndpoint deleteSinkEndpoint = this.httpBridgeContext.getHttpSinkEndpoints().get(item.getKey());
+                    if (deleteSinkEndpoint != null) {
+                        this.httpBridgeContext.getHttpSinkEndpoints().remove(item.getKey());
+                        deleteSinkEndpoint.close();
+                        log.warn("User {} deleted after timeout {} ms", item.getKey(), this.bridgeConfig.getHttpConfig().getConsumerTimeout());
```
ppatierno: It's "consumer", not "User" and actually, the value from the map is not the timeout.

config/application.properties
```diff
@@ -19,4 +19,5 @@ amqp.messageConverter=io.strimzi.kafka.bridge.amqp.converter.AmqpDefaultMessageC
 #HTTP related settings
 http.enabled=true
 http.host=0.0.0.0
-http.port=8080
\ No newline at end of file
+http.port=8080
+http.timeout=60000000
```
ppatierno: if it's the default we can avoid to update this file.

src/main/resources/application.properties
```diff
@@ -19,4 +19,5 @@ amqp.messageConverter=io.strimzi.kafka.bridge.amqp.converter.AmqpDefaultMessageC
 #HTTP related settings
 http.enabled=true
 http.host=0.0.0.0
-http.port=8080
\ No newline at end of file
+http.port=8080
+http.timeout=60000000
```
ppatierno: if it's the default we can avoid to update this file.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
stanlyDoge: I am not using value from the map, but from the configuration.

ppatierno: Ah right ... I am blind :-)

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -84,6 +90,23 @@ private void bindHttpServer(Future<Void> startFuture) {
...
+                if (item.getValue() + this.bridgeConfig.getHttpConfig().getConsumerTimeout() < System.currentTimeMillis()) {
+                    final SinkBridgeEndpoint deleteSinkEndpoint = this.httpBridgeContext.getHttpSinkEndpoints().get(item.getKey());
+                    if (deleteSinkEndpoint != null) {
+                        this.httpBridgeContext.getHttpSinkEndpoints().remove(item.getKey());
+                        deleteSinkEndpoint.close();
+                        log.warn("Consumer {} deleted after inactivity timeout ({} ms).", item.getKey(), this.bridgeConfig.getHttpConfig().getConsumerTimeout());
+                        timestampMap.remove(item.getKey());
+                    }
+                }
+            }
```
ppatierno: I would move this timer set above only if the HTTP bridge is started successfully otherwise it's useless.
To avoid more code difficult to read I would factor out the timer logic in a private method as well.

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
```diff
@@ -20,10 +20,12 @@
     public static final String HTTP_ENABLED = HTTP_CONFIG_PREFIX + "enabled";
     public static final String HTTP_HOST = HTTP_CONFIG_PREFIX + "host";
     public static final String HTTP_PORT = HTTP_CONFIG_PREFIX + "port";
+    public static final String HTTP_CONSUMER_TIMEOUT = HTTP_CONFIG_PREFIX + "timeout";
 
     public static final boolean DEFAULT_HTTP_ENABLED = true;
     public static final String DEFAULT_HOST = "0.0.0.0";
     public static final int DEFAULT_PORT = 8080;
+    public static final long DEFAULT_CONSUMER_TIMEOUT = 60 * 1000 * 1000;
```
ppatierno: Still in doubt about having 1 hour as default timeout ... @tombentley wdyt?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3399,110 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
         context.completeNow();
         assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
     }
+
+    @Test
+    void consumerDeletedAfterInactivity(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        String topic = "consumerDeletedAfterInactivity";
+        kafkaCluster.createTopic(topic, 1, 1);
```
ppatierno: I don't think that we need all the complexity of creating a topic, subscribing, etc etc ...
We have just to create the consumer because the timeout checker will check anyway; we can just simplify the test code in this way.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3399,110 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
...
+                        assertEquals(HttpResponseStatus.NOT_FOUND.code(), response.statusCode());
+                        assertEquals(HttpResponseStatus.NOT_FOUND.code(), error.getCode());
+                        assertEquals("The specified consumer instance was not found.", error.getMessage());
+                    });
+                    subscribeAfterTimeout.complete(true);
+                });
+        subscribeAfterTimeout.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+        context.completeNow();
+
+        this.bridgeConfig.getConfig().remove(HttpConfig.HTTP_CONSUMER_TIMEOUT);
```
ppatierno: do you need this? I don't think that the other tests don't have activity for 5 seconds.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3399,110 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
...
+
+                    create.complete(true);
+                });
+
+        create.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+
+        // After
+        Thread.sleep(timeout - 1000);
+
+        //////////////////
```
ppatierno: where are you checking about the consumer deleted after 5 seconds? it's not clear to me, sorry :-(

stanlyDoge: The test does this scenario
1. create consumer
2. wait a time < timeout
3. try to subscribe - this should be successful only if the consumer was found
4. wait a time > timeout
5. try to subscribe - this should fail because consumer was not found (it is deleted after timeout)

ppatierno: so you are calling sleep in the main event loop so blocking the Vert.x event loop ... it's clear why you choose 5 secods as timeout because it's the timeout for Vert.x event loop as well so you wanted to avoid warning messages from Vert.x :-).
What I would do is ...

1. consumer creation
2. create a vert.x timer, if the creation was ok, running after the timeout
3. try to delete the consumer and check that it fails because there is no consumer with that name

wdyt?

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
ppatierno: Regarding the good point about disabling the timeout.
The way that makes more sense to me is just not having the `http.timeout` property in the configuration. Actually, it doesn't work, because not putting a property means using the default value which is how all the other parameters work, so I would leave in that way.
The solutions could be just relying on the user, documenting it, that for "disabling" he should se a pretty high value there or using a special value as 0 (some tools/library disable timing related features in this way). Wdyt? @stanlyDoge @tombentley

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -84,6 +90,23 @@ private void bindHttpServer(Future<Void> startFuture) {
                         startFuture.fail(httpServerAsyncResult.cause());
                     }
                 });
+
+        vertx.setPeriodic(1000, a -> {
```
tombentley: If the timeout is high then checking for expiry every second seems rather pointless.

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
tombentley: I think -1 is more obviously a special value than 0 (even though 0 doesn't have a sensible meaning).

Also @stanlyDoge as a matter of habit when setting a `long`-typed variable to a result of a multiplication like this it is always a good idea to say `60L * 1000L * 1000L` so the expression is evaluated using longs. Mostly you get away with it, but that makes those occasions where the result overflows `int` atrithmetic and is then widened into a `long` all the more confusing.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -77,6 +83,10 @@ private void bindHttpServer(Future<Void> startFuture) {
                                         .get(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG)
                         );
 
+                        if (this.bridgeConfig.getHttpConfig().getConsumerTimeout() > -1) {
+                            startInactiveConsumerDeletionTimer(this.bridgeConfig.getHttpConfig().getConsumerTimeout());
```
ppatierno: I think we have a problem with this condition. If the user sets 0 as timeout we'll have consumers removed after the first timer tick (1 sec).
The condition in the timer handler:

`if (item.getValue() + timeout < System.currentTimeMillis())`

will always be true (the consumer timestamp was put in the map before the first tick timer ever and because timeout is 0, it will always be less than the current time). So we are goint to remove the consumer after max one sec (or even less, it depends on the time alignment with the timer when the consumer is created and the timestamp is put into the map). I am still thinking that we should not allow negavite values (as -1) and using 0 for disabling while missing the `http.timeout` property in the configuration means using the default value (as for the other properties). @stanlyDoge @tombentley any more thoughts?

tombentley: I'm not sure that helps, because the user can still set the timeout as, say, 1 with the terrible consequence that the consumer gets removed after 1 second or less. 

We could say that the config is either a strictly positive integer or the word `off`, which avoids any confusion around having a special value.

Also. I do hope the config will be `http.timeoutMs` or (even better) `http.timeoutSeconds` (since ms precision just isn't warranted), rather than the units-ambiguous `http.timeout`.

ppatierno: > I'm not sure that helps, because the user can still set the timeout as, say, 1 with the terrible consequence that the consumer gets removed after 1 second or less.

In this case it's a user mistake, isn't it? You should expect that if you set timeout to 1 sec.

> We could say that the config is either a strictly positive integer or the word off, which avoids any confusion around having a special value.

Tbh I don't like the Kafka way to have `acks` which can be 0, 1 or "all" (mixing numbers and words). I would stay with a strict positive number or 0 to disable ...

Finally, I agree on the parameter name. I would go for `timeoutSeconds`

tombentley: I just don't think that saying timeout=0 means no timeout is at all intuitive. `-1` (and `off` even more so) is more obviously not a normal value. By all means make 0 an error and bail out on start up, but giving it an unintuitive meaning is just going to confuse people.

ppatierno: At this point I think we need one more comment from a maintainer, @scholzj your opinion is important here :-)

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
scholzj: I guess `-1` makes more sense. You can also add another boolean option which would be true by default and would be enabling / disabling the timeout.

ppatierno: Now you provided an opinion on the suggested values but then adding a new option :-) Let's go with -1 as this is ok for you and Tom. @stanlyDoge now you have a quorum!

scholzj:
@ppatierno and @tombentley seem to be already reviewing this ... so I leave it up to them :-o

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -86,6 +96,26 @@ private void bindHttpServer(Future<Void> startFuture) {
                 });
     }
 
+    private void startInactiveConsumerDeletionTimer(Long timeout) {
+        Long timeoutInMs = timeout * 1000L;
+        vertx.setPeriodic(1000, ignore -> {
```
tombentley: I ask again, does this really need to run every second?  Why 1 second?

stanlyDoge: I used the polling value we are using for another operation. 
Of course, we can use another value but what it should be? Should it be related to the timeout?

tombentley: I would use something like ½ the timeout. Remember, this is a defence against badly written clients, so it doesn't matter if a client consumer is closed "late" (i.e. after the timeout), so long as it is closed. We also expect that this actually ends up closing a consumer only rarely, so it's not worth spending cycles on trying to be accurate.

Issue #286, opened 2019-07-29
Add tests on the liveness and readiness endpoint


ppatierno:
Add tests for `/healthy` and `/ready` endpoints.

PR #293, opened 2019-07-31
Readiness and liveness test

stanlyDoge:
https://github.com/strimzi/strimzi-kafka-bridge/issues/286

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
@@ -7,6 +7,7 @@ package io.strimzi.kafka.bridge.http;
 
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.strimzi.kafka.bridge.BridgeContentType;
+import io.strimzi.kafka.bridge.HealthChecker;
 import io.strimzi.kafka.bridge.KafkaClusterTestBase;
 import io.strimzi.kafka.bridge.amqp.AmqpConfig;
 import io.strimzi.kafka.bridge.config.BridgeConfig;
@@ -59,7 +60,7 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @ExtendWith(VertxExtension.class)
-@SuppressWarnings({"checkstyle:JavaNCSS"})
+@SuppressWarnings({"checkstyle:JavaNCSS", "checkstyle:ClassDataAbstractionCoupling"})
 class HttpBridgeTest extends KafkaClusterTestBase {
 
     private static final Logger log = LoggerFactory.getLogger(HttpBridgeTest.class);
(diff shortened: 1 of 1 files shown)


src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3397,35 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
         context.completeNow();
         assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
     }
+
+    @Test
+    void readyHealthyTest(VertxTestContext context) throws InterruptedException {
+        int iterations = 5;
+        httpBridge.setHealthChecker(new HealthChecker());
```
ppatierno: can't it be called in the "before" method?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3397,35 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
         context.completeNow();
         assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
     }
+
+    @Test
+    void readyHealthyTest(VertxTestContext context) throws InterruptedException {
```
ppatierno: I would prefer to have two different tests

ppatierno:
Are you talking about the `status` subresources in the `Kafka` resource?
For now it's just checking an internal variable set when it starts but, of course, it's not a real health check mechanism.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3398,40 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
...
+                    .send(ar -> {
+                        context.verify(() -> {
+                            assertTrue(ar.succeeded());
+                            assertEquals(HttpResponseStatus.OK.code(), ar.result().statusCode());
+                        });
+                    });
+            Thread.sleep(1000);
+            if (i == iterations) {
+                context.completeNow();
+            }
```
ppatierno: because it's all async it could happen that if the last call to the `/ready` endpoint takes more than 2 seconds, you are completing the context before the actual outcome from the GET request.
it means that you have to do the check inside the handler.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3396,4 +3398,40 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
...
```
ppatierno: ditto as above.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -59,7 +60,7 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @ExtendWith(VertxExtension.class)
-@SuppressWarnings({"checkstyle:JavaNCSS"})
+@SuppressWarnings({"checkstyle:JavaNCSS", "checkstyle:ClassDataAbstractionCoupling"})
```
ppatierno: adding the `HealthChecker` was the need for this new suppress warning rule?

stanlyDoge: Yes.

PR #294, opened 2019-08-05
ST: Tests by groups, and general methods for ST

serrss:
Changes in PR:
1) STs were split by groups (Consumer, Producer, Seek)
2) Services with general methods were added (the main idea to make code in DSL style and remove duplicates)
```
// Create a consumer and subscribe in 3 lines
        consumerService()
                .createConsumer(context, groupId, consumerJson)
                .subscribeConsumer(context, groupId, name, topic);
```
3) Hardcode were updated by constants (Headers, Urls, etc)

TODO:
- [ ] Move hardcode with Urls to class `Urls` (maybe make it as Enum)
- [ ] Add Java docs for public methods
- [ ]


src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
```diff
@@ -0,0 +1,193 @@
...
+    public HttpRequest<JsonObject> offsetsRequest(String groupId, String name, JsonObject json) {
+        return postRequest(Urls.consumerInstancesOffsets(groupId, name))
+                .putHeader(CONTENT_LENGTH, String.valueOf(json.toBuffer().length()))
+                .putHeader(CONTENT_TYPE, BridgeContentType.KAFKA_JSON)
+                .as(BodyCodec.jsonObject());
+    }
+
+    // Consumer actions
+
+    public ConsumerService deleteConsumerSubscription(VertxTestContext context, String groupId, String name, String... topicNames) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno: technically it's an "unsubscribe" request, I would not call it as a "delete". Anyway, it's going to use the HTTP DELETE method of course.

src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
```diff
@@ -0,0 +1,193 @@
...
+                .sendJsonObject(topicsRoot, ar -> {
+                    context.verify(() -> {
+                        assertTrue(ar.succeeded());
+                        assertEquals(HttpResponseStatus.NO_CONTENT.code(), ar.result().statusCode());
+                        context.completeNow();
+                    });
+                    unsubscribe.complete(true);
+                });
+
+        unsubscribe.get(HTTP_REQUEST_TIMEOUT, TimeUnit.SECONDS);
```
ppatierno: I am not pretty sure about this pattern here ... calling the complete on the context and another complete on the unsubscribe future.

src/test/java/io/strimzi/kafka/bridge/http/services/ProducerService.java
```diff
@@ -0,0 +1,36 @@
...
+                .putHeader(CONTENT_TYPE, BridgeContentType.KAFKA_JSON_JSON)
+                .as(BodyCodec.jsonObject());
+    }
+
+    public HttpRequest<JsonObject> sendRecordsToPartitionRequest(String topic, Object partition, JsonObject jsonObject) {
+        return postRequest(Urls.producerTopicsPartitions(topic, partition))
+                .putHeader(CONTENT_LENGTH, String.valueOf(jsonObject.toBuffer().length()))
+                .putHeader(CONTENT_TYPE, BridgeContentType.KAFKA_JSON_JSON)
+                .as(BodyCodec.jsonObject());
+    }
```
ppatierno: even for producer, it could not be always json but even binary so the content-type header will be different.

src/test/java/io/strimzi/kafka/bridge/http/services/SeekService.java
```diff
@@ -0,0 +1,45 @@
...
+                .putHeader(CONTENT_TYPE, BridgeContentType.KAFKA_JSON)
+                .as(BodyCodec.jsonObject());
+    }
+
+    public HttpRequest<JsonObject> positionsBeginningEnd(String groupId, String name, JsonObject json) {
+        return postRequest(Urls.consumerInstancesPositionsEnd(groupId, name))
+                .putHeader(CONTENT_LENGTH, String.valueOf(json.toBuffer().length()))
+                .putHeader(CONTENT_TYPE, BridgeContentType.KAFKA_JSON)
+                .as(BodyCodec.jsonObject());
+    }
```
ppatierno: these three methods are pretty the same. The only different thing is the URL. I guess you can factor out a common method to be called by them.

src/test/java/io/strimzi/kafka/bridge/utils/Urls.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final String POSITIONS_PATH =  "/positions";
+    private static final String SUBSCRIPTION_PATH =  "/subscription";
+    private static final String TOPICS_PATH = "/topics/";
+    private static final String PARTITIONS_PATH = "/partitions/";
+    private static final String ASSIGMENTS_PATH = "/assignments";
+    private static final String OFFSETS_PATH = "/offsets";
+    private static final String RECORDS_PATH = "/records";
+
+    public static String consumer(String groupId) {
+        return TRANSFER_PROTOCOL + BRIDGE_HOST + ":" + BRIDGE_PORT + CONSUMERS_PATH + groupId;
```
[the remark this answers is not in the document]
serrss: Fixed

src/test/java/io/strimzi/kafka/bridge/utils/Urls.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public static String consumerInstances(String groupId, String name) {
```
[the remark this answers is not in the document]
serrss: Yes, but I called it because of wanting to have the same names as in the address, for example, address `/consumers/{groupid}/instances/{name}` have names in the plural. SO maybe it's better to have methods name in plural also? @ppatierno WDYT?

src/test/java/io/strimzi/kafka/bridge/utils/Urls.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public static String consumerInstancesOffsets(String groupId, String name) {
+        return consumerInstances(groupId, name) + OFFSETS_PATH;
+    }
+
+    public static String consumerInstancesRecords(String groupId, String name, Integer timeout, Integer maxBytes) {
+        return consumerInstances(groupId, name) + RECORDS_PATH
+                + "?"
+                + (timeout != null ? "timeout=" + timeout : "")
+                + (maxBytes != null  ? "max_bytes=" + maxBytes : "");
```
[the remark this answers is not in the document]
serrss: right. right, and this is why ST failed for me, thanks @ppatierno

src/test/java/io/strimzi/kafka/bridge/utils/Urls.java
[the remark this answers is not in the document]
ppatierno: No it's not in this way. The path represents a resource in the REST API world. You are getting a consumer instance named {name} in the {groupid}. You see "consumers" and "instances" because you are accessing at a group (in consumers) and a specific name instance (in instances). So you are actually getting a single thing, not a collection.

src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
serrss: Good point, I guess it will be enough just call `context.awaitCompletion(HTTP_REQUEST_TIMEOUT, TimeUnit.SECONDS)` instead of having CompletableFuture in these methods.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
```diff
@@ -0,0 +1,97 @@
...
+        return new ConsumerService(client);
+    }
+
+    SeekService seekService() {
+        return new SeekService(client);
+    }
+
+    ProducerService producerService() {
+        return new ProducerService(client);
+    }
```
ppatierno: about the above `xxxService` methods.
Every time you call one of this (and you do that even more times during a single test) you are creating a new `xxxService` instance and always with the same client.
The simpler solution could be calling just one time an `xxxService' method in each test and save a reference to the returned object to avoid creating a new one later.
The other solution, which could be better imho, could be having this `xxxService` as singleton?

src/test/java/io/strimzi/kafka/bridge/http/DefaultTest.java
```diff
@@ -0,0 +1,135 @@
...
+                        assertTrue(paths.containsKey("/openapi"));
+                        assertEquals(HttpOpenApiOperations.OPENAPI.toString(), bridgeResponse.getJsonObject("paths").getJsonObject("/openapi").getJsonObject("get").getString("operationId"));
+                        assertFalse(paths.containsKey("/karel"));
+                        assertEquals(15, bridgeResponse.getJsonObject("definitions").getMap().size());
+                        assertEquals(4, bridgeResponse.getJsonArray("tags").size());
+                    });
+                    context.completeNow();
+                });
+    }
+}
```
ppatierno: if you are grouping I would not leave these tests in the base class. Maybe the first two are about "healthy". The openapi one ... I don't have a strong opinion.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
serrss: Yes, sure, having singleton is a good idea here

src/test/java/io/strimzi/kafka/bridge/http/DefaultTest.java
serrss: Sorry @ppatierno, but we can't have tests in base class because of having child classes. Every child class will execute these ST and we will have 4-5 excess-runs for each ST

ppatierno: My mistake I was seeing the wrong file. ;)

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
```diff
@@ -0,0 +1,97 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@ExtendWith(VertxExtension.class)
+@SuppressWarnings({"checkstyle:JavaNCSS"})
+class HttpBridgeTestBase extends KafkaClusterTestBase {
+
+    private static final Logger log = LoggerFactory.getLogger(HttpBridgeTestBase.class);
```
ppatierno: this seems not be used. I guess we are going to use it for logging during tests, you should make it `protected` then

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
serrss: After trying to implement singleton here I found blocker for this approach... we initialize `WebClient` instance for every test method in `before` method and because of the rule that xxxService should use `WebClient` we need to recreate it for every new method. @ppatierno Can we move code from `before` method to static block?

serrss: Looks like moving `before` to static block is not possible because of having code ``` vertx.deployVerticle(this.httpBridge, context.succeeding(id -> context.completeNow()));``` where we use `VertxTestContext`
@ppatierno any other ideas?

ppatierno: But I don't think that we need a new WebClient for each test. Even that client can be a singleton :-)

serrss: But, the client uses `Vertx` during initialization ```WebClient.create(vertx, new WebClientOptions()...``` which completes after the execution of test method. Or I don't understand the hint.

ppatierno: so the WebClient depends on Vert.x instance which is create as new instance before each test which is something we don't really need.
We should try to create the Vert.x instance once for all the tests (in a @BeforeAll method)).
In this way I think all the services and the WebClient can be singleton.
The httpbridge creation and the verticle deployment should stay in the `HttpBridgeTestBase.before` method (it's possible that we want to change something on the bridge for each test). Wdyt? Is it feasible?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -0,0 +1,1116 @@
...
+        String topic1 = "receiveSimpleMessage-1";
+        String topic2 = "receiveSimpleMessage-2";
+        kafkaCluster.createTopic(topic1, 1, 1);
+        kafkaCluster.createTopic(topic2, 1, 1);
+
+        String sentBody1 = "Simple message-1";
+        String sentBody2 = "Simple message-2";
+
+        CompletableFuture<Boolean> produce = new CompletableFuture<>();
+        kafkaCluster.useTo().produce("", 1, new KafkaJsonSerializer(), new KafkaJsonSerializer(),
```
Frawless: In general, there should be `CompletableFuture` for each producer, this should be fixed. I suggest to factor out native kafka producer/consumer piece of code to methods

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -0,0 +1,1116 @@
...
+    }
+
+    @Test
+    void receiveSimpleMessageFromMultiplePartitions(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        String topic = "receiveSimpleMessageFromMultiplePartitions";
+        kafkaCluster.createTopic(topic, 2, 1);
+
+        String sentBody = "Simple message from partition";
+
+        CompletableFuture<Boolean> produce = new CompletableFuture<>();
```
Frawless: Same as above

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
serrss: @ppatierno Finally, singleton was implemented, look the last commit plz

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
```diff
@@ -0,0 +1,99 @@
...
+
+    static Map<String, Object> config = new HashMap<>();
+
+    static {
+        config.put(AmqpConfig.AMQP_ENABLED, true);
```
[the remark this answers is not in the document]
serrss: Good point, I guess we no need it. All ST passed without this line

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
ppatierno: Good to know! You mentioned that you had a lot of failures initializing Vert.x instance just one time at the beginning of all tests.

serrss:
@Frawless @ppatierno All comments to this PR were resolved, could you plz review it again?

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -0,0 +1,353 @@
+/*
+ * Copyright 2018, Strimzi authors.
```
ppatierno: new file, let's use 2019.

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -0,0 +1,353 @@
...
+public class SeekTest extends HttpBridgeTestBase {
+
+    private String name = "my-kafka-consumer";
+    private String groupId = "my-group";
+    private JsonObject jsonConsumer = new JsonObject()
+        .put("name", name)
+        .put("format", "json");
+
+    @Test
+    void consumerOrPartitionNotFound(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno: let's call it `seekToBeginningConsumerOrPartitionNotFound`

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -0,0 +1,353 @@
...
+        String topic = "notFoundToBeginningAndReceive";
```
ppatierno: let's use the test method name as the other tests, so "seekToBeginningConsumerOrPartitionNotFound"

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -0,0 +1,353 @@
...
+                    assertEquals(0, body.size());
+                    consumeSeek.complete(true);
+                });
+
+        consumeSeek.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+
+        CompletableFuture<Boolean> delete = new CompletableFuture<>();
+        // consumer deletion
+        consumerService()
+            .deleteRequest(Urls.consumerInstance(groupId, name))
```
ppatierno: if we have to delete the consumer, why not using the `consumerService().deleteConsumer(context, groupId, name);` as for all the other consumer tests?

src/test/java/io/strimzi/kafka/bridge/http/services/BaseService.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2018, Strimzi authors.
```
ppatierno: new file, let's use 2019.

src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
```diff
@@ -0,0 +1,197 @@
+/*
+ * Copyright 2018, Strimzi authors.
```
ppatierno: new file, let's use 2019.

src/test/java/io/strimzi/kafka/bridge/http/services/ProducerService.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2018, Strimzi authors.
```
ppatierno: new file, let's use 2019.

src/test/java/io/strimzi/kafka/bridge/http/services/SeekService.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2018, Strimzi authors.
```
ppatierno: new file, let's use 2019.

src/test/java/io/strimzi/kafka/bridge/utils/Urls.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2018, Strimzi authors.
```
ppatierno: new file, let's use 2019.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+public class ConsumerSubscriptionTest extends HttpBridgeTestBase {
+
+    @Test
+    void unsubscribeConsumerNotFound(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        String topic = "unsubscribeConsumerNotFound";
+        kafkaCluster.createTopic(topic, 1, 1);
+
+        String sentBody = "Simple message";
+        kafkaCluster.produceStrings(topic, sentBody, 1, 0);
```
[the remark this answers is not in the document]
serrss: ye, you are right, no need to have it

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -0,0 +1,182 @@
...
+        JsonArray topics = new JsonArray();
+        topics.add(topic);
+
+        JsonObject topicsRoot = new JsonObject();
+        topicsRoot.put("topics", topics);
+        // create consumer
+        // subscribe to a topic
+        consumerService()
+            .createConsumer(context, groupId, json)
+            .subscribeConsumer(context, groupId, name, topicsRoot);
```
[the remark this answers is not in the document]
serrss: Okay, sounds reasonable. I guess we also can skip topic creation in this ST, right? @ppatierno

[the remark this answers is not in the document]
ppatierno: From a theoretical point of view, topic creation is needed because you cannot subscribe to a not existing topic.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -0,0 +1,1063 @@
...
+        // consumer deletion
+        consumerService()
+            .deleteConsumer(context, groupId, name);
+        context.completeNow();
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+    }
+
+
+    @Test
+    void postToNonexistentEndpoint(VertxTestContext context) {
```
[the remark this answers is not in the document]
ppatierno: Good catch. It should go maybe in the `DefaultTest` class. At this point, I would change the name to this class, what "Default" means in this context? It should be "AnyOtherTestsNotRelatedToConsumerProducerAndSeek" :-D
Just kidding ... any idea about a valid name?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -0,0 +1,1063 @@
...
+                        assertNull(key);
+                    });
+                    consume.complete(true);
+                });
+
+        consume.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+
+        CompletableFuture<Boolean> commit = new CompletableFuture<>();
+        baseService()
+            .postRequest(baseUri + "/offsets")
```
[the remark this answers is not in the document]
serrss: The base service here because we don't use any headers here and don't send any JSON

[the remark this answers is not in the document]
ppatierno: so we should have a method in the consumer service allowing to do that because it's a valid request from a consumer.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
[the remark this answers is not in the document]
serrss: Well, I just used the same pattern that we have in Swagger UI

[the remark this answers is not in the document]
serrss: `OtherServicesTest` is it sounds good to you folks? @Frawless @ppatierno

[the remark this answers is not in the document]
Frawless: Ye, this sounds good to me

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -31,6 +32,7 @@
     private static final String FORWARDED = "Forwarded";
     private static final String X_FORWARDED_HOST = "X-Forwarded-Host";
     private static final String X_FORWARDED_PROTO = "X-Forwarded-Proto";
+    private static long timeout = 5L;
```
ppatierno: Can we avoid to have this declared twice? It's in the HttpBridgeTestBase class as well.

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -0,0 +1,590 @@
...
+
+        JsonObject root = new JsonObject();
+        root.put("records", records);
+
+        producerService().sendRecordsRequest(topic, root)
+            .sendJsonObject(root, verifyOK(context));
+
+        Properties config = kafkaCluster.useTo().getConsumerProperties("groupId", null, OffsetResetStrategy.EARLIEST);
+
+        KafkaConsumer<String, String> consumer = KafkaConsumer.create(this.vertx, config,
```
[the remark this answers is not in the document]
serrss: Will be factored out in next PR

PR #296, opened 2019-08-09
Test for consumer name prefix (bridge ID)

stanlyDoge:
I also changed the cleaning of the consumers. It does not throw concurrent modification exception now.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -161,15 +161,17 @@ public class HttpBridge extends AbstractVerticle implements HealthCheckable {
 
         this.isReady = false;
 
-        //Consumers cleanup
-        this.httpBridgeContext.getHttpSinkEndpoints().forEach((consumerId, httpEndpoint) -> {
-            if (httpEndpoint != null) {
-                httpEndpoint.close();
-            }
-        });
+        // Consumers cleanup
+        Map<String, SinkBridgeEndpoint> consumers = this.httpBridgeContext.getHttpSinkEndpoints();
+
+        for (Map.Entry<String, SinkBridgeEndpoint> consumer: consumers.entrySet()) {
+            if (consumer.getValue() != null)
+                consumer.getValue().close();
+        }
+
         this.httpBridgeContext.getHttpSinkEndpoints().clear();
 
-        //producer cleanup
+        // producer cleanup
         // for each connection, we have to close the connection itself but before that
         // all the sink/source endpoints (so the related links inside each of them)
         this.httpBridgeContext.getHttpSourceEndpoints().forEach((connection, endpoint) -> {
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -163,13 +163,11 @@ public class HttpBridge extends AbstractVerticle implements HealthCheckable {
 
         // Consumers cleanup
         this.httpBridgeContext.closeAllSinkBridgeEndpoints();
-        this.httpBridgeContext.getHttpSinkEndpoints().clear();
 
         // producer cleanup
         // for each connection, we have to close the connection itself but before that
         // all the sink/source endpoints (so the related links inside each of them)
         this.httpBridgeContext.closeAllSourceBridgeEndpoints();
-        this.httpBridgeContext.getHttpSourceEndpoints().clear();
 
         if (this.httpServer != null) {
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -131,15 +133,17 @@ public void stop(Future<Void> stopFuture) throws Exception {
...
-        });
+        // Consumers cleanup
+        Map<String, SinkBridgeEndpoint> consumers = this.httpBridgeContext.getHttpSinkEndpoints();
+
+        for (Map.Entry<String, SinkBridgeEndpoint> consumer: consumers.entrySet()) {
+            if (consumer.getValue() != null)
+                consumer.getValue().close();
+        }
+
         this.httpBridgeContext.getHttpSinkEndpoints().clear();
```
tombentley: Why don't we have a method on the `httpBridgeContext` to closeSinkEndpoints()? Because clearing another object's Map like this seems be a violation of encapsulation.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -67,10 +67,12 @@
 
     private static Map<String, Object> config = new HashMap<>();
 
+    private static String bridgeID = "franta";
```
ppatierno: "franta"?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3486,4 +3488,30 @@ void openapiTest(VertxTestContext context) {
...
+                .putHeader("Content-type", BridgeContentType.KAFKA_JSON)
+                .as(BodyCodec.jsonObject())
+                .sendJsonObject(json, ar -> {
+                    context.verify(() -> {
+                        assertTrue(ar.succeeded());
+                        HttpResponse<JsonObject> response = ar.result();
+                        assertEquals(HttpResponseStatus.OK.code(), response.statusCode());
+                        JsonObject bridgeResponse = response.body();
+                        String consumerInstanceId = bridgeResponse.getString("instance_id");
+                        assertTrue(consumerInstanceId.startsWith(bridgeID));
```
ppatierno: can we avoid the `bridgeID` variable and just using `config.get(BridgeConfig.BRIDGE_ID` ?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3486,4 +3488,30 @@ void openapiTest(VertxTestContext context) {
...
     }
+
+    @Test
+    void createConsumerWithGeneratedName(VertxTestContext context) {
+        String groupId = "my-group";
+
+        JsonObject json = new JsonObject();
+        json.put("auto.offset.reset", "earliest");
+        json.put("enable.auto.commit", "true");
+        json.put("fetch.min.bytes", "100");
```
ppatierno: I don't think this test needs other parameters for consumer creation. Let's keep it simple.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTest.java
```diff
@@ -3486,4 +3488,30 @@ void openapiTest(VertxTestContext context) {
                 context.completeNow();
             });
     }
+
+    @Test
+    void createConsumerWithGeneratedName(VertxTestContext context) {
```
ppatierno: We also need a test to check the instanceid when the bridgeid is not provided.

ppatierno: After our offline discussion, let's leave this test for a future PR when the tests refactoring will be done. It's because we need to start the bridge with a different configuration, actually with no bridgei

stanlyDoge: We agreed to wait till https://github.com/strimzi/strimzi-kafka-bridge/pull/294 is merged

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tombentley: @stanlyDoge ^^

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -131,23 +131,14 @@ public void stop(Future<Void> stopFuture) throws Exception {
...
 
-        //Consumers cleanup
-        this.httpBridgeContext.getHttpSinkEndpoints().forEach((consumerId, httpEndpoint) -> {
-            if (httpEndpoint != null) {
-                httpEndpoint.close();
-            }
-        });
+        // Consumers cleanup
+        this.httpBridgeContext.closeAllSinkBridgeEndpoints();
         this.httpBridgeContext.getHttpSinkEndpoints().clear();
```
tombentley: The clear should be part of `closeAllSinkBridgeEndpoints`, since it doesn't make sense to leave endpoints in the map which have been closed. Likewise for the source endpoints below.

PR #302, opened 2019-08-19
Add a way how to list subscribed topics

### LINKED ISSUE #298: Adding operation for getting current subscribed topics
On the `/consumers/{groupid}/instances/{name}/subscription` endpoint should be possible to make a GET (HTTP request) in order to get the currently subscribed topics for the consumer.
It's not an administrative API because it's provided by the underlying native Kafka consumer client via the `assignment` method (which provides topics with related partitions).
The GET should Accept `application/vnd.kafka.v2+json` and the bridge should return something like:

```shell
HTTP/1.1 200 OK
Content-Type: application/vnd.kafka.v2+json

{
  "topics": [
    "my-topic1",
    "my-topic2"
  ]
}
```

stanlyDoge:
Fixes #298 

Created a topic with 5 partitions, subscribed a consumer via the bridge, then added a new consumer via console.
Before
```
{
    "topics": [
        "test"
    ],
    "partitions": [
        {
            "test": [
                1,
                0,
                3,
                2,
                4
            ]
        }
    ]
}
```

After
```
{
    "topics": [
        "test"
    ],
    "partitions": [
        {
            "test": [
                3,
                4
            ]
        }
    ]
}
```

### DIFF
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -774,7 +774,7 @@ __required__|List of topics to which the consumer will subscribe.|<<_topics,Topi
 === GET /consumers/{groupid}/instances/{name}/subscription
 
 ==== Description
-Retrieves a list of the topics to which is the consumer subscribed.
+Retrieves a list of the topics to which the consumer is subscribed.
 
 
 ==== Parameters
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -770,6 +770,78 @@ __required__|List of topics to which the consumer will subscribe.|<<_topics,Topi
 ----
 
 
+[[_listsubscriptions]]
+=== GET /consumers/{groupid}/instances/{name}/subscription
+
+==== Description
+Retrieves a list of the topics to which is the consumer subscribed.
+
+
+==== Parameters
+
+[options="header", cols=".^2,.^3,.^9,.^4"]
+|===
+|Type|Name|Description|Schema
+|**Path**|**groupid** +
+__required__|ID of the consumer group to which the subscribed consumer belongs.|string
+|**Path**|**name** +
+__required__|Name of the consumer that you want to unsubscribe from topics.|string
+|===
+
+
+==== Responses
+
+[options="header", cols=".^2,.^14,.^4"]
+|===
+|HTTP Code|Description|Schema
+|**200**|List of subscribed topics and partitions.|<<_subscribedtopics,SubscribedTopics>>
+|**404**|The specified consumer instance was not found.|<<_error,Error>>
+|===
+
+
+==== Consumes
+
+* `application/vnd.kafka.v2+json`
+
+
+==== Produces
+
+* `application/vnd.kafka.v2+json`
+
+
+==== Tags
+
+* Consumers
+
+
+==== Example HTTP response
+
+===== Response 200
+[source,json]
+----
+{
+  "topics" : [ "my-topic1", "my-topic2" ],
+  "partitions" : [ {
+    "my-topic1" : [ "1", "2", "3" ]
+  }, {
+    "my-topic2" : [ "1" ]
+  } ]
+}
+----
+
+
+===== Response 404
+[source,json]
+----
+{
+  "application/vnd.kafka.v2+json" : {
+    "error_code" : 404,
+    "message" : "The specified consumer instance was not found."
+  }
+}
+----
+
+
 [[_unsubscribe]]
 === DELETE /consumers/{groupid}/instances/{name}/subscription
 
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -794,14 +794,9 @@ __required__|Name of the consumer that you want to unsubscribe from topics.|stri
 [options="header", cols=".^2,.^14,.^4"]
 |===
 |HTTP Code|Description|Schema
-|**200**|List of subscribed topics and partitions.|<<_subscribedtopics,SubscribedTopics>>
+|**200**|List of subscribed topics and partitions.|<<_subscribedtopiclist,SubscribedTopicList>>
 |**404**|The specified consumer instance was not found.|<<_error,Error>>
 |===
-
-
-==== Consumes
-
-* `application/vnd.kafka.v2+json`
 
 
 ==== Produces
@@ -822,9 +817,9 @@ __required__|Name of the consumer that you want to unsubscribe from topics.|stri
 {
   "topics" : [ "my-topic1", "my-topic2" ],
   "partitions" : [ {
-    "my-topic1" : [ "1", "2", "3" ]
+    "my-topic1" : [ 1, 2, 3 ]
   }, {
-    "my-topic2" : [ "1" ]
+    "my-topic2" : [ 1 ]
   } ]
 }
 ----
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -81,8 +81,6 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
     private Handler<AsyncResult<Void>> subscribeHandler;
     // handler called after an unsubscription request
     private Handler<AsyncResult<Void>> unsubscribeHandler;
-    // handler called after an get subscriptions request
-    private Handler<AsyncResult<Set<TopicPartition>>> listSubscriptionsHandler;
     // handler called after a request for a specific partition
     private Handler<AsyncResult<Optional<PartitionInfo>>> partitionHandler;
     // handler called after a topic partition assign request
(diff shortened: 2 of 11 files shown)


scholzj:
Have you tried it as I suggested with antoher consumer using the same group connected directly to Kafka?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -81,6 +81,8 @@
     private Handler<AsyncResult<Void>> subscribeHandler;
     // handler called after an unsubscription request
     private Handler<AsyncResult<Void>> unsubscribeHandler;
+    // handler called after an get subscriptions request
+    private Handler<AsyncResult<Set<TopicPartition>>> listSubscriptionsHandler;
```
ppatierno: In this case I don't see the need for having a field for the handler. The listSubscriptions method can get the handler as parameter as it happens for other methods. Wdyt?

src/main/resources/openapiv2.json
```diff
@@ -219,6 +219,39 @@
       ]
     },
     "/consumers/{groupid}/instances/{name}/subscription": {
+      "get": {
+        "consumes": [
```
ppatierno: It's a GET, it doesn't consume any data. The client doesn't send any JSON in the request.

src/main/resources/openapi.json
```diff
@@ -1462,6 +1499,53 @@
...
+                    },
+                    "partitions": {
+                        "type": "array",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "my-topic1": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "string"
```
ppatierno: Why the partition number is a string. We are using an int32 in all definitions.

src/main/resources/openapi.json
```diff
@@ -1462,6 +1499,53 @@
...
+                        "items": {
+                            "type": "string"
+                        }
+                    },
+                    "partitions": {
+                        "type": "array",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "my-topic1": {
```
ppatierno: Why a specific topic name "my-topic1" is in an object definition?

src/main/resources/openapi.json
```diff
@@ -1500,29 +1500,37 @@
...
             },
-            "SubscribedTopics": {
-                "title": "Root Type for SubscribedTopics",
-                "description": "",
+            "SubscribedTopic": {
+                "title": "SubscribedTopic",
                 "type": "object",
                 "properties": {
-                    "topics": {
+                    "my-topic1": {
```
ppatierno: I still don't understand how a topic name can be part of a definition.

src/main/resources/openapi.json
```diff
@@ -1462,6 +1499,61 @@
...
+                        3
+                    ]
+                }
+            },
+            "SubscribedTopicList": {
+                "title": "SubscribedTopicList",
+                "type": "object",
+                "properties": {
+                    "topics": {
+                        "$ref": "#/components/schemas/Topics"
```
ppatierno: Where is this Topics declaration. Anyway I don't think we need it, it's just an array of strings. How we do for subscription request for example?

stanlyDoge: We already had the Topics data type there (for subscription of topic list).

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -156,4 +157,62 @@ void subscriptionConsumerDoesNotExist(VertxTestContext context) throws Interrupt
...
+        topicsRoot.put("topics", topics);
+
+        consumerService()
+                .createConsumer(context, groupId, json)
+                .subscribeConsumer(context, groupId, name, topic)
+                .subscribeConsumer(context, groupId, name, topic2);
+
+        CompletableFuture<Boolean> consume = new CompletableFuture<>();
+
+        // hack to subscribe immediately
```
ppatierno: Don't call it an hack, because it's not :)
Just say that the poll is needed for starting rebalancing and partitions assignment as for Kafka protocol

src/main/resources/openapi.json
```diff
@@ -1462,6 +1499,61 @@
                         "topic2"
                     ]
                 }
+            },
+            "SubscribedTopic": {
```
ppatierno: It's more about assigned topic partitions than subscribed topics. We have the list of topics and then for each topic the assigned partitions. What do you think about AssignedTopicPartitions?

documentation/book/api/paths.adoc
```diff
@@ -770,6 +770,73 @@ __required__|List of topics to which the consumer will subscribe.|<<_topics,Topi
 ----
 
 
+[[_listsubscriptions]]
+=== GET /consumers/{groupid}/instances/{name}/subscription
+
+==== Description
+Retrieves a list of the topics to which is the consumer subscribed.
```
laidan6000:
```suggestion
Retrieves a list of topics that a consumer is subscribed to. The request for this operation must use the base URL (including the host and port) returned in the response from the `POST` request to `/consumers/{groupid}` that was used to create this consumer.
```

src/main/resources/openapi.json
```diff
@@ -235,6 +235,43 @@
...
+                                            "message": "The specified consumer instance was not found."
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                },
+                "operationId": "listSubscriptions",
+                "description": "Retrieves a list of the topics to which is the consumer subscribed."
```
laidan6000:
```suggestion
                "description": "Retrieves a list of the topics to which the consumer is subscribed."
```

src/main/resources/openapiv2.json
```diff
@@ -219,6 +219,36 @@
...
+            "examples": {
+              "application/vnd.kafka.v2+json": {
+                "error_code": 404,
+                "message": "The specified consumer instance was not found."
+              }
+            }
+          }
+        },
+        "operationId": "listSubscriptions",
+        "description": "Retrieves a list of the topics to which is the consumer subscribed."
```
laidan6000:
```suggestion
        "description": "Retrieves a list of the topics to which the consumer is subscribed."
```

documentation/book/api/paths.adoc
stanlyDoge: This file is generated. It takes the text from openapi file.

src/main/resources/openapi.json
```diff
@@ -1462,6 +1499,61 @@
                         "topic2"
                     ]
                 }
+            },
+            "AssignedTopicPartitions": {
+                "title": "AssignedTopicPartitions",
+                "type": "object",
+                "properties": {
+                    "topic": {
```
ppatierno: I still don't understand one thing. From this declaration it seems that this object has to have a field named "topic" which is not the case, or?

documentation/book/api/paths.adoc
laidan6000: Please can you update it in the definition file?

src/main/resources/openapi.json
stanlyDoge:
```"partitions": [
        {
            "test": [
                3,
                4
            ]
        }
    ]```
From this should be clear the topic `test` has partitions 3 and 4.

Comment (ppatierno) [reply src/main/resources/openapi.json]: I was really not convinced about that declaration because it clearly says that the JSON can have a field "topic" which is an array of int and it's not mandatory. I had a chat with a colleague with more OpenAPI knowledge and the outcome is the following. 
In our case, the example with "topic1" is working just because we are leveraging the `additionalProperties` field which is by default always true. It means that the OpenAPI validation just considers "topic1" as an additional property and not as the "topic" field mentioned above. The right specification is to explicit the `additionalProperties` defining them as arrays of int so something like this:

```yaml
title: AssignedTopicPartitions
type: object
additionalProperties:
        type: array
        items:
            format: int32
            type: integer
```

This is the YAML, you should change it to JSON.
Can you try this one but even having a more complex example with more arrays, so for example ...

```json
{
    "topic1": [
        1,
        2,
        3
    ],
    "topic2": [
        1,
        2
    ]
}
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -345,6 +345,44 @@ private void doSubscribe(RoutingContext routingContext, JsonObject bodyAsJson) {
...
+                JsonObject root = new JsonObject();
+                JsonArray topicsArray = new JsonArray();
+                JsonArray partitionsArray = new JsonArray();
+
+                HashMap<String, JsonArray> partitions = new HashMap<>();
+                for (TopicPartition topicPartition: listSubscriptions.result()) {
+                    if (!topicsArray.contains(topicPartition.getTopic())) {
+                        topicsArray.add(topicPartition.getTopic());
+                    }
+                    if (partitions.get(topicPartition.getTopic()) == null) {
```
ppatierno: why not using `contains` here as well?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -345,6 +345,44 @@ private void doSubscribe(RoutingContext routingContext, JsonObject bodyAsJson) {
         }
     }
 
+    public void doListSubscriptions(RoutingContext routingContext) {
+        this.listSubscriptions(listSubscriptions -> {
```
ppatierno: naming `listSubscriptions` it seems it's the actual list of subscriptions while it's the asyncresult, maybe we can use the same naming as other places like `done`, or `listSubscriptionsResult` ... whatever you prefer :-)

PR #304, opened 2019-08-26
Add consumer record data type

stanlyDoge:


src/main/resources/openapiv2.json
```diff
@@ -896,45 +896,56 @@
...
       "example": [
         {
-          "topic": "topic",
           "key": "key1",
-          "value": "value1",
+          "offset": 2,
           "partition": 0,
-          "offset": 2
+          "topic": "topic",
+          "value": "value1"
```
ppatierno: why same example with order changed?

src/main/resources/openapiv2.json
```diff
@@ -896,45 +896,56 @@
...
         },
         {
-          "topic": "topic",
           "key": "key2",
-          "value": "value2",
+          "offset": 3,
           "partition": 1,
-          "offset": 3
+          "topic": "topic",
+          "value": "value2"
```
ppatierno: ditto as above

src/main/resources/openapi.json
```diff
@@ -1066,30 +1066,41 @@
...
+                    "partition": 0,
+                    "topic": "topic",
+                    "value": "value1"
+                },
+                "properties": {
+                    "key": {
+                        "type": "string"
+                    },
+                    "offset": {
+                        "format": "int32",
```
ppatierno: can you check that format is int64 for other definitions having offset please? In that case we should have int64 here as well.

documentation/book/api/definitions.adoc
```diff
@@ -25,20 +25,16 @@ __optional__|The unique name for the consumer instance. The name is unique withi
...
+=== ConsumerRecord
 
 [options="header", cols=".^3,.^4"]
 |===
 |Name|Schema
 |**key** +
 __optional__|string
 |**offset** +
-__optional__|integer (int64)
+__optional__|integer (int32)
```
ppatierno: why int32? it should be int64

PR #311, opened 2019-08-30
Multi records partition bug

### LINKED ISSUE #310: Records are not saved if at least one of them has validation error
Kafka Bridge API returns error messages for all records which you want to save if at list one of them has an issue with validation. For example, if try to save 3 records and one of them has wrong `partition`, Bridge API will return an error for all records, even for those which are valid.

**Step to reproduce:**
1. Deploy Kafka Bridge
2. Create a topic with the name "test"
3. Send a POST request to the endpoint `/topics/{topicname}` with the following JSON:
```
{
  "records": [
    {
      "key": "key1",
      "value": "value1"
    },
    {
      "value": "value2",
      "partition": 1 <------- wrong value (should be 0)
    },
    {
      "value": "value3"
    }
  ]
}
```
Actual result:
Response has 3 errors for all messages ```
{
    "offsets": [
        {
            "error_code": 404,
            "message": "Invalid partition given with record: 1 is not in the range [0...1)."
        },
        {
            "error_code": 404,
            "message": "Invalid partition given with record: 1 is not in the range [0...1)."
        },
        {
            "error_code": 404,
            "message": "Invalid partition given with record: 1 is not in the range [0...1)."
        }
    ]
}
```
but actually we have only one wrong record with ```"partition": 1```
Expected result:
Record 1, 3 should be saved and only record 2 should have error.

### PR DESCRIPTION
**Pull request by stanlyDoge**
Fixes #310 and also refers to #312

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -96,13 +96,13 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
 
             for (int i = 0; i < sendHandlers.size(); i++) {
                 // check if, for each future, the sending operation is completed successfully or failed
-                if (((CompositeFutureImpl) done).resultAt(i) != null) {
-                    RecordMetadata metadata = ((CompositeFutureImpl) done).resultAt(i);
+                if (sendHandlers.get(i).succeeded() && sendHandlers.get(i).result() != null) {
+                    RecordMetadata metadata = (RecordMetadata) sendHandlers.get(i).result();
                     log.debug("Delivered record {} to Kafka on topic {} at partition {} [{}]", finalRecords.get(i), metadata.getTopic(), metadata.getPartition(), metadata.getOffset());
                     results.add(new HttpBridgeResult<>(metadata));
                 } else {
-                    String msg = done.cause().getMessage();
-                    int code = getErrorCode(done.cause());
+                    String msg = sendHandlers.get(i).cause().getMessage();
+                    int code = getErrorCode(sendHandlers.get(i).cause());
                     log.error("Failed to deliver record {}", finalRecords.get(i), done.cause());
                     results.add(new HttpBridgeResult<>(new HttpBridgeError(code, msg)));
                 }
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -21,7 +21,6 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.impl.CompositeFutureImpl;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.RoutingContext;
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -21,6 +21,7 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.impl.CompositeFutureImpl;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.RoutingContext;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by stanlyDoge** [comment]
Well, with this request on the 5 partition topic
```
{
  "records": [
    {
      "key": "key3",
      "value": "msg1"
    },
    {
      "value": "msg2",
      "partition" : 121
    }
  ]
}
```
, currently the bridge returns 200 OK with this body.
```
{
    "offsets": [
        {
            "error_code": 404,
            "message": "Invalid partition given with record: 121 is not in the range [0...5)."
        },
        {
            "error_code": 404,
            "message": "Invalid partition given with record: 121 is not in the range [0...5)."
        }
    ]
}
```

But it should be (which is after this PR).
```
{
    "offsets": [
        {
            "partition": 0,
            "offset": 28
        },
        {
            "error_code": 404,
            "message": "Invalid partition given with record: 121 is not in the range [0...5)."
        }
    ]
}
```

File: src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java [hunk 1]
```diff
@@ -95,8 +96,8 @@ public void handle(Endpoint<?> endpoint) {
 
             for (int i = 0; i < sendHandlers.size(); i++) {
                 // check if, for each future, the sending operation is completed successfully or failed
-                if (done.result() != null && done.result().succeeded(i)) {
-                    RecordMetadata metadata = done.result().resultAt(i);
+                if (((CompositeFutureImpl) done).resultAt(i) != null) {
+                    RecordMetadata metadata = ((CompositeFutureImpl) done).resultAt(i);
```
ppatierno: I don't think we need to do this hack casting to the actual `CompositeFutureImpl` class.
In this specific use case where an overall operation can have mixed failures and success, we have to use the `sendHandlers` list directly so the following code will work ...

```java
for (int i = 0; i < sendHandlers.size(); i++) {
                // check if, for each future, the sending operation is completed successfully or failed
                if (sendHandlers.get(i).succeeded() && sendHandlers.get(i).result() != null) {
                    RecordMetadata metadata = (RecordMetadata) sendHandlers.get(i).result();
                    log.debug("Delivered record {} to Kafka on topic {} at partition {} [{}]", finalRecords.get(i), metadata.getTopic(), metadata.getPartition(), metadata.getOffset());
                    results.add(new HttpBridgeResult<>(metadata));
                } else {
                    String msg = sendHandlers.get(i).cause().getMessage();
                    int code = getCodeFromMsg(msg);
                    log.error("Failed to deliver record " + finalRecords.get(i) + " due to {}", sendHandlers.get(i).cause());
                    results.add(new HttpBridgeResult<>(new HttpBridgeError(code, msg)));
                }
            }
```

ppatierno:
@scholzj I can see your point. This is a specific use case where a response can have success and failure in the body so we could think that both returning a 2xx or 4xx is wrong because in both cases you have a body with a partial failure or a partial success. 
I think that checking potential problems (i.e. if the partitions exist) for all the messages before sending them add more complexity (are we able to check every kind of problem that could happen?); relying on the underlying Kafka native error is simpler.
I would say that at this point returning 200 OK is fine IMHO.

scholzj:
I can understand the problems with the complexity of checking them before sending them. But does it cause for example issues with things like ordering if you skip message in the middle of the stream and continue with other messages? 

Also, I think if not everything is ok, we should prefer to return the error code and not 200. 200 will lead to people to assume all is well and not even decoding the body.

ppatierno:
> But does it cause for example issues with things like ordering if you skip message in the middle of the stream and continue with other messages?

currently, that's true. If you are sending 3 messages to partition 1 of a topic but for for the second one you put the wrong partition value i.e. 10 instead of 1 and it doesn't exist, then you have message 1 and 3 delivered but the message 2 fails which means you have to resend it ... and it will be delivered after message 3 (where the ordering is related to your business logic). But isn't it anyway a common Kafka producer problem? If you use the Kafka producer API `send` method quickly which means that the client is batching internally, the messages will be sent in one request and not all of them could be successfully delivered messing up the order.

> Also, I think if not everything is ok, we should prefer to return the error code and not 200. 200 will lead to people to assume all is well and not even decoding the body.

tbh I was thinking about that option but I wasn't sure because If we return a 4xx when everything fails, the user could assume that getting a 2xx means that everything is fine (which we know could not be the case) so being consistent with the response code sounds better to me.

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -592,4 +592,49 @@ void sendMessageWithUnknownProperty(VertxTestContext context) throws Throwable {
...
+
+                        JsonArray offsets = bridgeResponse.getJsonArray("offsets");
+                        assertEquals(2, offsets.size());
+                        JsonObject metadata = offsets.getJsonObject(0);
+                        assertNotNull(metadata.getInteger("partition"));
+                        assertEquals(partition, metadata.getInteger("partition"));
+                        assertEquals(0L, metadata.getLong("offset"));
+                        JsonObject metadata2 = offsets.getJsonObject(1);
+                        assertEquals(HttpResponseStatus.NOT_FOUND.code(), metadata2.getInteger("error_code"));
+                        assertEquals("Invalid partition given with record: 500 is not in the range [0...3).", metadata2.getString("message"));
```
ppatierno: Can you do something like this as we already do for other tests where an error is expected please?

```java
HttpBridgeError error = HttpBridgeError.fromJson(offsets.getJsonObject(1));
assertEquals(HttpResponseStatus.NOT_FOUND.code(), error.getCode());
assertEquals("Invalid partition given with record: 500 is not in the range [0...3).", error.getMessage());
```

ppatierno:
@stanlyDoge there are conflicts to fix.

PR #314, opened 2019-09-03
Upgrade to Vertx 3.8.1

tombentley:


ppatierno:
With this update, the `sendToNonExistingPartitionsTest` is now failing due to a different way they are dealing with a "wrong" requested partition (greater than the actual partitions count).
The change is related to this JIRA [KAFKA-6833](https://issues.apache.org/jira/browse/KAFKA-6833) and the code diff is [here](https://github.com/apache/kafka/pull/6154/files).

So the test is failing because the send method doesn't return immediately (after waiting for metadata) but just loops and the Vert.x Junit test timeout is less than the Kafka `max.block.ms` (60000 ms). Anyway, the returned error message is misleading "Topic sendToNonExistingPartitionsTest not present in metadata after 60000 ms." while it should be more as "Partition 1000 of topic sendToNonExistingPartitionsTest with partition count 3 is not present in metadata after 60000 ms.".

In conclusion, it could be a fix as suggested by @tombentley about the right error message but meanwhile, we have to change our test to deal with this new semantic.
I am going to close this PR and opening a new one with Vert.x Kafka client update plus the changed test.
Are you ok @tombentley ?

tombentley:
:+1:

PR #315, opened 2019-09-03
Scenarios for parameterized ST to create consumer

serrss:
Added negative scenarios for parameterized ST to create consumer

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -212,9 +212,9 @@ public class ConsumerTest extends HttpBridgeTestBase {
     }
 
     @Test
-    void createConsumerWithWrongKeyConsumerRequestTimeoutMs(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
-        checkCreatingConsumer("request.timeout.ms", "30000", HttpResponseStatus.BAD_REQUEST,
-                "Validation error on: body - $.request.timeout.ms: is not defined in the schema and the schema does not allow additional properties", context);
+    void createConsumerWithNotExistingParameter(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
+        checkCreatingConsumer("foo", "bar", HttpResponseStatus.BAD_REQUEST,
+                "Validation error on: body - $.foo: is not defined in the schema and the schema does not allow additional properties", context);
 
         context.completeNow();
         assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -15,6 +15,8 @@ import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.ext.web.codec.BodyCodec;
 import io.vertx.junit5.VertxTestContext;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvFileSource;
 
 import javax.xml.bind.DatatypeConverter;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


ppatierno:
I would love to have a feedback from @tombentley on this because I had with him more discussions around parameterized tests which are a pain to debug (you have to hack part of the code for debugging one specific test case which is failing for example). How does it work in this case?

tombentley:
In this case the parameters are coming from a CSV file, so it should be easy to only set a single set of parameters. by just removing other lines. This is only a parameterized test _method_, not a parameterized test _class_, as we have elsewhere, which is admittedly awkward to debug.

However, unless we anticipate more rows being added to the CSV file, I think it would probably be simpler/more obvious to just have a `private` parameterized method and have four `@Test` methods calling the private common method.

serrss:
@tombentley well, thanks for the answer. I guess 4 scenarios is everything that we can cover at the moment. Maybe if ENG team will add new HTTP codes and error messages or new fields in JSON for this endpoint, the number of the scenario will increase.
@Frawless it is also very interesting to hear your opinion on parameterized tests in this case.

Frawless:
I agree, that debug of parametrized tests is hard. Instead of CSV I prefer JSON as a parameters source - https://github.com/strimzi/strimzi-kafka-operator/pull/1907  I used it there.

From my current POV (and from operator tests POV), it's better to not use parametrized tests as I suggested. I am facing issue with upgrade tests (operator, 10m+ for each test), which some of them are working only on specific env (same for nodePort tests for example). It's hard to skip them, because we cannot annotate each test case. 

If I take a look here, the test is pretty simple and fast (few seconds instead of few minutes as we have in operator) so I think usage of parametrized test here is fine. But for more complex tests (runtime > 1m) we should use approach which @tombentley suggests.

serrss:
@ppatierno @tombentley @Frawless Thanks folks for your recommendations, changes have been applied.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -186,26 +184,39 @@ void createConsumerWithForwardedHeaderWrongProto(VertxTestContext context) throw
...
+        checkCreatingConsumer("fetch.min.bytes", "foo", HttpResponseStatus.INTERNAL_SERVER_ERROR,
+                "Invalid value foo for configuration fetch.min.bytes: Not a number of type INT", context);
+
+        context.completeNow();
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+    }
+
+    @Test
+    void createConsumerWithWrongRequestTimeoutMs(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
+        checkCreatingConsumer("request.timeout.ms", "30000", HttpResponseStatus.BAD_REQUEST,
```
ppatierno: well this test should pass, because 30000 ms is a valid value for a timeout in ms configuration parameter.
As you look at the error message it's about OpenAPI validation because the right parameter name is `consumer.request.timeout.ms` and not just `request.timeout.ms`. After fixing it, you should use a value which is not a valid timeout (i.e. just "foo") for doing the test.

serrss: I'm happy that you highlighted this scenario! I don't want to check `consumer.request.timeout.ms` in this scenario as you mentioned before. I want to check validation error for wrong JSON `KEY` (not value). This is a reason why I have expected error message as `Validation error on: body - $.request.timeout.ms...`

ppatierno: so the method name is misleading :-)
It should not be `createConsumerWithWrongRequestTimeoutMs` but more `createConsumerWithWrongConfigParameter` or `createConsumerWithNotExistingParameter`. The name it has now seems to relate to the value, so wrong value for the request timeout parameter.

serrss: okay, if you don't mind I will call it like `createConsumerWithWrongKeyConsumerRequestTimeoutMs`. Let's say that in JSON data has two types, key and value and having `Key` at the method name should be more clear that we check validation for syntax `consumer.request.timeout.ms`

serrss: Pushed changes

ppatierno: Tbh checking that the name for `request.timeout.ms` is wrong is like checking that you are using a "wrong parameter name". Using the request timeout ms as a reference is misleading.

serrss: Well, if we have an error message for this case? why do you think this test can be misleading? If we have some logic at the back-end we have to check it. thanks

ppatierno: Testing that you are using a wrong name for request timeout is like testing that you are providing a wrong parameter name. It's a superset of the specific case you are testing. For this reason it just makes sense testing the usage of a wrong parameter name; you tell will just be a specific case of that.

serrss: So, what do you recommend to do?

ppatierno: changing the test method in `createConsumerWithNotExistingParameter` and using `foo` or something else as parameter name

serrss: thanks, done

PR #317, opened 2019-09-03
Upgrade to Vertx 3.8.1

ppatierno:
Other than upgrading to Vert.x 3.8.1 This PR fixes the problem described in a previous closed PR #314

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -138,7 +138,7 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
     }
 
     private int getCodeFromMsg(String msg) {
-        if (msg.contains("Invalid partition")) {
+        if (msg.contains("not present in metadata")) {
             return HttpResponseStatus.NOT_FOUND.code();
         } else {
             return HttpResponseStatus.INTERNAL_SERVER_ERROR.code();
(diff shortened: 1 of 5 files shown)


ppatierno:
My concern is that the Kafka producer is covering more different things with one timeout exception, at least when topic and partition don't exist. Do we know any other cases where the same exception is raised? It means that checking the error message should be done at bridge level searching for a "not present in metadata" in order to return a 404.

tombentley:
Yeah, I know it requires checking the exception message. But you have the test to catch regressions should the message change in the future. And not breaking your REST API is important. Otherwise you risk propagating this bug onto your clients, just in a more HTTP flavour.

ppatierno:
@tombentley I agreed with you and updated the bridge side to check the message. I made a small change in a code path which wasn't used anymore.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -138,7 +138,7 @@ private JsonObject buildOffsets(List<HttpBridgeResult<?>> results) {
     }
 
     private int getCodeFromMsg(String msg) {
```
tombentley: There's nothing stopping this method from taking `Exception` as parameter. That way you can say `e instanceof TimeoutException && e.getMessage() != null && e.getMessage().contains("not present in metadata")`, which is less likely to catch a false positive.

ppatierno: good point! done ;)

PR #319, opened 2019-09-04
ST: Case when test want to seek first offset in non existent topic

serrss:
One more scenario when Kafka Topic has the wrong name. It's necessary to check 404 and error message.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
@@ -69,10 +69,20 @@ public class SeekTest extends HttpBridgeTestBase {
                 });
         instanceNotFound.get(TEST_TIMEOUT, TimeUnit.SECONDS);
 
-        // Specified consumer instance did not have one of the specified partitions assigned.
-        CompletableFuture<Boolean> consumerInstanceDontHavePartition = new CompletableFuture<>();
 
-        int nonExistenPartition = Integer.MAX_VALUE;
+    }
+
+    @Test
+    void seekToNotExistentPartitionInSubscribedTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        String topic = "seekToNotExistentPartitionInSubscribedTopic";
+        kafkaCluster.createTopic(topic, 1, 1);
+
+        // create consumer
+        consumerService()
+            .createConsumer(context, groupId, jsonConsumer)
+            .subscribeTopic(context, groupId, name, new JsonObject().put("topic", topic).put("partition", 0));
+
+        int nonExistenPartition = 2;
         JsonArray nonExistentPartitionJSON = new JsonArray();
         nonExistentPartitionJSON.add(new JsonObject().put("topic", topic).put("partition", nonExistenPartition));
 
@@ -88,10 +98,17 @@ public class SeekTest extends HttpBridgeTestBase {
                     assertEquals(HttpResponseStatus.NOT_FOUND.code(), response.statusCode());
                     assertEquals(HttpResponseStatus.NOT_FOUND.code(), error.getCode());
                     assertEquals("No current assignment for partition " + topic + "-" + nonExistenPartition, error.getMessage());
+                    context.completeNow();
                 });
-                consumerInstanceDontHavePartition.complete(true);
             });
-        consumerInstanceDontHavePartition.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+    }
+
+    @Test
+    void seekToNotExistentTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        // create consumer
+        consumerService()
+                .createConsumer(context, groupId, jsonConsumer);
 
         // Specified consumer instance did not have one of the specified topics.
         CompletableFuture<Boolean> consumerInstanceDontHaveTopic = new CompletableFuture<>();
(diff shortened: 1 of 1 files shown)


ppatierno:
Taking a look at the `seekToBeginningConsumerOrPartitionNotFound` method I would change it "completely" :-)

This method doesn't subscribe to the topic when trying to seek to a not existing partition and trying to seek to a not subscribed topic should have its own test.
I think that it would be better to break down this method in different methods with the following tests:

- trying to seek to a not existing consumer (this test doesn't need to create a consumer and then seeking to a not existing one).
- trying to seek to a not existing partition in a subscribed topic. It means that here you need to create a consumer, subscribe to an existing topic and then seeking to a not existing partition.
- trying to see to a not existing topic. If you try to subscribe in this case you get the error earlier, so the created consumer should not subscribe but just trying to seek ... maybe could make sense to test.

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -64,15 +50,24 @@ void seekToBeginningConsumerOrPartitionNotFound(VertxTestContext context) throws
...
                     });
-                    instanceNotFound.complete(true);
                 });
-        instanceNotFound.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+    }
+
+    @Test
+    @Disabled // This test was disabled because of known issue described in https://github.com/strimzi/strimzi-kafka-bridge/issues/320
+    void seekToNotExistentPartitionInSubscribedTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno:
```suggestion
    void seekToNotExistingPartitionInSubscribedTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -90,9 +85,40 @@ void seekToBeginningConsumerOrPartitionNotFound(VertxTestContext context) throws
...
                     context.completeNow();
                 });
-                consumerInstanceDontHavePartition.complete(true);
             });
-        consumerInstanceDontHavePartition.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+    }
+
+    @Test
+    void seekToNotExistentTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno:
```suggestion
    void seekToNotExistingTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -64,15 +50,24 @@ void seekToBeginningConsumerOrPartitionNotFound(VertxTestContext context) throws
...
-                    instanceNotFound.complete(true);
                 });
-        instanceNotFound.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+    }
+
+    @Test
+    @Disabled // This test was disabled because of known issue described in https://github.com/strimzi/strimzi-kafka-bridge/issues/320
+    void seekToNotExistentPartitionInSubscribedTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        String topic = "seekToNotExistentPartitionInSubscribedTopic";
```
ppatierno:
```suggestion
        String topic = "seekToNotExistingPartitionInSubscribedTopic";
```

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -64,15 +50,24 @@ void seekToBeginningConsumerOrPartitionNotFound(VertxTestContext context) throws
...
 
-        // Specified consumer instance did not have one of the specified partitions assigned.
-        CompletableFuture<Boolean> consumerInstanceDontHavePartition = new CompletableFuture<>();
+        // create consumer
+        consumerService()
+            .createConsumer(context, groupId, jsonConsumer)
+            .subscribeTopic(context, groupId, name, new JsonObject().put("topic", topic).put("partition", 0));
 
-        int nonExistenPartition = Integer.MAX_VALUE;
+        int nonExistenPartition = 2;
```
ppatierno:
```suggestion
        int notExistingPartition = 2;
```

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -90,9 +85,40 @@ void seekToBeginningConsumerOrPartitionNotFound(VertxTestContext context) throws
...
+    @Test
+    void seekToNotExistentTopic(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        // create consumer
+        consumerService()
+                .createConsumer(context, groupId, jsonConsumer);
+
+        // Specified consumer instance did not have one of the specified topics.
+        CompletableFuture<Boolean> consumerInstanceDontHaveTopic = new CompletableFuture<>();
+
+        String nonExistentTopic = "NotExistentTopic";
```
ppatierno:
```suggestion
        String notExistingTopic = "notExistingTopic";
```

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -90,9 +85,40 @@ void seekToBeginningConsumerOrPartitionNotFound(VertxTestContext context) throws
...
+        JsonArray nonExistentTopicJSON = new JsonArray();
```
ppatierno:
```suggestion
        JsonArray notExistingTopicJSON = new JsonArray();
```

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
serrss: Kafka Topic has only 1 partition... so it's more clear to me to have value 2 here

ppatierno: why did you think that I wanted to change the value? :-) I changed just the name. It was you changing ffrom MAX_VALUE to 2 :P

Issue #318, opened 2019-09-04
The Model for Consumer contains wrong types if compare with Kafka official doc


serrss:
If look at the official Kafka documentation https://kafka.apache.org/documentation/ some of the parameters from Consumer model which we use for endpoint `POST ​/consumers​/{groupid}` has the wrong type.
For example:
1. At the Kafka Bridge OpenAPI doc, parameter `consumer.request.timeout.ms` has type `String`, but in official Kafka doc, `request.timeout.ms` has type `int`
2. Parameter `fetch.min.bytes` has type `String` at Kafka Bridge OpenAPI doc, in Kafka official doc it has type `int`
3. Parameter `enable.auto.commit` has type `String` at Kafka Bridge OpenAPI doc, in Kafka official doc it has type `boolean`

I also tried to create a consumer with types which we have in official Kafka doc and had HTTP 400 Bad request with a message: `Validation error on: body.consumer.request.timeout.ms - $.consumer.request.timeout.ms: integer found, string expected`

So this is not only documentation issue, we also need to change engineering code to support the same types as at Kafka official doc.

PR #322, opened 2019-09-06
Fixed consumer parameters validation

### LINKED ISSUE #308: There is no validation for `consumer.request.timeout.ms` during consumer creation
It's possible to create a consumer with value `foo` for parameter `consumer.request.timeout.ms` via endpoint `POST /consumers/{groupid}`

An example of JSON:
```
{
  "name": "consumer1",
  "format": "binary",
  "auto.offset.reset": "earliest",
  "enable.auto.commit": "false",
  "fetch.min.bytes": "512",
  "consumer.request.timeout.ms": "foo" <------- need validation for this field
}
```

ppatierno:
This PR fixes #308 and #318

### DIFF
--- a/documentation/book/api/definitions.adoc
+++ b/documentation/book/api/definitions.adoc
@@ -18,11 +18,11 @@ __optional__|Resets the offset position for the consumer.
 If set to `earliest`, messages are read from the first offset.
 If set to `latest`, messages are read from the latest offset.|string
 |**consumer.request.timeout.ms** +
-__optional__|Sets the maximum amount of time, in milliseconds, for the consumer to wait for messages for a request. If the timeout period is reached without a response, an error is returned.|string
+__optional__|Sets the maximum amount of time, in milliseconds, for the consumer to wait for messages for a request. If the timeout period is reached without a response, an error is returned.|integer
 |**enable.auto.commit** +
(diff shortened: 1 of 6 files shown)


ppatierno:
In this PR I left the current `consumer.request.timeout.ms` as it is. 
Tbh, and I noticed that just now, it's a nonsense to me because the other parameters, `fetch.min.bytes` for example, don't have the `consumer.` prefix. I would get rid of it because it's clear that the parameter is related to consumer (the request is about consumer creation).
We are using prefixes in the properties configuration file for overall bridge configuration parameters where you can set the configuration for producers and consumers using `producer.` and `consumer.` prefix.
I know that removing right now means a breaking change but not doing that now in an early stage means taking this "nonsense" forward.
Wdyt @tombentley @scholzj ?

ppatierno:
Do you mean adding this to the CHANGELOG? If yes, thinking more we have other stuff to add there. Maybe it make sense to open a PR with all the changes.
BTW do you want to keep this consumer. prefix? :(

scholzj:
> Do you mean adding this to the CHANGELOG? If yes, thinking more we have other stuff to add there. Maybe it make sense to open a PR with all the changes.

I guess so - or some other place. But when we release this we need to tell the users that this changed and they need to update their applications.

> BTW do you want to keep this consumer. prefix? :(

Yes, I think slightly bad name does not warrant breaking the compatibility.

PR #336, opened 2019-09-13
Add client side OAUTH support (for Kafka clients only)

scholzj:
This PR adds support to the OAuth authentication for the Kafka client when connecting to the Kafka brokers. It:
* Adds the dependency to the client callback library
* Updates the Docker scripts to be able to configure OAuth

### DIFF
--- a/bin/docker/kafka_bridge_config_generator.sh
+++ b/bin/docker/kafka_bridge_config_generator.sh
@@ -38,12 +38,12 @@ if [ -n "$KAFKA_BRIDGE_SASL_MECHANISM" ]; then
         SECURITY_PROTOCOL="SASL_PLAINTEXT"
     fi
 
-    PASSWORD=$(cat /opt/strimzi/bridge-password/$KAFKA_BRIDGE_SASL_PASSWORD_FILE)
-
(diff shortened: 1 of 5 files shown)


bin/docker/kafka_bridge_config_generator.sh
```diff
@@ -31,26 +31,47 @@ EOF
...
+
+        if [ ! -z "$KAFKA_BRIDGE_OAUTH_CLIENT_SECRET" ]; then
+            OAUTH_CLIENT_SECRET="oauth.client.secret=\"$KAFKA_BRIDGE_OAUTH_CLIENT_SECRET\""
+        fi
+
+        if [ -f "/tmp/strimzi/oauth.truststore.p12" ]; then
+            OAUTH_TRUSTSTORE="oauth.ssl.truststore.location=\"/tmp/strimzi/oauth.truststore.p12\" oauth.ssl.truststore.password=\"${CERTS_STORE_PASSWORD}\" oauth.ssl.truststore.type=\"PKCS12\""
+        fi
+
+        SASL_MECHANISM="OAUTHBEARER"
```
ppatierno: can you move this line up having the same pattern as the other SASL mechanism blocks. Just setting the SASL_MECHANISM as the first thing.

PR #340, opened 2019-09-16
Add spotbugs

### LINKED ISSUE #270: Use spotbugs rather than unmaintained findbugs
The current pom.xml uses findbugs maven plugin to check the code, but findbugs is unmaintained. We should switch to the spotbugs plugin.

stanlyDoge:
This PR fixes #270

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -494,7 +494,7 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
                     new io.vertx.kafka.client.consumer.OffsetAndMetadata(offsetAndMetadata.offset() + 1, offsetAndMetadata.metadata()));
         });
 
-        if (offsets != null && !offsets.isEmpty()) {
+        if (!offsets.isEmpty()) {
             this.consumer.commit(kafkaOffsets, ar -> {
                 if (ar.succeeded()) {
                     this.offsetTracker.commit(offsets);
--- a/src/main/java/io/strimzi/kafka/bridge/converter/DefaultSerializer.java
+++ b/src/main/java/io/strimzi/kafka/bridge/converter/DefaultSerializer.java
@@ -5,6 +5,7 @@
 
 package io.strimzi.kafka.bridge.converter;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.kafka.common.errors.SerializationException;
 import org.apache.kafka.common.serialization.Serializer;
 
@@ -20,11 +21,12 @@ public class DefaultSerializer<T> implements Serializer<T> {
 
     }
 
+    @SuppressFBWarnings("PZLA_PREFER_ZERO_LENGTH_ARRAYS")
     @Override
     public byte[] serialize(String topic, T data) {
 
         if (data == null)
-            return new byte[0];
+            return null;
 
         ByteArrayOutputStream b = new ByteArrayOutputStream();
         ObjectOutputStream o = null;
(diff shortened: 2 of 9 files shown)


src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -494,7 +494,7 @@ private void commitOffsets(boolean clear) {
                     new io.vertx.kafka.client.consumer.OffsetAndMetadata(offsetAndMetadata.offset() + 1, offsetAndMetadata.metadata()));
         });
 
-        if (offsets != null && !offsets.isEmpty()) {
+        if (!offsets.isEmpty()) {
```
ppatierno: Is it impossible for offsets to be null?

src/main/java/io/strimzi/kafka/bridge/converter/DefaultSerializer.java
```diff
@@ -24,7 +24,7 @@ public void configure(Map<String, ?> configs, boolean isKey) {
     public byte[] serialize(String topic, T data) {
 
         if (data == null)
-            return null;
+            return new byte[0];
```
ppatierno: Why?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -408,7 +408,12 @@ private void errorHandler(RoutingContext routingContext) {
         String message = null;
         // in case of validation exception, building a meaningful error message
         if (routingContext.failure() != null && routingContext.failure() instanceof ValidationException) {
-            ValidationException validationException = (ValidationException) routingContext.failure();
+            ValidationException validationException = new ValidationException("null");
+            try {
+                validationException = (ValidationException) routingContext.failure();
+            } catch (ClassCastException e) {
+                e.printStackTrace();
+            }
```
ppatierno: Why?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
stanlyDoge: If you take a look at both implementations of `getOffsets()`, but are returning instancied objects. So `offsets` cannot be null.

src/main/java/io/strimzi/kafka/bridge/converter/DefaultSerializer.java
stanlyDoge: This is what spotbugs suggested. I can revert and suppress a warning in this method.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: what damn does it mean, we have an `if` before that

`if (routingContext.failure() != null && routingContext.failure() instanceof ValidationException)`

which confirm that it's a `ValidationException`

src/main/java/io/strimzi/kafka/bridge/converter/DefaultSerializer.java
ppatierno: what's the warning it's raising?

stanlyDoge:
```[ERROR] Should io.strimzi.kafka.bridge.converter.DefaultSerializer.serialize(String, Object) return a zero length array rather than null? [io.strimzi.kafka.bridge.converter.DefaultSerializer] At DefaultSerializer.java:[line 27] PZLA_PREFER_ZERO_LENGTH_ARRAYS
```

ppatierno: I would suppress this warning tbh

src/main/java/io/strimzi/kafka/bridge/tracker/FullOffsetTracker.java
```diff
@@ -108,23 +108,13 @@ public void delivered(int partition, long offset) {
...
                 state.settlements.clear();
             }
-
-        } else if (offset > state.firstUnsettled) {
-
-            // do nothing
-
-        } else {
-
-            // impossible ?
```
ppatierno: can you suppress the warning here as well? It's better having these comments for now.

PR #347, opened 2019-09-20
Logging bridge version during startup

### LINKED ISSUE #346: The current bridge version isn't printed in the log
It would be useful having the bridge version logged on startup.

ppatierno:
This PR fixes #346

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -35,7 +35,7 @@ public class Application {
     private static final int DEFAULT_HEALTH_SERVER_PORT = 8080;
 
     public static void main(String[] args) {
-        log.info("Apache Kafka Bridge {} is starting", Application.class.getPackage().getImplementationVersion());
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -35,10 +35,11 @@
     private static final int DEFAULT_HEALTH_SERVER_PORT = 8080;
 
     public static void main(String[] args) {
+        log.info("Apache Kafka Bridge {} is starting", Application.class.getPackage().getImplementationVersion());
```
scholzj: I'm quite sure we should not call this `Apache Kafka Bridge`. At least `Strimzi Kafka Bridge` would be needed.

ppatierno: but in the CO with don't have Strimzi Cluster Operator, we have just Cluster Operator.
So in this case we can just say Bridge or Kafka Bridge ... is Apache Kafka which is not ok?

PR #349, opened 2019-09-23
[MO] - [test] -> consumer with invalid conf

Seequick1:
This PR just add one simple test case to consumer suite.
- adding to coverage 422 status code with also message.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -1416,7 +1416,7 @@ public class ConsumerTest extends HttpBridgeTestBase {
     }
 
     @Test
-    void createConsumerWithInvalidValueOfConfiguration(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+    void createConsumerWithInvalidFormat(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
         CompletableFuture<Boolean> create = new CompletableFuture<>();
 
         JsonObject requestHeader = new JsonObject();
@@ -1425,9 +1425,6 @@ public class ConsumerTest extends HttpBridgeTestBase {
         LOGGER.info("Adding invalid value 'biary' to 'format' property configuration to invoke |422| status code");
 
         requestHeader.put("format", "biary");
-        requestHeader.put("auto.offset.reset", "earliest");
-        requestHeader.put("fetch.min.bytes", "512");
-        requestHeader.put("consumer.request.timeout.ms", "30000");
 
         consumerService()
                 .createConsumerRequest(groupId, requestHeader)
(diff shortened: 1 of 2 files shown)


src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1407,4 +1414,34 @@ private void checkCreatingConsumer(String key, String value, HttpResponseStatus
             });
         consumer.get(TEST_TIMEOUT, TimeUnit.SECONDS);
     }
+
+    @Test
+    void createConsumerWithInvalidValueOfConfiguration(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno: we have other methods checking for other wrong parameters, we should be more precise here so using something like `createConsumerWithInvalidFormat`, wdyt?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1407,4 +1414,34 @@ private void checkCreatingConsumer(String key, String value, HttpResponseStatus
...
+
+        JsonObject requestHeader = new JsonObject();
+        requestHeader.put("name", name);
+
+        LOGGER.info("Adding invalid value 'biary' to 'format' property configuration to invoke |422| status code");
+
+        requestHeader.put("format", "biary");
+        requestHeader.put("auto.offset.reset", "earliest");
+        requestHeader.put("fetch.min.bytes", "512");
+        requestHeader.put("consumer.request.timeout.ms", "30000");
```
ppatierno: I think that in this test we can avoid to set these optional parameters

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1407,4 +1414,34 @@ private void checkCreatingConsumer(String key, String value, HttpResponseStatus
...
+
+        consumerService()
+                .createConsumerRequest(groupId, requestHeader)
+                .as(BodyCodec.jsonObject())
+                .sendJsonObject(requestHeader, ar -> {
+                    context.verify(() -> {
+                        assertTrue(ar.succeeded());
+                        HttpResponse<JsonObject> response = ar.result();
```
ppatierno: in all the test we expect an error we get an `HttpBridgeError` from the body testing the related error code and message other than the HTTP status code. Can you do the same here?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
Seequick1: make sense :+1:

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -26,13 +28,18 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class ConsumerTest extends HttpBridgeTestBase {
 
+    private static final Logger LOGGER = LogManager.getLogger(ConsumerTest.class);
```
ppatierno: what about putting the logger in the base class?

Seequick1: Yeah, good idea. I will `create task in JIRA and issue` for that and after release i will take a look on that. I moved the `logger` to base class.

Seequick1: [Issue migrate](https://github.com/strimzi/strimzi-kafka-bridge/issues/352)

PR #353, opened 2019-09-26
Check for duplicated consumer creation

granzoto:
Checking for duplicated consumer creation, also checking for other return codes in the consumer creation tests

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -64,6 +64,21 @@ public class ConsumerTest extends HttpBridgeTestBase {
     }
 
     @Test
+    void createConsumerAlreadyExists(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
+        // create consumer
+        consumerService().createConsumer(context, groupId, consumerWithEarliestReset, HttpResponseStatus.OK);
+
+        // Create the same consumer
+        consumerService().createConsumer(context, groupId, consumerWithEarliestReset, HttpResponseStatus.CONFLICT);
+
+        context.completeNow();
+        assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+
+        consumerService()
+                .deleteConsumer(context, groupId, name);
+    }
+
+    @Test
     void createConsumerWrongFormat(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
         JsonObject consumerJson = new JsonObject()
             .put("name", name)
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -64,12 +64,31 @@ public class ConsumerTest extends HttpBridgeTestBase {
     }
 
     @Test
-    void createConsumerAlreadyExists(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
+    void createDuplicatedConsumer(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
         // create consumer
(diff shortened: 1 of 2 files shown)


src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -63,6 +63,21 @@ void createConsumer(VertxTestContext context) throws InterruptedException, Timeo
...
 
+    @Test
+    void createConsumerAlreadyExists(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
+        // create consumer
+        consumerService().createConsumer(context, groupId, consumerWithEarliestReset, HttpResponseStatus.OK);
+
+        // Create the same consumer
+        consumerService().createConsumer(context, groupId, consumerWithEarliestReset, HttpResponseStatus.CONFLICT);
+
+        context.completeNow();
```
Frawless: This has to be at the end of test. It set test as `PASS`, so if you perform any other action like `deleteConsumer()` it will not allow test result and I am sure, we don't want it. `assert` for context status after `context.completeNow()` is fine.

src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
```diff
@@ -140,13 +140,20 @@ public ConsumerService subscribeTopic(VertxTestContext context, String groupId,
     }
 
     public ConsumerService createConsumer(VertxTestContext context, String groupId, JsonObject json) throws InterruptedException, ExecutionException, TimeoutException {
+        return this.createConsumer(context, groupId, json, HttpResponseStatus.OK);
+    }
+
+    public ConsumerService createConsumer(VertxTestContext context, String groupId, JsonObject json, HttpResponseStatus statusCode) throws InterruptedException, ExecutionException, TimeoutException {
```
Frawless: In general, I expect that `createConsumer` will create consumer which I can use, I think you should name it as `createDuplicateConsumer` which will call `createConsumer` with `HttpResponseStatus` as a parameter (default `createaConsumer` will remain as they are). @ppatierno  WDYT?

ppatierno: I think that we need a little bit of refactoring here :-)
While I agree with @Frawless about the name I think that the implementation of the current tests is spreading checks across classes. So ... 
Creating a consumer can return a 200 OK, a 409 CONFLICT (which is the one we are adding here) and 422 UNPROCESSABLE ENTITY. While we are testing 200 OK here with the `createConsumer` and trying to handle 409 still here with a `createDuplicateConsumer`, the 422 error is checked in the `ConsumerTest` class (i.e. `createConsumerWrongFormat`, `createConsumerEnableAutoCommit`, ... methods for building wrong consumer configuration). In those cases the consumer service is used but not the `createConsumer` but the `createConsumerRequest`. I think that we should try be consistent. Maybe the `createConsumer` that we have here should just try to create a consumer but all the checks should be moved in the specific `ConsumerTest` class? At same time, if it's consistent for tests where we are testing consumer creation, it's verbose to have this check-in tests where you want to create a consumer as first step and then testing polling, sending, etc etc. Does it make sense? Any thoughts from you tests "masters" ? :)

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
granzoto: Ok, changed the tst name to createDuplicatedConsumer

src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
granzoto: I have added a test to cover 422 return code on consumer creation, so we are now testing all possible return codes during consumer creation, instead of only 200/non-200.
I think we should keep distinct tests for both createConsumer and createConsumerRequest, just in case in the future we add any new return code for any of them ( Consumer creation or Consumer request )

granzoto:
After a better code review, this PR is not necessary, the tests added here are already covered in other tests.

PR #355, opened 2019-10-01
Fixed consumer creation with empty body

### LINKED ISSUE #354: Creating consumer with empty HTTP POST request doesn't work
When executing an HTTP POST on the `/consumers/{groupid}` endpoint for creating a new consumer but with an empty body (no JSON at all), the bridge returns `400 Bad Request` with the following body:

```json
{
    "error_code": 400,
    "message": "Validation error on: body - $: string found, object expected"
}
```

It should just create a new consumer with a random name and default configuration because the body should not be mandatory.
Currently, it works if the body is an empty JSON like `{ }` which is a workaround but it's not ok.

ppatierno:
This PR fixes #354

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -23,7 +23,6 @@ import io.vertx.core.http.HttpConnection;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerOptions;
 import io.vertx.core.http.HttpServerRequest;
-import io.vertx.core.json.DecodeException;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
@@ -209,17 +208,12 @@ public class HttpBridge extends AbstractVerticle implements HealthCheckable {
     private void createConsumer(RoutingContext routingContext) {
         this.httpBridgeContext.setOpenApiOperation(HttpOpenApiOperations.CREATE_CONSUMER);
 
-        JsonObject bodyAsJson = null;
-        // TODO: it seems that getBodyAsJson raises an exception when the body is empty and not null
-        try {
-            bodyAsJson = routingContext.getBodyAsJson();
-        } catch (DecodeException ex) {
-            
-        }
+        // check for an empty body
+        JsonObject body = routingContext.getBody().length() != 0 ? routingContext.getBodyAsJson() : new JsonObject();
         SinkBridgeEndpoint<byte[], byte[]> sink = null;
 
         try {
-            EmbeddedFormat format = EmbeddedFormat.from(bodyAsJson != null ? bodyAsJson.getString("format", "binary") : "binary");
+            EmbeddedFormat format = EmbeddedFormat.from(body.getString("format", "binary"));
 
             sink = new HttpSinkBridgeEndpoint<>(this.vertx, this.bridgeConfig, this.httpBridgeContext,
                     format, new ByteArrayDeserializer(), new ByteArrayDeserializer());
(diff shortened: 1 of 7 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -208,11 +209,17 @@ private void sendToPartition(RoutingContext routingContext) {
     private void createConsumer(RoutingContext routingContext) {
         this.httpBridgeContext.setOpenApiOperation(HttpOpenApiOperations.CREATE_CONSUMER);
 
-        JsonObject body = routingContext.getBodyAsJson();
+        JsonObject bodyAsJson = null;
+        // TODO: it seems that getBodyAsJson raises an exception when the body is empty and not null
+        try {
+            bodyAsJson = routingContext.getBodyAsJson();
+        } catch (DecodeException ex) {
```
tombentley: I think this is dangerous, because I'm guessing `DecodeException` can be thrown in other circumstances (such as malformed JSON). So you should get the body in some other representation (is there a `getBodyAsString()`?) where you can check for the nullness of the body prior to calling `getBodyAsJson()`.

Also if you assign `bodyAsJson` an empty object rather than null in the case that the body is null you won't need to futz around adding `JsonObject requestBody` in `HttpSinkBridgeEndpoint.doCreateConsumer()`.

ppatierno: It's how it works even in the http sink and source but I see your point.

> Also if you assign bodyAsJson an empty object rather than null in the case that the body is null you won't need to futz around adding JsonObject requestBody in HttpSinkBridgeEndpoint.doCreateConsumer().

That was my first implementation but I was actually changing the "original" body which was something I didn't like. At the same time I didn't like to change the other env vars of course but actually changing the original one is something that could drive to collateral effects in general, maybe not in this case. I was trying to be a good developer :-)

ppatierno:
@tombentley I have pushed a different way for checking an empty body (at least in the `HttpBridge` class). I can do the same for sink and source removing the `DecodeException` catch for doing that. Anyway we need to catch `DecodeException` as you mentioned malformed JSON (even if the OpenAPI validation should protect from that, because the possible request bodies are well defined). If we decide to catch the exception anyway we need to reply with an error to a client not just continuing the processing as today. I would leave this for a different PR.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tombentley: You're not really "changing the original body", you're treating no body an synonymous with an empty (JSON) object, which doesn't care about the distinction. If it did care, then obviously you'd need to reflect that somehow. But another reason is the fewer nulls you pass around the less likely you are to be hit by an NPE due to a missing null check.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -72,8 +72,11 @@ public void doCreateConsumer(RoutingContext routingContext, JsonObject bodyAsJso
         // get the consumer group-id
         groupId = routingContext.pathParam("groupid");
 
+        // if body from HTTP request is empty (null) then set an empty JSON to use default behaviour/values
+        JsonObject requestBody = bodyAsJson != null ? bodyAsJson : new JsonObject();
```
tombentley: This shouldn't be needed now, right?

ppatierno: It is in a different point in the code.
I did the change in the HttpBridge class and not here yet. See the handle method in this class where the body is got. I have to do something similar as I did in HttpBridge.

tombentley: Ah, OK then.

PR #360, opened 2019-10-07
Fix log level and AMQP bridge

ppatierno:
This PR set the INFO log level for the bridge by default (instead of the current DEBUG one) and also disable the AMQP bridge by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ dist: xenial
 sudo: required
 language: java
 jdk:
-- oraclejdk8
-- oraclejdk11
+- openjdk8
+- openjdk11
 services:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,9 @@
+dist: xenial
 sudo: required
 language: java
 jdk:
-- oraclejdk8
-- oraclejdk11
+- openjdk8
+- openjdk11
```
stanlyDoge: Should this be in another PR?

ppatierno: could be but it fixes the current broken build

stanlyDoge: Ok

PR #361, opened 2019-10-10
Extend bridge tests to use a bridge instance running in the OS

granzoto:


src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
```diff
@@ -58,7 +60,7 @@ void createConsumerNameNotSet(VertxTestContext context) throws InterruptedExcept
                         JsonObject bridgeResponse = response.body();
                         String consumerInstanceId = bridgeResponse.getString("instance_id");
                         name = consumerInstanceId;
-                        assertTrue(consumerInstanceId.startsWith("kafka-bridge-consumer-"));
+                        assertTrue(consumerInstanceId.startsWith("kafka-bridge-consumer"));
```
ppatierno: Why removing the dash here?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
```diff
@@ -85,9 +85,11 @@ static void beforeAll() {
         kafkaCluster.start();
         vertx = Vertx.vertx();
 
-        bridgeConfig = BridgeConfig.fromMap(config);
-        httpBridge = new HttpBridge(bridgeConfig);
-        httpBridge.setHealthChecker(new HealthChecker());
+        if (System.getenv("STRIMZI_USE_SYSTEM_BRIDGE") == null) {
+            bridgeConfig = BridgeConfig.fromMap(config);
```
ppatierno: With external bridge we are not starting the bridge here and it is fine but what about config? Aren't tests which need info about the bridge config? Should we have the config of the external bridge?

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -216,6 +222,8 @@ void sendBinaryMessageWithKey(VertxTestContext context) {
         consumer.subscribe(topic, done -> {
             if (!done.succeeded()) {
                 context.failNow(done.cause());
+            } else {
+                context.completeNow();
```
ppatierno: This is wrong. If subscription fails that the test has to fail otherwise the test complete when the consumer gets the messages which happens in the previous handler. With this addition the test pass just because after subscription you are calling completeNow. The consumer doesn't have time to get messages that the test is shutdown and the handler code is not executed.

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -345,6 +353,8 @@ void sendMultipleMessages(VertxTestContext context) {
...
```
ppatierno: Same reason as described above

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -76,6 +76,8 @@ void sendSimpleMessage(VertxTestContext context) throws Throwable {
...
```
ppatierno: Same as described for another test

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -123,6 +125,8 @@ void sendSimpleMessageToPartition(VertxTestContext context) {
...
```
ppatierno: Same as described for another test

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -169,6 +173,8 @@ void sendSimpleMessageWithKey(VertxTestContext context) {
...
```
ppatierno: Same as described for another test

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
granzoto: right now we are using it only once in the test class ConsumerGeneratedNameTest, to setup a bridge, but that's not used for external bridge.
But we can improve the tests later to handle also the configs from the external bridge as well

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
granzoto: Typo. Fixed

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
granzoto: Removed

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
granzoto: Removed

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
granzoto: Removed

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
granzoto: Removed

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
ppatierno: do you mean that `ConsumerGeneratedNameTest` is not run with the external bridge?

granzoto:
> From where exactly the bridge is started? It's started manually before tests execution and tests just use URL to there? I think there should be some mechanism, which will run specific bridge from jar or at least env variable with URL for external bridge - https://github.com/Apicurio/apicurio-registry/pull/89/files#diff-845d93f2ac542f8639357cfcdbba7fa5R32.
> 
> Currently it will connect to the same URL every time `localhost:8080` which is not good approach since you can specify different port for the bridge.

I agree, This first PR is for allow us to run the test in an automated way using an external bridge. For future tests we might provide a way to config/deploy/tear down bridges using any provided jar or address.

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
granzoto: ok, removed wrong code, and disabling this test if using external bridge

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
granzoto: Automated tests will setup a bridge without bridgeId to run this specific test

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -90,6 +90,7 @@ void createConsumerWrongFormat(VertxTestContext context) throws InterruptedExcep
     }
 
     @Test
+    @DisabledIfEnvironmentVariable(named = "STRIMZI_USE_SYSTEM_BRIDGE", matches = "TRUE")
```
ppatierno: you are disabling the test checking that `STRIMZI_USE_SYSTEM_BRIDGE` env var is set to `TRUE` but you start the local bridge if the env var is null (so it's not set). It means that if it's set to `TRUE`, `FALSE`, `Paolo`, `Renato` or whatever it doesn't start the local bridge to use the external one. We should have consistency so disabling this test if env var is set, the value seems to be not important to us or checking that the env var is `FALSE` not just null when we decide to start local bridge or not.

granzoto: Ok, I am using the external bridge only if the env var STRIMZI_USE_SYSTEM_BRIDGE is set to TRUE.
Any other value or unset env var will use internal bridge.

ppatierno: the check you do when start the bridge or not is using ignoring case, while here we are using "TRUE". What happens with "true"?

ppatierno:
@Frawless @stanlyDoge do you have any more comments?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -214,5 +216,7 @@ void listConsumerSubscriptions(VertxTestContext context) throws InterruptedExcep
                 });
 
         assertTrue(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS));
+        consumerService()
```
Seequick1: It seems that we are deleting consumer in each test case or am i wrong ? Could we specify this in @AfterEach notation to reduce some code lines ? If it is problem now to reworked that you can create issue. What do you think ?

ppatierno: The problem is that consumers don't have always the same name so how to deal with it in a common @aftereach code? How to deal with it? @granzoto changed the consumer names in this PR to have more detailed information for each test.

src/test/java/io/strimzi/kafka/bridge/http/SeekTest.java
```diff
@@ -274,6 +274,13 @@ void seekToEndAndReceive(VertxTestContext context) throws InterruptedException,
     @Test
     @SuppressWarnings("checkstyle:MethodLength")
     void seekToOffsetAndReceive(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+
+        String name = "my-kafka-consumer-SeekOffset";
+        JsonObject jsonConsumer = new JsonObject()
+            .put("name", name)
+            .put("format", "json");
+
+
```
stanlyDoge: empty line

Issue #352, opened 2019-09-26
Migrate all tests asserts to `org.hamcrest.MatcherAssert.assertThat` 


Seequick1:
Currently our bridge tests has `org.junit.jupiter.api.Assertions.assertEqual`. Ideally, it would be great to have it in `org.hamcrest.MatcherAssert.assertThat`.

PR #362, opened 2019-10-14
Replace jupiter by hamcrest

stanlyDoge:

Fixes: https://github.com/strimzi/strimzi-kafka-bridge/issues/352

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,6 @@
 		<vertx.version>3.8.1</vertx.version>
 		<debezium.version>0.8.3.Final</debezium.version>
 		<maven.checkstyle.version>3.0.0</maven.checkstyle.version>
-		<jupiter.version>5.4.0</jupiter.version>
 		<junit.platform.version>1.4.0</junit.platform.version>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -154,6 +134,12 @@
 			<version>2.12.4</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.hamcrest</groupId>
+			<artifactId>hamcrest-all</artifactId>
+			<version>1.3</version>
```
ppatierno: avoid to use fixed version here, define a property instead as for other dependencies.

stanlyDoge:
It should be because of better failure messages: https://objectpartners.com/2013/09/18/the-benefits-of-using-assertthat-over-other-assert-methods-in-unit-tests/

ppatierno:
About failure messages, I agree but saying that `assertEquals(expected, actual);` is " very cryptic" ... :-D

ppatierno:
@Seequick1  @Frawless any more comments?

Seequick1:
Everything seems fine for me. LGTM

PR #363, opened 2019-10-16
Add support for tracing

ppatierno:
This PR adds support for tracing in the bridge using the "interceptor" approach.
Supported tracing implementation is just Jaeger for now.

As known issue related to the Jaeger UI, in case the bridge gets more messages through a single poll, the "poll" span is not shown for each incoming "From" message span.
I am in touch with Jaeger guys in order to open an issue to fix that.

### DIFF
--- a/config/application.properties
+++ b/config/application.properties
@@ -1,6 +1,7 @@
 #Bridge related settings
 bridge.id=my-bridge
-bridge.tracing=jaeger
+# uncomment the following line to enable Jaeger tracing
+#bridge.tracing=jaeger
 
 #Apache Kafka common
 kafka.bootstrap.servers=localhost:9092
--- a/config/application.properties
+++ b/config/application.properties
@@ -1,6 +1,6 @@
 #Bridge related settings
 bridge.id=my-bridge
-# uncomment the following line to enable Jaeger tracing
+# uncomment the following line to enable Jaeger tracing, check the documentation how to configure the tracer
 #bridge.tracing=jaeger
 
 #Apache Kafka common
--- a/config/application.properties
+++ b/config/application.properties
@@ -1,5 +1,6 @@
 #Bridge related settings
 bridge.id=my-bridge
+bridge.tracing=jaeger
 
 #Apache Kafka common
 kafka.bootstrap.servers=localhost:9092
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -5,6 +5,7 @@
 
 package io.strimzi.kafka.bridge;
 
+import io.opentracing.contrib.kafka.TracingConsumerInterceptor;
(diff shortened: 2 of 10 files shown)


config/application.properties
```diff
@@ -1,5 +1,6 @@
 #Bridge related settings
 bridge.id=my-bridge
+bridge.tracing=jaeger
```
scholzj: Should this be enabled by default?

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
```diff
@@ -102,6 +103,9 @@ public void open() {
         Properties props = new Properties();
         props.putAll(kafkaConfig.getConfig());
         props.putAll(kafkaConfig.getProducerConfig().getConfig());
+        if (this.bridgeConfig.getTracing() != null && "jaeger".equals(this.bridgeConfig.getTracing())) {
```
scholzj: I suspect that this will not be dependent on the tyke of the tracing client. Every client supporting OpenTracing will use this. So maybe `!= null` should be enough?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -151,6 +152,9 @@ protected void initConsumer(boolean shouldAttachBatchHandler, Properties config)
         props.putAll(kafkaConfig.getConfig());
         props.putAll(kafkaConfig.getConsumerConfig().getConfig());
         props.put(ConsumerConfig.GROUP_ID_CONFIG, this.groupId);
+        if (this.bridgeConfig.getTracing() != null && "jaeger".equals(this.bridgeConfig.getTracing())) {
```
scholzj: I suspect that this will not be dependent on the tyke of the tracing client. Every client supporting OpenTracing will use this. So maybe `!= null` should be enough?

src/main/resources/application.properties
```diff
@@ -1,5 +1,6 @@
 #Bridge related settings
 bridge.id=my-bridge
+bridge.tracing=jaeger
```
scholzj: Again, should this be enabled by default?

config/application.properties
ppatierno: What do you mean? I am kinf of reproducing what we have in the operator for enabling tracing which is not enabled by default right? You have this `tracing` property to set in the custom resource.

src/main/java/io/strimzi/kafka/bridge/SourceBridgeEndpoint.java
ppatierno: Yes you are right.

config/application.properties
scholzj: Yeah. ut the user has to add it there. Where as here you seem to add it as default into the configs in the repository. Or did I misunderstood how this file is used?

ppatierno: Now I got what you meant. You were saying that having it year will enable tracing by default which we don't want :) I will remove it

PR #364, opened 2019-10-17
Maven JAR plugin missing a version

stanlyDoge:


pom.xml
```diff
@@ -12,6 +12,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
+		<maven.jar.plugin.version>3.1.2</maven.jar.plugin.version>
```
scholzj: Maybe you can use the same format as for the other plugin? `maven-jar-plugin.version`?

PR #367, opened 2019-10-23
Added bridge tracing configuration from Strimzi cluster operator

ppatierno:
Signed-off-by: Paolo Patierno <ppatierno@live.com>

### DIFF
--- a/bin/docker/kafka_bridge_config_generator.sh
+++ b/bin/docker/kafka_bridge_config_generator.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [ "$STRIMZI_TRACING" = "jaeger" ]; then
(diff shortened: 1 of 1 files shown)


bin/docker/kafka_bridge_config_generator.sh
```diff
@@ -1,8 +1,13 @@
 #!/usr/bin/env bash
 
+if [ "$STRIMZI_TRACING" = "jaeger" ]; then
```
stanlyDoge: I think we should check just empty string. Consider adding different type of tracing in the future, you would have to change this script too.

ppatierno: good point, even because the bridge checks that the value is "jaeger" actually for checking the right env vars.

PR #371, opened 2019-11-06
Update hamcrest dependency

stanlyDoge:

Use version 2.2 instead of very old 1.3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
 		<debezium.version>0.8.3.Final</debezium.version>
 		<maven.checkstyle.version>3.0.0</maven.checkstyle.version>
 		<junit.platform.version>1.4.0</junit.platform.version>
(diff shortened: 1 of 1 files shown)


ppatierno:
@stanlyDoge just genuine question. Any specific reasons for this upgrade?

stanlyDoge:
I was alterted by QE guys the 1.3 is 7 years old so we should use a newer version.

ppatierno:
Let's try to align with the operator first and see if all works there.

ppatierno:
So before approving and merging this one, I would align with the operator as well. You could open a PR for that.

Frawless:
@ppatierno https://github.com/strimzi/strimzi-kafka-operator/pull/2156

ppatierno:
@Frawless ah! it's part of the huge JUnit 5 migration PR :-)

PR #376, opened 2019-12-19
Test if the kafka bridge version is displayed on startup

granzoto:
Signed-off-by: Renato Granzoto <rgranzot@redhat.com>

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
@@ -13,9 +13,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 
 import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
 
 public class StaticTest {
 
@@ -43,8 +41,6 @@ public class StaticTest {
     void bridgeVersionDisplayedInStartupTest() throws Exception {
 
         String kBVersion = getVersionFromFile("release.version");
-        assertThat(kBVersion, is(not("Unable to get version")));
-        Boolean versionFound = false;
 
         ProcessBuilder bridgeJar = new ProcessBuilder(
                "target/kafka-bridge-" + kBVersion + "/kafka-bridge-" + kBVersion + "/bin/kafka_bridge_run.sh",
--- a/src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
@@ -19,26 +19,30 @@ import static org.hamcrest.Matchers.is;
 
 public class StaticTest {
 
-    public String getVersionFromFile() {
+    public String getVersionFromFile(String releaseFile) throws Exception {
 
-        String versionFromFile = "Unable to get version";
+        String versionFromFile;
 
         try {
-            BufferedReader bufferedReader = new BufferedReader(new FileReader("release.version"));
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(releaseFile));
             versionFromFile = bufferedReader.readLine();
             bufferedReader.close();
         } catch (FileNotFoundException e) {
-            e.printStackTrace();
+            throw new Exception("File not found : " + releaseFile);
         } catch (IOException e) {
-            e.printStackTrace();
+            throw new Exception("Unable to open file : " + releaseFile);
+        }
+
+        if ((versionFromFile == null) || (versionFromFile.isEmpty())) {
+            throw new Exception("Unable to get Version from file : " + releaseFile);
         }
         return versionFromFile;
     }
 
     @Test
-    void bridgeVersionDisplayedInStartupTest() throws IOException, InterruptedException {
+    void bridgeVersionDisplayedInStartupTest() throws Exception {
 
-        String kBVersion = getVersionFromFile();
+        String kBVersion = getVersionFromFile("release.version");
         assertThat(kBVersion, is(not("Unable to get version")));
         Boolean versionFound = false;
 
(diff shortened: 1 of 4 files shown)


src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,79 @@
...
+        String procOutput;
+        while ((procOutput = bufferedInputReader.readLine()) != null) {
+
+            if ((procOutput.contains("Strimzi Kafka Bridge")) && (procOutput.contains("starting"))) {
+                assertThat(procOutput, containsString("Strimzi Kafka Bridge " + kBVersion + " is starting"));
+                versionFound = true;
+
+            // If we achieved this point, and the version was not displayed yet, test failed
+            } else if ((procOutput.contains("Starting HTTP-Kafka bridge verticle")) &&  (!versionFound)) {
+                assertThat("Test Failed", false);
```
Frawless: You should assert after destroying the bridge. This will end the test before cleanup phase. Not sure if start/stop would be better in before/after for that class.

Frawless:
It looks a little bit complicated to me now. I think the better approach is is to start bridge, collect few first lines of the long and put it into string, stop bridge, assert that string contains versions, @granzoto  wdyt?

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,79 @@
...
+            e.printStackTrace();
+        }
+        return versionFromFile;
+    }
+
+    @Test
+    void bridgeVersionDisplayedInStartupTest() throws IOException, InterruptedException {
+
+        String kBVersion = getVersionFromFile();
+        assertThat(kBVersion, is(not("Unable to get version")));
```
scholzj: Why not just throw some exception out of the `getVersionFromFile` method and let it fail the test? That would be IMHO more Java native solution than the `Unable to get version` text.

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,79 @@
...
+        InputStreamReader inputStreamReader = new InputStreamReader(bridgeProc.getInputStream());
+        BufferedReader bufferedInputReader = new BufferedReader(inputStreamReader);
+
+        int lineCount = 0;
+        String procOutput;
+        while ((procOutput = bufferedInputReader.readLine()) != null) {
+
+            if ((procOutput.contains("Strimzi Kafka Bridge")) && (procOutput.contains("starting"))) {
+                assertThat(procOutput, containsString("Strimzi Kafka Bridge " + kBVersion + " is starting"));
+                versionFound = true;
```
scholzj: Should we maybe add break here to simply stop the cycle instead of looping through the log even after we found the correct line?

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,77 @@
...
+            throw new Exception("Unable to get Version from file : " + releaseFile);
+        }
+        return versionFromFile;
+    }
+
+    @Test
+    void bridgeVersionDisplayedInStartupTest() throws Exception {
+
+        String kBVersion = getVersionFromFile("release.version");
+        assertThat(kBVersion, is(not("Unable to get version")));
```
scholzj: I think you don't need this assert anymore because you now throw exception.

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,77 @@
...
+        Boolean versionFound = false;
```
scholzj: I think this variable is not used anywhere anymore.

granzoto: ok, removed

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class StaticTest {
```
ppatierno: out of curiosity, why this name?

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public String getVersionFromFile(String releaseFile) throws Exception {
```
ppatierno: could it be a static method in a Utils class for the tests?

src/test/java/io/strimzi/kafka/bridge/http/StaticTest.java
granzoto: I set this name as StaticTest becasue I am setting the kafka bridge up and down using the the bridge jar file, not using the test classes. This Static Classes does not extends any test class, like HttpBridgeTestBase for instance

ppatierno: So this class expects a bridge already running not deploying one by itself. I have no great names in mind, at least please add some JavaDoc on the class.

granzoto: actually no, what I meant is that I am setting it up and down using the ProcessBuilder.
So I had to skip all the @afterall, @aftereach and @beforeall.
I will add the javadocs

src/test/java/io/strimzi/kafka/bridge/utils/KafkaBridgeVersion.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package io.strimzi.kafka.bridge.utils;
+
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class KafkaBridgeVersion {
```
ppatierno: I think that just an Utils class could be enough, not adding a specific class for every utility method we could need.

PR #382, opened 2020-01-31
Fix dependencies, so tests run

tombentley:
Signed-off-by: Tom Bentley <tbentley@redhat.com>


ppatierno:
I forced the Vert.x OpenAPI library version because in a newer version I am getting:

```
"Validation error on: body.records[0].value - $.records[0].value: string found, object expected"
```

even if declaration in OpenAPI is like following:

```json
"value": {
                        "oneOf": [
                            {
                                "type": "object"
                            },
                            {
                                "type": "string"
                            }
                        ]
                    },
```

and it was working before.
Maybe something to check in the OpenAPI library. /cc @slinkydeveloper

tombentley:
@ppatierno lgtm. Shall I merge it?

PR #395, opened 2020-03-12
Fixing missing handling not existing resources

### LINKED ISSUE #394: Message "null" when hitting a not existing resource
When the HTTP client sends a request to a non existing resource (i.e. `curl http://localhost:8080/` or `curl http://localhost:8080/not-existing`), the bridge replies with the following JSON:

```json
{"error_code":404,"message":null}
```

it contains the `404` error code but not the related message (i.e. Not Found in this case).

ppatierno:
This PR fixes #394

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -409,15 +409,20 @@ public class HttpBridge extends AbstractVerticle implements HealthCheckable {
             routingContext.request().path());
 
         String message = null;
-        // in case of validation exception, building a meaningful error message
-        if (routingContext.failure() != null && routingContext.failure() instanceof ValidationException) {
-            ValidationException validationException = (ValidationException) routingContext.failure();
-            StringBuilder sb = new StringBuilder();
-            if (validationException.parameterName() != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -409,25 +409,30 @@ private void errorHandler(RoutingContext routingContext) {
...
+                if (validationException.parameterName() != null) {
+                    sb.append("Validation error on: " + validationException.parameterName() + " - ");
+                }
+                sb.append(validationException.getMessage());
+                message = sb.toString();
             }
-            sb.append(validationException.getMessage());
-            message = sb.toString();
+        } else if (routingContext.statusCode() == HttpResponseStatus.NOT_FOUND.code()) {
+            message = HttpResponseStatus.NOT_FOUND.reasonPhrase();
```
tombentley: Why do you only set the non-null message for NOT_FOUND and BAD_REQUEST?

ppatierno: The error handler is set on the Vertx route only for these two errors. It means we are sure that it's called only in this two cases. All the others follow the normal flow.

PR #396, opened 2020-03-30
Upgrade SpotBugs version

stanlyDoge:

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -18,6 +18,7 @@ import io.opentracing.util.GlobalTracer;
 import io.strimzi.kafka.bridge.BridgeContentType;
 import io.strimzi.kafka.bridge.EmbeddedFormat;
 import io.strimzi.kafka.bridge.Endpoint;
+import io.strimzi.kafka.bridge.IllegalEmbeddedFormatException;
 import io.strimzi.kafka.bridge.SourceBridgeEndpoint;
 import io.strimzi.kafka.bridge.config.BridgeConfig;
 import io.strimzi.kafka.bridge.converter.MessageConverter;
@@ -72,7 +73,7 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
 
         String topic = routingContext.pathParam("topicname");
 
-        List<KafkaProducerRecord<K, V>> records;
+        List<KafkaProducerRecord<K, V>> records = new ArrayList<>();
         Integer partition = null;
         if (routingContext.pathParam("partitionid") != null) {
             try {
@@ -107,7 +108,9 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
         HttpTracingUtils.setCommonTags(span, routingContext);
 
         try {
-            records = messageConverter.toKafkaRecords(topic, partition, routingContext.getBody());
+            if (messageConverter != null) {
+                records = messageConverter.toKafkaRecords(topic, partition, routingContext.getBody());
+            }
 
             for (KafkaProducerRecord<K, V> record :records)   {
                 tracer.inject(span.context(), Format.Builtin.TEXT_MAP, new TextMap() {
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -72,7 +73,7 @@ public void handle(Endpoint<?> endpoint) {
 
         String topic = routingContext.pathParam("topicname");
 
-        List<KafkaProducerRecord<K, V>> records;
+        List<KafkaProducerRecord<K, V>> records = new ArrayList<>();
```
scholzj: MAbe I don't see the whole code ... but isn't it better to just for example return when `messageConverter == null` instead of creating the emepty arraylist which is never used?

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -108,9 +108,10 @@ public void handle(Endpoint<?> endpoint) {
         HttpTracingUtils.setCommonTags(span, routingContext);
 
         try {
-            if (messageConverter != null) {
-                records = messageConverter.toKafkaRecords(topic, partition, routingContext.getBody());
+            if (messageConverter == null) {
+                return;
```
ppatierno: Quick look, I don't think it's going to work. With the empty list we are going to procede anyway and having the possibility to reach line 170 where we are sending back a response to the client. Of course, it doesn't mean that empty records collection works. You should create a test with null messageConverter if possible and then finding a good way to send a response to the HTTP client that could make more sense. Anyway, it should be done in a different PR because I don't see it related to the spotbugs version, or?

stanlyDoge:
That will solve the bug which is found by spotbugs (`UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`)

ppatierno:
So spotbugs is right from its point of view but this cannot happen. Our `EmbeddedFormat` enum raises an exception when an invalid format type is provided on consumer creation and a specific HTTP reply is sent (see test `createConsumerWrongFormat`).
It's also true that is messageConverter is null (which as I said should not happen) in the `handle` then it's going to raise an NPE. So we could 1) just suppress this working or 2) checking the messageConverter is not null and if it is reply back to the HTTP client with the right error code.

scholzj:
@stanlyDoge I understand that SpotBugs is complaining about it. I'm just not sure this is the best solution if the value from the constructor is never used.

stanlyDoge:
I can just disable this warning if you think it is the best solution.

ppatierno:
@stanlyDoge I know that ignoring the error was my suggestion but I don't like it anymore even because who reads the code can argue that `messageConverter` to be null can happen (because they don't know the entire flow of the request). I would check that it's not null and if it is (should never happen) sending an appropriate HTTP response to the client. Maybe a 500 Internal Server Error with some useful message. Wdyt?

stanlyDoge:
@ppatierno  I am ok with that

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -215,6 +226,6 @@ private int handleError(Throwable ex) {
             case BINARY:
                 return (MessageConverter<K, V, Buffer, Buffer>) new HttpBinaryMessageConverter();
         }
-        return null;
+        throw new IllegalEmbeddedFormatException("Invalid format type.");
```
ppatierno, line 229: this should still be `return null`

ppatierno:
@stanlyDoge what's the status of this?

stanlyDoge:
All the comments are resolved so it is waiting just for the reviews I guess.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
ppatierno: @stanlyDoge there is still this comment. Is removing this code that causes spotbugs to warn?

stanlyDoge: I think it is stale. If you take a look at the updated code it is
```
if (messageConverter == null) {
   // this never happens, but let SB be happy
   sendCode500Response();
} else {
   //original code
   ...
}
```

ppatierno:
@scholzj anything to add on this?

PR #398, opened 2020-04-01
Removing the jackson binding override

ppatierno:
Currently we are overriding jackson binding (indirect dependency coming from debezium) with 2.10.2 version because of CVE-2019-14379.
The debezium 1.0.0.Final version seems to use jackson binding 2.10.0 which already fixes CVE-2019-14379 so we should not need the override anymore.
See https://github.com/advisories/GHSA-6fpp-rgj9-8rwc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,25 +23,12 @@
 		<maven-surefire-plugin.version>2.22.1</maven-surefire-plugin.version>
 		<swagger2markup.version>1.3.7</swagger2markup.version>
 		<jackson-core.version>2.10.2</jackson-core.version>
-		<jackson-databind.version>2.10.2</jackson-databind.version>
 		<spotbugs.version>3.1.12</spotbugs.version>
 		<strimzi-oauth.version>0.3.0</strimzi-oauth.version>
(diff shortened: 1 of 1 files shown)


ppatierno:
@tombentley I was even thinking about using 2.10.0 version for our jackson-core dependency but it means doing a step back from 2.10.2 currently used just for consistency but I doubt how much it makes sense. Wdyt?

tombentley:
Honestly, I think there are good reasons to keep this in. `jacskon-databind` suffers from a steady stream of CVEs, so we could remove this now only to need it back again next week in order to pick up a new version with a CVE fix.

ppatierno:
What I don't like is that we are overriding just because this is used by Debezium and not by us so we are poisoning our pom with not necessary stuff. Anyway I see a 2.10.3 already released and a 2.11 in RC. If Debezium project won't update it, we are going to have this override forever in our pom :-(

tombentley:
Does the bridge depend on debezium on compile scope?

ppatierno:
Absolutely not.

tombentley:
Then I guess we don't care so much.

PR #399, opened 2020-04-16
JSON payload test

im-konge:
Signed-off-by: Lukas Kral <lkral@redhat.com>

**Type of change**
- new test

**Description**
This PR will add test for JSON payload:

Send message that will contains: `key, value, partition` to Kafka Bridge endpoint _partition_ -> here should be verification error.
After that try to send string type message that will contains: `key, value` to Kafka Bridge endpoint _topic_ and the same thing but with JSON type message value.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
@@ -644,4 +644,82 @@ public class ProducerTest extends HttpBridgeTestBase {
                     context.completeNow();
                 });
     }
+
+    @Test
+    void jsonPayloadTest(VertxTestContext context) {
+        String kafkaTopic = "breakOpenApiRules";
+        kafkaCluster.createTopic(kafkaTopic, 3, 1);
+
+        String value = "Hello from the other side";
+        String key = "message-key";
+
+        JsonArray records = new JsonArray();
+        JsonObject json = new JsonObject();
+        json.put("value", value);
+        json.put("key", key);
+        json.put("partition", 0);
+        records.add(json);
+
+        JsonObject root = new JsonObject();
+        root.put("records", records);
+
+        producerService()
+            .sendRecordsToPartitionRequest(kafkaTopic, 0, root, BridgeContentType.KAFKA_JSON_JSON)
+            .sendJsonObject(root, verifyBadRequest(context, "Validation error on: body.records[0] - " +
+                "$.records[0].partition: is not defined in the schema and the schema does not allow additional properties"));
+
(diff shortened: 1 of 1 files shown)


src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -644,4 +644,82 @@ void sendMultipleRecordsWithOneInvalidPartitionTest(VertxTestContext context) {
...
+
+        JsonObject root = new JsonObject();
+        root.put("records", records);
+
+        producerService()
+            .sendRecordsToPartitionRequest(kafkaTopic, 0, root, BridgeContentType.KAFKA_JSON_JSON)
+            .sendJsonObject(root, verifyBadRequest(context, "Validation error on: body.records[0] - " +
+                "$.records[0].partition: is not defined in the schema and the schema does not allow additional properties"));
+
+        records.remove(json);
```
stanlyDoge, line 671: Just a nit: I would use `Json{Array|Object}.clear()` to be sure the object is empty. My memory is crappy and it is more complicated to follow what is in the object. However LGTM.

im-konge: Thanks for the comment and for the review :)

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -644,4 +644,84 @@ void sendMultipleRecordsWithOneInvalidPartitionTest(VertxTestContext context) {
...
+                    assertThat(ar.succeeded(), is(true));
+                    HttpResponse<JsonObject> response = ar.result();
+                    assertThat(response.statusCode(), is(HttpResponseStatus.OK.code()));
+
+                    JsonObject bridgeResponse = response.body();
+                    JsonArray offsets = bridgeResponse.getJsonArray("offsets");
+                    assertThat(offsets.size(), is(1));
+
+                    JsonObject metadata = offsets.getJsonObject(0);
+                    assertThat(metadata.getInteger("partition"), notNullValue());
```
ppatierno, line 692: if you are using the send to partition, you can assert that the partition is 0 (where are you sending) not just not null.

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -644,4 +644,84 @@ void sendMultipleRecordsWithOneInvalidPartitionTest(VertxTestContext context) {
...
```
ppatierno, line 722: ditto as above

PR #402, opened 2020-04-21
[MO] - [bridge fixes 1st] -> 1 bridge instance for 1 test suite

### LINKED ISSUE #404: New test for HTTP consumer creation with no name and no bridge id
Currently, we are testing the HTTP consumer creation and its related name in two scenarios:

* the consumer doesn't specify the name and the `bridge.id` is not set
* the consumer specifies its own name

It's missing a test related to the consumer doesn't providing a name and the `bridge.id` is set.

see-quick:

This PR adding the different way how to test the Bridge. Changes are as follows:
- execute 1 bridge instance for 1 test suite 
- replacing non-sematic name when we use external Kafka Bridge.

This PR fixes #404 as well.

Question:
When I creating consumer and do not specifically name sometimes in the test case I get different results:

1.run -> name of the consumer is "my-bridge-" + hash...
2.run -> name of the consumer is "kafka-bridge-consumer-" + hash....

Is this behaviour desired?


see-quick:
That question is relating to the errors in the Travis so we can consider it as some sort of flakiness.

ppatierno:
@see-quick regarding your question, the bridge properties file has an optional `bridge.id` property. When it's specified, the consumer name is the `bridge.id` + UUID, when it's not the name is "kafka-bridge-consumer-" + UUID. It's possible that some tests are using a properties file where you have `bridge.id=my-bridge`.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
```diff
@@ -54,19 +58,21 @@ void createConsumerNameNotSet(VertxTestContext context) throws InterruptedExcept
                 .as(BodyCodec.jsonObject())
                 .sendJsonObject(json, ar -> {
                     context.verify(() -> {
+                        LOGGER.info("Verifying that consumer name is created with random suffix");
```
ppatierno, line 61: at this point this test should be something like `createConsumerNameAndBridgeIdNotSet` to check that it starts with "kafka-bridge-consumer-" and it has a random suffix. Maybe we should havea  test where the consumer name is not set but the `bridge.id` is and checking that the consumer name is the `bridge.id` value plus a random suffix.

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
```diff
@@ -81,15 +80,21 @@ ProducerService producerService() {
     }
 
     @BeforeAll
-    static void beforeAll() {
+    static void beforeAll(VertxTestContext context) {
         kafkaCluster.start();
         vertx = Vertx.vertx();
 
-        if (!"TRUE".equalsIgnoreCase(System.getenv("STRIMZI_USE_SYSTEM_BRIDGE"))) {
+        LOGGER.info("Environment variable " + " EXTERNAL_BRIDGE:" + BRIDGE_EXTERNAL_ENV);
```
ppatierno, line 87: why the "+" sign between the two strings and not just one string?

src/test/java/io/strimzi/kafka/bridge/http/HttpBridgeTestBase.java
```diff
@@ -81,15 +80,21 @@ ProducerService producerService() {
...
+
+        if ("FALSE".equals(BRIDGE_EXTERNAL_ENV)) {
             bridgeConfig = BridgeConfig.fromMap(config);
             httpBridge = new HttpBridge(bridgeConfig);
             httpBridge.setHealthChecker(new HealthChecker());
+
+            LOGGER.info("Deploying in-memory bridge");
+            vertx.deployVerticle(httpBridge, context.succeeding(id -> context.completeNow()));
         }
+        // else we create external bridge from the OS invoked by `.jar`
```
ppatierno, line 97: is this a TODO?

see-quick: No, it is not. We will start .jar in the pipeline before tests will be triggered.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
see-quick: I agree with you, can you please open an issue for that? I will then work on that.

ppatierno: Here it is #404 (funny issue code!)

see-quick: Done I have also added a new test case which you described.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
```diff
@@ -5,47 +5,103 @@
...
 
+import static io.strimzi.kafka.bridge.http.HttpBridgeTestBase.TEST_TIMEOUT;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 
-public class ConsumerGeneratedNameTest extends HttpBridgeTestBase {
+@ExtendWith(VertxExtension.class)
+public class ConsumerGeneratedNameTest {
+
+    private static final Logger LOGGER = LogManager.getLogger(HttpBridgeTestBase.class);
```
ppatierno, line 47: wrong class name?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
```diff
@@ -54,24 +110,26 @@ void createConsumerNameNotSet(VertxTestContext context) throws InterruptedExcept
...
                         HttpResponse<JsonObject> response = ar.result();
+                        LOGGER.info("Response code from the Bridge is " + response.statusCode());
                         assertThat(response.statusCode(), is(HttpResponseStatus.OK.code()));
                         JsonObject bridgeResponse = response.body();
-                        String consumerInstanceId = bridgeResponse.getString("instance_id");
-                        name = consumerInstanceId;
-                        assertThat(consumerInstanceId.startsWith("kafka-bridge-consumer-"), is(true));
+                        consumerInstanceId = bridgeResponse.getString("instance_id");
+                        LOGGER.info("Consumer instance of the consumer is " + consumerInstanceId);
+                        assertThat(consumerInstanceId.startsWith("-"), is(true));
```
ppatierno, line 121: why changing this condition? Name is not set, bridge id is not set ... the instance id starts with "kafka-bridge-consumer-"

see-quick: changed the line 55 and let the bridge.id == null and not the empty string ("")

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
ppatierno: @see-quick it's github maybe, but I still see the wrong class name here.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1486,4 +1487,32 @@ void createConsumerWithInvalidFormat(VertxTestContext context) throws Interrupte
         create.get(TEST_TIMEOUT, TimeUnit.SECONDS);
         context.completeNow();
     }
+
+    @Test
+    void createConsumerNameIsNotSetAndBridgeIdIsSet(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno, line 1492: not sure why this test is in this file and not in `ConsumerGeneratedNameTest` with the other ones.

see-quick: the test that i added here is because of shared bridge. All the test cases in the ConsumerGeneratedNameTest has null bridge id and other ones has bridge id my-bridge

ppatierno: Yes right.

PR #403, opened 2020-04-22
[MO] - [bridge test rework 2nd] ->  major fixes + new clients

see-quick:
This PR adding:
1. support and little refactor with the cleaning phase, which costs race-conditions, topics already created etc.
2. I have replaced in HTTP package the Debezium dependency for the `Test containers`. Moreover, added AdminClientFacade for encapsulating the async calls. 
3. New Kafka Clients 
4. Changing the logic in the `HTTP package` whole suite.
   - previously: one test case -> one bridge, one Kafka cluster 
   - currently: all test suites -> one Kafka (HTTP package) for all test suites &  one bridge for one test suite


src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private final String bootstrapServer;
+
+    public BasicKafkaClient(String bootstrapServer) {
+        this.bootstrapServer = bootstrapServer;
+    }
+
+    /**
+     * Send messages to external entrypoint of the cluster with PLAINTEXT security protocol setting
+     * @return sent message count
```
ppatierno, line 36: more Javadoc with the parameters

src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public int sendMessagesPlain(long timeoutMs, String topicName, int messageCount, String message, int partition,
```
ppatierno, line 38: this client will send messages to the Kafka cluster where the bridge is connect as well. You are using "plain" just because maybe this code comes from the strimzi operator or because you want to add a TLS sender to the Kafka cluster? Do you see any value in this?

src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
```diff
@@ -0,0 +1,175 @@
...
+public class BasicKafkaClient {
+
+    private final String bootstrapServer;
+
+    public BasicKafkaClient(String bootstrapServer) {
+        this.bootstrapServer = bootstrapServer;
+    }
+
+    /**
+     * Send messages to external entrypoint of the cluster with PLAINTEXT security protocol setting
```
ppatierno, line 35: "external"? code from strimzi operator?

src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
```diff
@@ -0,0 +1,175 @@
...
+            "{\"Hello\" : \"World\"}", 0, false);
+    }
+
+    /**
+     * Receive messages to external entrypoint of the cluster with PLAINTEXT security protocol setting
+     * @return received message count
+     */
+    @SuppressWarnings("Regexp") // for the `.toLowerCase()` because kafka needs this property as lower-case
+    @SuppressFBWarnings("DM_CONVERT_CASE")
+    public int receiveMessagesPlain(long timeoutMs, String topicName, int messageCount) {
```
ppatierno, line 142: same as for producer about "plain" and "external"

src/test/java/io/strimzi/kafka/bridge/facades/AdminClientFacade.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Method createAsyncTopic used for the race condition between in-memory kafka cluster and also encapsulate the get
+     *
+     * @param topicName         name of the topic
+     * @param partitions        number of partitions
+     * @param replicationFactor number of replication factor
+     */
+    public void createAsyncTopic(String topicName, int partitions, int replicationFactor) throws ExecutionException, InterruptedException {
+        CreateTopicsResult createTopicsResult = adminClient.createTopics(Collections.singletonList(new NewTopic(topicName, partitions, (short) replicationFactor)));
+        createTopicsResult.all().get();
```
ppatierno, line 50: it's not "Async" anymore if you block with `get` on the future

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -617,21 +626,32 @@ void receiveFromMultipleTopics(VertxTestContext context) throws InterruptedExcep
...
-        String sentBody2 = "Simple message-2";
-        kafkaCluster.produce(topic1, sentBody1, 1, 0);
-        kafkaCluster.produce(topic2, sentBody2, 1, 0);
+        adminClientFacade.createAsyncTopic(topic1);
+        adminClientFacade.createAsyncTopic(topic2);
+
+        String message = "Simple message";
+
+//        kafkaCluster.produce(topic1, sentBody1, 1, 0);
+//        kafkaCluster.produce(topic2, sentBody2, 1, 0);
```
ppatierno, line 652: code to remove

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1047,10 +1086,12 @@ void offsetsConsumerDoesNotExist(VertxTestContext context) {
...
     void doNotRespondTooLongMessage(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
         String topic = "doNotRespondTooLongMessage";
-        kafkaCluster.createTopic(topic, 1, 1);
 
-        String sentBody = "Simple message";
-        kafkaCluster.produceStrings(topic, sentBody, 1, 0);
+        adminClientFacade.createAsyncTopic(topic);
+
+//        String sentBody = "Simple message";
+//        kafkaCluster.produceStrings(topic, sentBody, 1, 0);
```
ppatierno, line 1093: code to remove

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -222,14 +238,16 @@ void sendBinaryMessageWithKey(VertxTestContext context) {
         });
     }
 
+    @Disabled("FLAKY somethi")
```
ppatierno, line 241: what about this?

src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeTestBase.java
```diff
@@ -0,0 +1,150 @@
...
+    protected static Map<String, Object> config = new HashMap<>();
+
+    // for periodic/multiple messages test
+    protected static final int PERIODIC_MAX_MESSAGE = 10;
+    protected static final int PERIODIC_DELAY = 200;
+    protected static final int MULTIPLE_MAX_MESSAGE = 10;
+    protected static final int TEST_TIMEOUT = 60;
+    protected int count;
+
+    public static final KafkaContainer KAFKA_CONTAINER;
```
ppatierno, line 60: Why naming a variable/member like a constant?

src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeTestBase.java
```diff
@@ -0,0 +1,150 @@
...
+
+    protected static long timeout = 5L;
+
+    static {
+        KAFKA_CONTAINER = new KafkaContainer();
+        // this is needed because by default in kafka container we have __metrics_topic, which is not needed...
+        KAFKA_CONTAINER.withEnv("KAFKA_CONFLUENT_SUPPORT_METRICS_ENABLE", "false");
```
ppatierno, line 67: what is this about Confluent stuff?

src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
see-quick: Of course. In the future, we will be able to use TLS as well as tracing and OAuth clients.

src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
see-quick: fixed

src/test/java/io/strimzi/kafka/bridge/facades/AdminClientFacade.java
see-quick: Yeah, that's correct replaced.

src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeTestBase.java
see-quick: It is no longer needed because we have our own strimzi Kafka container. So ignore it... )

src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeTestBase.java
see-quick: Hmmm, for instance, we using KafkaContainer and it is semantically best name for this. Do you have any alternatives? Maybe, `KAFKA` or `STRIMZI_KAFKA` would better fit in this? @ppatierno

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
see-quick: I must investigate, why this test is flaky.

see-quick: It is caused  by PERIODIC_DELAY were set on 200ms,  and some messages were `inorder`. I have set that  PERIODIC_DELAY to 1s a now it's everything ok.

.travis.yml
```diff
@@ -4,8 +4,6 @@ language: java
 jobs:
   fast_finish: true
   include:
-  - os: linux
-    jdk: openjdk8
   - os: linux
```
ppatierno, line 7: My understanding is that we should not remove JDK 8 for building. We removed it just as runtime.

pom.xml
```diff
@@ -191,6 +189,47 @@
...
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+			<version>${junit5.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.strimzi</groupId>
+			<artifactId>test-container</artifactId>
+			<version>0.19.0-SNAPSHOT</version>
```
ppatierno, line 231: could you use `${project.version}`?

src/test/java/io/strimzi/kafka/bridge/http/AdminClientTest.java
```diff
@@ -66,11 +67,11 @@ void getTopicTest(VertxTestContext context) throws Exception {
                         for (int i = 0; i < 2; i++) {
                             JsonObject partition = partitions.getJsonObject(i);
                             assertThat(partition.getInteger("partition"), is(i));
-                            assertThat(partition.getInteger("leader"), is(1));
+                            assertThat(partition.getInteger("leader"), is(0));
                             JsonArray replicas = partition.getJsonArray("replicas");
                             assertThat(replicas.size(), is(1));
                             JsonObject replica0 = replicas.getJsonObject(0);
-                            assertThat(replica0.getInteger("broker"), is(1));
+                            assertThat(replica0.getInteger("broker"), is(0));
```
ppatierno, line 74: why this change? the test was failing?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
```diff
@@ -40,33 +41,35 @@
...
-    private static final KafkaFacade KAFKA_CLUSTER;
+    private static final StrimziKafkaContainer KAFKA_CLUSTER;
 
     static {
         config.put(KafkaConfig.KAFKA_CONFIG_PREFIX + ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
         config.put(KafkaConsumerConfig.KAFKA_CONSUMER_CONFIG_PREFIX + ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
         config.put(KafkaProducerConfig.KAFKA_PRODUCER_CONFIG_PREFIX + ProducerConfig.MAX_BLOCK_MS_CONFIG, "10000");
         config.put(HttpConfig.HTTP_CONSUMER_TIMEOUT, 5L);
-        KAFKA_CLUSTER = new KafkaFacade();
+        KAFKA_CLUSTER = new StrimziKafkaContainer("latest-kafka-2.4.0");
```
ppatierno, line 62: hardcoding the version means that we have to update this every time? And 2.4.0 is already old :-)

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -494,7 +496,7 @@ void receiveSimpleMessage(VertxTestContext context) throws InterruptedException,
                     long offset = jsonResponse.getLong("offset");
 
                     assertThat(kafkaTopic, is(topic));
-                    assertThat(value, is(sentBody));
+                    assertThat(value, is(sentBody + "-0"));
```
ppatierno, line 499: who is adding this "-0" ? tbh who reads the test gets confused and I am the first. I read that `sentBody` is "Simple message" at the top but then you are comparing with the same with "-0" suffix. I think it's really confusing.

ppatierno:
I had a review but I think there are things confusing the reader. I also noticed that part of the Consumer and Producer stuff comes from Strimzi operator, right? We should have a common place and not duplicating code I think.

src/test/java/io/strimzi/kafka/bridge/http/AdminClientTest.java
see-quick: Correct.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameTest.java
see-quick: Good catch..replaced by empty constructor where we will guarantee the latest version of Kafka.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
see-quick: Inside our vertex producer, we are adding this suffix. If we send more than 1 message it is better to check for instance `Simple message-50` that is the latest instead of the same `Simple message`.  I can rename that variable `sentBody` to `message`

see-quick:
These clients are inspired by the Strimzi's operators but they are not the same. Specifically, they have other options like sendJson, sendNullKeyRecord and so on. Would it be ok if I create a task for this refactor to merge these two sets of clients into one because I really would like to set up the automating pipeline for the Bridge and I am only waiting for this PR :D

src/test/java/io/strimzi/kafka/bridge/http/AdminClientTest.java
ppatierno: Really? I rebuilt the bridge a lot of times during these days for release without getting any error ...

pom.xml
```diff
@@ -19,19 +19,30 @@
 		<junit.platform.version>1.5.2</junit.platform.version>
 		<hamcrest.version>2.2</hamcrest.version>
 		<maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
-		<maven-surefire-plugin.version>2.22.1</maven-surefire-plugin.version>
+		<maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
```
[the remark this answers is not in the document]
see-quick: Yes, for sure.

see-quick:
This is for partition -> `[{"partition":0,"leader":0,"replicas":[{"broker":0,"leader":true,"in_sync":true}]},{"partition":1,"leader":0,"replicas":[{"broker":0,"leader":true,"in_sync":true}]}]`
This is for replicas -> `[{"broker":0,"leader":true,"in_sync":true}]`

I was getting error because broker + and leader value was never set to 1 but `0`.

ppatierno:
@see-quick what I am asking is why these tests are not failing in Travis with the current code? I rebuilt the bridge tons of times these days and they always passed, no need to change from 1 to 0. So I am wondering if there is something worst underneath.

see-quick:
@ppatierno OK, now this PR is ready to review again ! :D

see-quick:
we discusseed your last question but to have answer here. It is because test-container using broker.id = 0 and debezium kafka has broker.id = 1

pom.xml
```diff
@@ -19,19 +19,30 @@
...
 		<swagger2markup.version>1.3.7</swagger2markup.version>
 		<jackson-core.version>2.10.2</jackson-core.version>
 		<spotbugs.version>4.0.1</spotbugs.version>
 		<strimzi-oauth.version>0.5.0</strimzi-oauth.version>
 		<jaeger.version>1.1.0</jaeger.version>
 		<opentracing.version>0.33.0</opentracing.version>
 		<opentracing-kafka-client.version>0.1.12</opentracing-kafka-client.version>
+		<test-containers.version>1.14.0</test-containers.version>
```
ppatierno, line 30: do we still need this version? Aren't we using the test-containers project from the new repo now?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -129,9 +131,11 @@ void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwabl
 
     @Test
     void subscriptionConsumerDoesNotExistBecauseNotCreated(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
-        String topic = "subscriptionConsumerDoesNotExistBecauseNotCreated";
-        kafkaCluster.createTopic(topic, 1, 1);
         String name = "my-kafka-consumer-does-not-exists-because-not-created";
+        String topic = "subscriptionConsumerDoesNotExist";
```
ppatierno, line 135: why changing the name?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
ppatierno: tbh whatever the name is I still don't think it's clear for who reads the test. From outside, no one knows that the basicKafkaClient is going to add a suffix. So I read the test expecting "Simple message" but you are testing with "Simple message-0". At this point the interface of the basicKafkaClient is not clear: I pass a `sentBody` parameter, asking for sending "Simple message" but the client actually sends "Simple message-0". It should be changed tbh.

see-quick:
Ok, I have changed it.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
see-quick: Honestly, I do not remember but it is not so important i would say.

ppatierno: So we can revert back to having the name as the method? I think it's the rule for other tests.

see-quick: done

Issue #183, opened 2019-05-21
Support record headers for HTTP


tombentley:
The bridge should support record headers for HTTP. Note there is a [different issue](https://github.com/strimzi/strimzi-kafka-bridge/issues/69) discussing headers for AMQP.

ppatierno:
Closing because it's now available thanks to #413

PR #413, opened 2020-05-25
Kafka headers support for HTTP

codeJack:
Purpose of the change is to allow usage of [Kafka Headers](https://issues.apache.org/jira/browse/KAFKA-4208) within strimzi-kafka-bridge HTTP API.
Kafka Headers would be attached to the [ConsumerRecord](https://strimzi.io/docs/bridge/latest/#_consumerrecord) and the [ProducerRecord](https://strimzi.io/docs/bridge/latest/#_producerrecord) definitions as an _optional_ array of key/value pairs.

Example

```
"kafkaRecord": {
  "key": "key1",
  "partition": 0,
  "value": "value1",
  "headers": [
    {
      "key": "key1",
      "value": "value1"
    },
    {
      "key": "key2",
      "value": "value2"
    }
  ]
}
```

 The feature is also tracked by this issue : https://github.com/strimzi/strimzi-kafka-bridge/issues/183

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
@@ -12,8 +12,11 @@ import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecords;
+import io.vertx.kafka.client.producer.KafkaHeader;
 import io.vertx.kafka.client.producer.KafkaProducerRecord;
+import io.vertx.kafka.client.producer.impl.KafkaHeaderImpl;
 
+import javax.xml.bind.DatatypeConverter;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -25,6 +28,7 @@ public class HttpJsonMessageConverter implements MessageConverter<byte[], byte[]
         Integer partitionFromBody = null;
         byte[] key = null;
         byte[] value = null;
+        List<KafkaHeader> headers = new ArrayList<>();
 
         JsonObject json = message.toJsonObject();
 
@@ -34,6 +38,15 @@ public class HttpJsonMessageConverter implements MessageConverter<byte[], byte[]
             }
             if (json.containsKey("value")) {
                 value = Json.encodeToBuffer(json.getValue("value")).getBytes();
+            }
+            if (json.containsKey("headers")) {
+                for (Object obj: json.getJsonArray("headers")) {
+                    JsonObject jsonObject = (JsonObject) obj;
+                    headers.add(new KafkaHeaderImpl(
+                        jsonObject.getString("key"),
+                        Buffer.factory.buffer(
+                            DatatypeConverter.parseBase64Binary(jsonObject.getString("value")))));
+                }
             }
             if (json.containsKey("partition")) {
                 partitionFromBody = json.getInteger("partition");
@@ -47,6 +60,7 @@ public class HttpJsonMessageConverter implements MessageConverter<byte[], byte[]
         }
 
         KafkaProducerRecord<byte[], byte[]> record = KafkaProducerRecord.create(kafkaTopic, key, value, partitionFromBody);
+        record.addHeaders(headers);
 
         return record;
     }
@@ -90,6 +104,19 @@ public class HttpJsonMessageConverter implements MessageConverter<byte[], byte[]
             jsonObject.put("partition", record.partition());
             jsonObject.put("offset", record.offset());
 
+            if (!record.headers().isEmpty()) {
+                JsonArray headers = new JsonArray();
+
+                for (KafkaHeader kafkaHeader: record.headers()) {
+                    JsonObject header = new JsonObject();
+
+                    header.put("key", kafkaHeader.key());
+                    header.put("value", DatatypeConverter.printBase64Binary(kafkaHeader.value().getBytes()));
+
+                    headers.add(header);
+                }
+                jsonObject.put("headers", headers);
+            }
             jsonArray.add(jsonObject);
         }
 
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -1273,13 +1273,62 @@
     },
     "components": {
         "schemas": {
+            "KafkaHeader": {
+                "example": {
+                    "key": "key1",
+                    "value": "dmFsdWUx"
+                },
+                "properties": {
+                    "key": {
+                        "type": "string"
+                    },
+                    "value": {
+                        "format": "byte",
+                        "type": "string",
+                        "description": "The header value in binary format, base64-encoded"
+                    }
+                },
+                "title": "KafkaHeader",
+                "required": [
+                    "value",
+                    "key"
+                ],
+                "type": "object"
+            },
+            "KafkaHeaderList": {
+                "title": "KafkaHeaderList",
+                "type": "array",
+                "items": {
+                    "$ref": "#/definitions/KafkaHeader"
+                },
+                "example": [
+                    {
+                        "key": "key1",
+                        "value": "dmFsdWUx"
+                    },
+                    {
+                        "key": "key2",
+                        "value": "dmFsdWUy"
+                    }
+                ]
+            },
             "ConsumerRecord": {
                 "example": {
                     "key": "key1",
                     "offset": 2,
                     "partition": 0,
                     "topic": "topic",
-                    "value": "value1"
+                    "value": "value1",
+                    "headers": [
+                        {
+                            "key": "key1",
+                            "value": "dmFsdWUx"
+                        },
+                        {
+                            "key": "key2",
+                            "value": "dmFsdWUy"
+                        }
+                    ]
                 },
                 "properties": {
                     "key": {
@@ -1298,6 +1347,9 @@
                     },
                     "value": {
                         "type": "string"
+                    },
+                    "headers": {
+                        "$ref": "#/definitions/KafkaHeaderList"
                     }
                 },
                 "title": "ConsumerRecord",
@@ -1563,10 +1615,27 @@
                                 "type": "string"
                             }
                         ]
+                    },
+                    "headers": {
+                        "$ref": "#/definitions/KafkaHeaderList"
                     }
                 },
                 "additionalProperties": false,
-                "example": "{\n    \"key\": \"key\"\n    \"partition\": 23\n    \"value\": {\"foo\": \"bar\"}\n}"
+                "example": {
+                    "key": "key1",
+                    "partition": 0,
+                    "value": "value1",
+                    "headers": [
+                        {
+                            "key": "key1",
+                            "value": "dmFsdWUx"
+                        },
+                        {
+                            "key": "key2",
+                            "value": "dmFsdWUy"
+                        }
+                    ]
+                }
             },
             "ProducerRecordList": {
                 "title": "ProducerRecordList",
--- a/src/test/java/io/strimzi/kafka/bridge/clients/Producer.java
+++ b/src/test/java/io/strimzi/kafka/bridge/clients/Producer.java
@@ -5,6 +5,7 @@
 package io.strimzi.kafka.bridge.clients;
 
 import io.vertx.core.Vertx;
+import io.vertx.kafka.client.producer.KafkaHeader;
 import io.vertx.kafka.client.producer.KafkaProducer;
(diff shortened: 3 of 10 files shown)


codeJack:
Looks like CI failed because of no space left on device
image

scholzj:
I restarted the build -> maybe Travis cleaned the disks in the mean time. If not we will estart it later. I passed on amd64, so I think thsi should not indicate any issue with this PR.

codeJack:
Thanks @scholzj.
CI went a bit further, that said the failing test cases are not failing in local.
Could it be that they are unstable or do I have to look into the way I validated locally ?
Am I allowed to restart the build ?

scholzj:
TBH, I'm not sure how Travis decides who can and connot restart the builds ... seems a bit random sometimes :-). But I wouldn't worry about it for now.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
```diff
@@ -25,6 +28,7 @@
         Integer partitionFromBody = null;
         byte[] key = null;
         byte[] value = null;
+        MultiMap headers = MultiMap.caseInsensitiveMultiMap();
```
ppatierno, line 31: why not filling a `List<KafkaHeader>` that the Vert.x client provides and then adding the list to the record at the end instead of filling this map and then traversing it for adding one header at time?

src/main/resources/openapi.json
```diff
@@ -1104,13 +1104,60 @@
...
+                "example": {
+                    "key": "key1",
+                    "value": "value1"
+                },
+                "properties": {
+                    "key": {
+                        "type": "string"
+                    },
+                    "value": {
+                        "type": "string"
```
ppatierno, line 1117: I would like to start a discussion around this.
As for the Kafka protocol, the header value is just an array of bytes so it could contain binary data as well and not just string. Did you think about that? Is there any specific reason you decided to support string only? for your use case?
As for the value message body, the current bridge support JSON and binary encoding which is specified in the `format` field when a consumer is created. Should we provide support for a more generic possible way to encode header value as well? Just thinking aloud ..... @stanlyDoge @scholzj wdyt?

src/test/java/io/strimzi/kafka/bridge/facades/KafkaFacade.java
```diff
@@ -74,6 +93,20 @@ public void produce(String topic, byte[] bytes, int messageCount, int partition)
         produce.get(OPERATION_TIMEOUT, TimeUnit.SECONDS);
     }
 
+    public void produce(String topic, byte[] bytes, int messageCount, int partition, MultiMap headers) throws InterruptedException, ExecutionException, TimeoutException {
+        CompletableFuture<Boolean> produce = new CompletableFuture<>();
+
+        Headers kafkaHeaders = new RecordHeaders();
+        for (Map.Entry<String, String> header: headers) {
+            kafkaHeaders.add(new RecordHeader(header.getKey(), header.getValue().getBytes()));
+        }
```
ppatierno, line 102: maybe you can use `headers.forEach` ?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -565,6 +628,72 @@ void receiveBinaryMessage(VertxTestContext context) throws InterruptedException,
         assertThat(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS), is(true));
     }
 
+    @Test
+    void receiveBinaryMessageWithHeaders(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
```
ppatierno, line 632: I don't know how much it makes sense to have a different test with binary value when it doesn't impact the headers. Wdyt @see-quick ?

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java
codeJack: Indeed, I have update the PR accordingly.

src/main/resources/openapi.json
codeJack: String was enough for my use case, but indeed it might not be generic enough.
I can add a `format` field (default should probably be` String` ?) at `KafkaHeader` level that would qualify the format of the header value.

```
"kafkaRecord": {
  "key": "key1",
  "partition": 0,
  "value": "value1",
  "headers": [
    {
      "key": "key1",
      "value": "string default format"
    },
    {
      "key": "key2",
      "format": "string"
      "value": "string format"
    },
    {
      "key": "key3",
      "format": "binary"
      "value": "<base 64 encoded bytes>"
    }
  ]
}
```

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
codeJack: Given that I updated both [HttpJsonMessageConverter](https://github.com/strimzi/strimzi-kafka-bridge/blob/master/src/main/java/io/strimzi/kafka/bridge/http/converter/HttpJsonMessageConverter.java) and [HttpBinaryMessageConverter](https://github.com/strimzi/strimzi-kafka-bridge/blob/master/src/main/java/io/strimzi/kafka/bridge/http/converter/HttpBinaryMessageConverter.java) testing both cases was a way to increase coverage.

see-quick: Overall, if currently there are two different classes HttpJsonMessageConverter, HttpBinaryMessageConverter with the related logic now it makes sense. I do not know how were tests before but now it looks good.

src/main/resources/openapi.json
codeJack: Yes, I guess I got your point  : sticking to Kafka's wire format for headers  - that would be raw binary - as  the official Kafka APIs do not provided facilities to serialize / deserialize Headers in the first place. 
In other words you'd like the Strimzi HTTP bridge to provide the same level of service as the official APIs for what concerns headers.

While this makes perfectly sense, in real life headers will be mostly used to express messaging metadata and I assume that this would be commonly done in the form of key/value String pairs; an hypothetical base64 encoding / decoding overhead is minimal but I guess it would be handy to benefit from human readable headers within the bridge interfaces to further ease investigations and testing.

scholzj:
@ppatierno What do we need to do here to move this PR forward?

ppatierno:
@scholzj we started a discussion about how to bring headers (which kind of data/format) here: https://github.com/strimzi/strimzi-kafka-bridge/pull/413#discussion_r429988641
I was hoping for feedback from you and @stanlyDoge to make a final decision.

ppatierno:
@codeJack can you fix the conflicts, please?

scholzj:
@ppatierno I don't have any strong opinions. But I guess it should follow the same format rules as the payload or key which can also have different formats in a single message. So I guess right now we support either both in JSON format or both in _binary_ (base64 encoded form). So similarly we should handle the headers?

ppatierno:
I was making the point that native Kafka clients don't make any assumption about headers values, they are just bytes. With payload, even if it's really just bytes as well, you have serialize/deserialize on top of it so somehow you have a way to interpret your data in a specific format.
I have no strong opinion as well, for this reason I was raising the discussion.
If we decide to support JSON/binary do you see it defined for each header (so the client can mix header JSON values with binary values) with `format` field or we should be just tied to the "format" specified when the consumer is created (which is related to the payload).
Anyway, let's remember that for the @codeJack use case, bringing just strings in the header values means JSON strings so double quotes has to be part of the value to be a valid JSON string.

codeJack:
@scholzj @ppatierno 

I can fix the conflicts, but I'd also like to align the PR to the final API design and I'm not sure we have reached an agreement yet.

Personally I don't have a strong opinion, both proposals have their pros and cons and for my own personal use-case - where I would mostly have _String_ based headers - it would be acceptable to base64-encode them upfront if we go for the **binary only** solution.

To summarize
- **binary only** would stick to the official APIs, offering the same level of service while being slightly less user friendly
- **String/binary** would be easier to use but, header-wise, it would not fully adhere to the currently existing Kafka APIs

If we go for the **String/Binary** solution I think we should specify the `format` at header level; if we define a single `format` that applies to all headers, as soon as you have a _binary_ one, you would need to to _base64_-encode all of your _String_ headers. Having to do that would kind of defeat the purpose of having a flexible - _String_ and _binary_ compatible - interface.

scholzj:
@ppatierno Any thoughts on this?

ppatierno:
> But I guess it should follow the same format rules as the payload or key which can also have different formats in a single message.

Actually it's not possible. When you create a consumer you specify the embedded format as binary or json. From that point in time, the consumer can get messages with key and value all binary or all json.
If we want to follow the same, then I think that for the headers we don't need to specify a format at all. We can just use the one on consumer creation. In this way, key/value in header follows the same rule as key/value in the payload.

@scholzj @codeJack does it sound reasonable to you?

scholzj:
Expecting the headers to be always in JSON seems a bit weird. The Kafka Java clients seem to handle this as String key and byte[] value. I wonder if we should somehow follow this although in reality it would be quite often that the value would be some String, Integer etc.

ppatierno:
@scholzj so you are saying what I raised few comments before? https://github.com/strimzi/strimzi-kafka-bridge/pull/413#discussion_r432604476

scholzj:
Yeah, looks like it :-). I guess handling the headers always as base64 encoded byte[] is the simple option ... but it is obviously user unfriendly. I wonder if we can have some autodetection of common types and handle all others as Base64. But no idea how easy / complicated that is and what would be the performance cost when you do it _per message_.

ppatierno:
@codeJack any more thoughts on this about having the headers as always raw bytes (as per Kafka native behavior) but of course Base64 encoded? I know that for taking just  a String is not the best solution (you have to encode your string in Base64) but I think we reflect the real behavior from the native Kafka clients. Wdyt?

scholzj (commented):
LGTM. Thanks for the PR.

src/main/resources/openapi.json
```diff
@@ -1273,13 +1273,62 @@
...
+                    }
+                },
+                "title": "KafkaHeader",
+                "required": [
+                    "value",
+                    "key"
+
+
```
ppatierno, line 1294: can we remove these extra lines?

src/test/java/io/strimzi/kafka/bridge/clients/Producer.java
```diff
@@ -148,4 +132,44 @@ private Properties fillDefaultProperties() {
     public Properties getProperties() {
         return properties;
     }
+
+    public static class Builder {
```
ppatierno, line 136: I would name it `ProducerBuilder`

src/test/java/io/strimzi/kafka/bridge/clients/Producer.java
```diff
@@ -148,4 +132,44 @@ private Properties fillDefaultProperties() {
...
+
+        public Builder properties(Properties properties) {
+            this.properties = properties;
+            return this;
+        }
+
+        public Builder headers(List<KafkaHeader> headers) {
+            this.headers = headers;
+            return this;
+        }
```
ppatierno, line 164: should `headers` and `properties` prefixed with `with` (as for the null key record)?

src/main/resources/openapi.json
ppatierno: @codeJack now that we agreed on the base64 encoded value, we should document it.
I think that in the "value" and alongside "type", the OpenAPI spec should support a "description" to make it clear. Wdyt?

ppatierno (commented):
@codeJack I had another pass.
@Frawless would you mind to take a look at the tests' part, please? Maybe you have a different opinion on the builder pattern added here.

Frawless (commented):
@ppatierno I agree with your suggestions

codeJack:
@ppatierno 
Can't reply directly to your comment about adding a description for the value field, so I will do it here :
I have also added the `"format": "byte"` qualifier as per [openapi specifications](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#data-types): whenever coupled with `"type":"string"` it stands for _base64 encoded characters_.

Also fixed the other points!

ppatierno (approved):
LGTM! Thanks for the PR!

ppatierno:
@codeJack let's wait for others approval ;-)

stanlyDoge (approved):
Thanks for your contribution. LGTM

stanlyDoge:
Please regenerate the doc to make build passing. 
```
  make docu_api \
```

scholzj (approved):

Issue #375, opened 2019-12-16
Add an endpoint to display the bridge version.


granzoto:
Nowadays the Kafka bridge only displays its version in the startup.  It would be nice to have an endpoint to display its version.
Let's suppose we have a new feature available only on versions > 0.13.0.  With this /version endpoint a client can easily check if the version he is connecting matches the minimal version for that feature.

ppatierno:
We already expose the OpenAPI specification on the `/openapi` endpoint. Of course, it's all about the API and doesn't contain any details about the bridge. For example, the `info.version` field is the version of the API that could be not the same as the version of the bridge implementing that API version.
Said that we have two ways for doing this:

* using so-called "OpanAPI Extensions" (https://swagger.io/docs/specification/openapi-extensions/) adding a `x-bridge-version` (under `info` section) so that the client can just queries the `/openapi` endpoint getting the bridge versin from it.
* exposing a new `/version` endpoint.

The problem with the first solution is about having the OpenAPI spec files aligned with the version we are releasing (maybe we should automate "something" via make), while the second solution is fully dynamic.

elakito:
Hi,
I thought about this sometime ago. Maybe we could use the root path `GET /` to return a version file like what Kafka-connect returns at `GET /`? Confluent's Kafka-rest returns an empty json while Strimzi Kafka-bridge currently returns `404 not found`. I think returning something other than 404 would be good and we could return the version info.

Regarding the openapi spec approach, I have a related question. I wanted to use a template file and inject other values (e.g., for basePath) so that when it is hosted behind a gateway, we could inject the correct basePath when the gateway is passing x-forwarded-prefix header to the bridge host. If we take this approach, we can also inject the version into the document.

ppatierno:
@elakito but the openapi json file is a resource so what you said should happen at compile time when building the jar. So if you want to change the `basePath` you should recompile the bridge jar. Did I get it right what you mean?

ppatierno:
thinking more I would agree with @elakito about using the root `/` for getting information about the bridge which could be initially just the version. @stanlyDoge wdyt? I see you are working on a PR for this.

stanlyDoge:
Yes, probably it is a better idea to have this prepared for more info entries, not just version.

elakito:
it sounds good about using the root path to provide the bridge info.

[...]

PR #422, opened 2020-06-09
Add / endpoint

stanlyDoge:

Fixes https://github.com/strimzi/strimzi-kafka-bridge/issues/375

This PR adds new endpoint (`/`) where the bridge instance info will be exposed. Currently, just the version of the bridge is there.

When the bridge is built with maven, `Class.class.getPackage().getImplementationVersion()` fetches a version from pom file. Unit test from IDE would fail (the null value would be there) so if the version is null, `"null"` string is returned.

### DIFF
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -14,13 +14,24 @@ Retrieves information about the Kafka Bridge instance, in JSON format.
 [options="header", cols=".^2a,.^14a,.^4a"]
 |===
 |HTTP Code|Description|Schema
-|**200**|Information about Kafka Bridge instance.|string
+|**200**|Information about Kafka Bridge instance.|<<_bridgeinfo,BridgeInfo>>
 |===
 
 
 ==== Produces
 
 * `application/json`
+
+
+==== Example HTTP response
+
+===== Response 200
+[source,json]
+----
+{
+  "bridge_version" : "0.16.0"
+}
+----
 
 
 [[_createconsumer]]
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -1274,4 +1274,25 @@ __required__|List of records to send to a given topic partition, including a val
 ----
 
 
+[[_version]]
+=== GET /version
+
+==== Description
+Retrieves the Kafka Bridge instance version in JSON format.
+
+
+==== Responses
+
+[options="header", cols=".^2a,.^14a,.^4a"]
+|===
+|HTTP Code|Description|Schema
+|**200**|Info about Kafka Bridge instance version.|string
+|===
+
+
+==== Produces
+
+* `application/json`
+
+
 
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -2,6 +2,27 @@
 [[_paths]]
 == Paths
 
+[[_configuration]]
+=== GET /
+
+==== Description
+Retrieves the Kafka Bridge instance version in JSON format.
+
+
+==== Responses
+
+[options="header", cols=".^2a,.^14a,.^4a"]
+|===
+|HTTP Code|Description|Schema
+|**200**|Info about Kafka Bridge instance configuration.|string
+|===
+
+
+==== Produces
+
+* `application/json`
+
+
 [[_createconsumer]]
 === POST /consumers/{groupid}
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -151,6 +152,7 @@ public void start(Promise<Void> startPromise) throws Exception {
                 routerFactory.addHandlerByOperationId(this.HEALTHY.getOperationId().toString(), this.HEALTHY);
                 routerFactory.addHandlerByOperationId(this.READY.getOperationId().toString(), this.READY);
                 routerFactory.addHandlerByOperationId(this.OPENAPI.getOperationId().toString(), this.OPENAPI);
+                routerFactory.addHandlerByOperationId(this.CONFIGURATION.getOperationId().toString(), this.CONFIGURATION);
```
ppatierno, line 155: I would not call it `CONFIGURATION`. For example, the version is not a configuration parameter. I would use `INFO` or something like that.

src/main/resources/openapi.json
```diff
@@ -1100,6 +1100,24 @@
                 "operationId": "openapi",
                 "description": "Retrieves the OpenAPI v2 specification in JSON format."
             }
+        },
+        "/": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "Info about Kafka Bridge instance configuration.",
```
ppatierno, line 1108: it's not just configuration

src/main/resources/openapi.json
```diff
@@ -1100,6 +1100,24 @@
...
+                            "application/json": {
+                                "schema": {
+                                    "type": "string"
+                                }
+                            }
+                        }
+                    }
+                },
+                "operationId": "configuration",
+                "description": "Retrieves the Kafka Bridge instance version in JSON format."
```
ppatierno, line 1119: it's not just version anymore. For now, it is, but in general, it could contain more.

src/main/resources/openapi.json
```diff
@@ -1100,6 +1100,24 @@
...
+        },
+        "/": {
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "Info about Kafka Bridge instance configuration.",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "type": "string"
```
ppatierno, line 1112: it should be now a JSON with more information not just a string. For now, it's a JSON with just bridge version field. Can we define a new type definition? Furthermore, we should have an example.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tombentley: `CONFIGURATION` etc are not `static final`, so why do they have UPPER_CASE names?

ppatierno: it was in this way since the beginning, maybe for readability. We can change it but not in this PR.

ppatierno:
@laidan6000 can you take a look at this from a documentation point of view please?

documentation/book/api/paths.adoc
```diff
@@ -2,6 +2,27 @@
 [[_paths]]
 == Paths
 
+[[_info]]
+=== GET /
+
+==== Description
+Retrieves the Kafka Bridge instance info in JSON format.
```
laidan6000, line 9:
```suggestion
Retrieves information about the Kafka Bridge instance, in JSON format.
```

documentation/book/api/paths.adoc
```diff
@@ -2,6 +2,27 @@
...
+
+
+==== Responses
+
+[options="header", cols=".^2a,.^14a,.^4a"]
+|===
+|HTTP Code|Description|Schema
+|**200**|Info about Kafka Bridge instance.|string
```
laidan6000, line 17: Info > Information

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -476,6 +478,15 @@ private void openapi(RoutingContext routingContext) {
         });
     }
 
+    private void version(RoutingContext routingContext) {
```
ppatierno, line 481: it should be `info` not only `version` anymore.

ppatierno:
@scholzj @tombentley anything more to add?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
stanlyDoge: Oh... I messed up a rebase somehow :(

PR #426, opened 2020-06-15
Added JMX metrics collection for exposing in Prometheus format

ppatierno:
This PR adds a way for getting JMX metrics and exposing them in the Prometheus format through the HTTP server `/metrics` endpoint.
The metrics can be "filtered" providing a JMX configuration file follow the syntax provided by the JMX exporter.

Signed-off-by: Paolo Patierno <ppatierno@live.com>

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -26,10 +26,17 @@ import io.vertx.micrometer.MetricsDomain;
 import io.vertx.micrometer.MicrometerMetricsOptions;
 import io.vertx.micrometer.VertxPrometheusOptions;
 import io.vertx.micrometer.backends.BackendRegistries;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.DefaultParser;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.management.MalformedObjectNameException;
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
@@ -38,6 +45,7 @@ import java.util.Map;
 /**
  * Apache Kafka bridge main application class
  */
+@SuppressWarnings("checkstyle:ClassDataAbstractionCoupling")
 public class Application {
 
     private static final Logger log = LoggerFactory.getLogger(Application.class);
@@ -53,7 +61,7 @@ public class Application {
         VertxOptions vertxOptions = new VertxOptions().setMetricsOptions(
                 new MicrometerMetricsOptions()
                         .setPrometheusOptions(new VertxPrometheusOptions().setEnabled(true))
-                        // define the lables on the HTTP server related metrics
+                        // define the labels on the HTTP server related metrics
                         .setLabels(EnumSet.of(Label.REMOTE, Label.LOCAL, Label.HTTP_PATH, Label.HTTP_METHOD, Label.HTTP_CODE))
                         // disable metrics about pool and verticles
                         .setDisabledMetricsCategories(EnumSet.of(MetricsDomain.NAMED_POOLS, MetricsDomain.VERTICLES))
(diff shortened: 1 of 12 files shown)


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -203,19 +217,53 @@ public static void main(String[] args) {
...
+     *
+     * @param commandLine command line parameters containing the path to the JMX YAML configuration
+     * @return  JmxCollectorRegistry instanc
+     * @throws MalformedObjectNameException
+     * @throws IOException
+     */
+    private static JmxCollectorRegistry getJmxCollectorRegistry(CommandLine commandLine)
+            throws MalformedObjectNameException, IOException {
+        if (commandLine.hasOption("jmx-metrics-config")) {
+            return new JmxCollectorRegistry(new File(getFilePath(commandLine.getOptionValue("jmx-metrics-config"))));
```
scholzj, line 232: Does the file automatically reload like with the exporter?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -527,7 +531,12 @@ private void errorHandler(RoutingContext routingContext) {
 
     private void metricsHandler(RoutingContext routingContext) {
         PrometheusMeterRegistry prometheusMeterRegistry = (PrometheusMeterRegistry) meterRegistry;
-        routingContext.response().setStatusCode(HttpResponseStatus.OK.code()).end(prometheusMeterRegistry.scrape());
+
+        StringBuilder sb = new StringBuilder();
+        sb.append(jmxCollectorRegistry.scrape());
+        sb.append(prometheusMeterRegistry.scrape());
+
+        routingContext.response().setStatusCode(HttpResponseStatus.OK.code()).end(sb.toString());
```
scholzj, line 539: I wonder if this can be somehow put into one place instead of having it twice - here and in the EmbeddedHttpServer.

src/main/java/io/strimzi/kafka/bridge/JmxCollectorRegistry.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * @return Content that should be included in the response body for an endpoint designated for
+     * Prometheus to scrape from.
+     */
+    public String scrape() {
+        Writer writer = new StringWriter();
+        try {
+            TextFormat.write004(writer, collectorRegistry.metricFamilySamples());
```
scholzj, line 55: Out of curiosity ... does the method name (`write004`) have any hidden meaning?

src/test/java/io/strimzi/kafka/bridge/http/HttpCorsTests.java
```diff
@@ -61,6 +62,16 @@
     static KafkaFacade kafkaCluster = new KafkaFacade();
     static MeterRegistry meterRegistry = BackendRegistries.getDefaultNow();
 
+    static JmxCollectorRegistry jmxCollectorRegistry;
+
+    static {
+        try {
+            jmxCollectorRegistry = new JmxCollectorRegistry("");
+        } catch (Exception ex) {
```
samuel-hawker, line 70: So if we throw an exception here, it doesn't matter?

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -203,19 +217,53 @@ public static void main(String[] args) {
...
+                .desc("JMX Kafka metrics YAML configuration file")
+                .build();
+
+        Options options = new Options();
+        options.addOption(configFileOption);
+        options.addOption(jmxMetricsFileOption);
+        return options;
+    }
+
+    private static String getFilePath(String arg) {
```
samuel-hawker, line 265: If it's returning an absolute file path, why not call this function `absoluteFilePath`

I know Tom Bentley in the past has mentioned we shouldn't use `get` in a method name if we aren't actually retreiving a field value in a javabean fashion, so we can trim this down to `absoluteFilePath(String path)` and the intent and purpose of the function is clear I think

src/main/java/io/strimzi/kafka/bridge/Application.java
ppatierno: Yes, the underneath `JmxCollector` class does that. Every time you scrape for metrics, it checks if the configuration file is changed (checking the last modified time) and if needed, it reloads the configuration itself.

src/main/java/io/strimzi/kafka/bridge/JmxCollectorRegistry.java
ppatierno: I did what actually the agent does and it generates the output in this "exposition format" (version 0.0.4). It's the format you see when asking for metrics, i.e. the `# HELP ..... ` and `#TYPE ...` lines before each metric, then the metric name followed by the value and so on.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Agree I did a little bit of refactoring adding a common `MetricsReporter` class to be used by both.

src/test/java/io/strimzi/kafka/bridge/http/HttpCorsTests.java
ppatierno: After my refactoring using a `MetricsReporter` class, we can just set to null the registries instances.

ppatierno:
@scholzj @samuel-hawker any more comments on this?

src/main/java/io/strimzi/kafka/bridge/amqp/AmqpBridge.java
```diff
@@ -91,17 +91,17 @@
...
 
-    private MeterRegistry meterRegistry;
+    private MetricsReporter metricsReporter;
 
     /**
      * Constructor
      *
      * @param bridgeConfig bridge configuration
-     * @param meterRegistry registry for scraping metrics
+     * @@param metricsReporter MetricsReporter instance for scraping metrics from different registries
```
stanlyDoge, line 100: Is the second `@` a typo?

PR #429, opened 2020-06-16
Used Swagger2Markup 2.0.0-SNAPSHOT to fix path doc examples

### LINKED ISSUE #420: Wrong JSON for response example
Every response value example in the `paths.adoc` is wrong because it shows a JSON where the field is `application/vnd.kafka.v2+json` having as a value the actual response.
For example, on consumer creation:

```json
{
  "application/vnd.kafka.v2+json" : {
    "instance_id" : "consumer1",
    "base_uri" : "http://localhost:8080/consumers/my-group/instances/consumer1"
  }
}
```

while the actual response value is just

```json
{
   "instance_id" : "consumer1",   
   "base_uri" : "http://localhost:8080/consumers/my-group/instances/consumer1"
}
```

ppatierno:
This PR fixes #420.
It's using the Swagger2Markdown 2.0.0-SNAPSHOT for now, while waiting for the final release (which seems not having an ETA) but at least we can have API documentation fixed.

### DIFF
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -101,10 +101,8 @@ __required__|Name and configuration of the consumer. The name is unique within t
 [source,json]
 ----
 {
-  "application/vnd.kafka.v2+json" : {
-    "instance_id" : "consumer1",
-    "base_uri" : "http://localhost:8080/consumers/my-group/instances/consumer1"
-  }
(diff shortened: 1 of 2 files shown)


ppatierno:
That's a fair point I haven't thought. I will close this, but anyway we could live with the bugged doc for long time :-(

PR #431, opened 2020-06-17
Added JMX metrics configuration example file

ppatierno:
This PR adds a YAML file containing a configuration example for JMX exposed Kafka clients metrics useful for a Grafana dashboard.


ppatierno:
You are absolutely right, I will move it during the interval ;-)

ppatierno:
I moved the JMX config file but even removed the `application.properties` from the resources because it's already in the config folder as well.

scholzj:
> I moved the JMX config file but even removed the application.properties from the resources because it's already in the config folder as well.

Assuming it is not used for tests or anything I think that makes sense.

Issue #432, opened 2020-06-18
basePath value of the openapi spec at /openapi should reflect the actual path


elakito:
Currently, the openapi 2.0 spec returned at /openapi is statically set to `/`. When the Kafka-Bridge instance is behind a gateway and the gateway is taking requests at path e.g., `/app/kafka-bridge` and reverse-proxing to the bridge instance at path `/`, it would be more convenient if the returned openapi spec has `basePath` being set to `/app/kafka-bridge` instead of `/`, when the `x-forwarded-prefix` header of the request is available to derive the actual request path.

In this case, the client can,  for example,  issue a topics request at `/app/kafka-bridge/topics` instead of at `/topics`, knowing the correct `basePath` for this particular bridge instance.

This ticket is resulted from this preceding discussion
https://github.com/strimzi/strimzi-kafka-bridge/issues/375#issuecomment-643159596

ppatierno:
@elakito I have just noticed that `basePath` is something only supported by OpenAPI 2.0.
With OpenAPI 3.0, the new `servers` section is used.
Our plan is about getting rid of OpenAPI 2.0 specification soon, providing the 3.0 specification on the `/openapi` endpoint so I was wondering how much it makes really sense to do the change in the https://github.com/strimzi/strimzi-kafka-bridge/pull/433

elakito:
@ppatierno if we are getting rid of openapiv2 any time soon, we won't need the v2 specific change.  If we switch to return openapi(v3) at /openapi, we could either use a similar approach to overwrite the server's url between '/' or '/prefix-path' or simply use a variable notation  '{basePath}' for that property and set its default value to '/' so that the consumer can decide on its value. For the latter case, we would just need a change in the spec file.

PR #433, opened 2020-06-19
/openapi endpoint with forwarded header

sknot-rh:

If I understood https://github.com/strimzi/strimzi-kafka-bridge/issues/432 correctly, this should fix it.

If we call `/openapi` endpoint and the `x-Forwarded-Path` header is set, its value is projected into `basePath` in the result openapi spec.

### DIFF
--- a/documentation/book/api/overview.adoc
+++ b/documentation/book/api/overview.adoc
@@ -25,7 +25,6 @@ __Terms of service__ : http://swagger.io/terms/
 
 === URI scheme
 [%hardbreaks]
-__Host__ : bridge.swagger.io
 __BasePath__ : /
 __Schemes__ : HTTP
 
--- a/documentation/book/api/overview.adoc
+++ b/documentation/book/api/overview.adoc
@@ -25,6 +25,7 @@ __Terms of service__ : http://swagger.io/terms/
 
 === URI scheme
 [%hardbreaks]
+__Host__ : bridge.swagger.io
 __BasePath__ : /
 __Schemes__ : HTTP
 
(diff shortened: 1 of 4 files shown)


ppatierno:
@elakito it would be great to have your feedback on this

elakito:
Ok. It looks good but I have a few comments.

1. Regarding the header name, both x-forwarded-prefix and x-forwarded-path headers are used for this purpose and it seems x-forwarded-prefix is more widely used or mentioned (in google search, `"x-forwarded-prefix"` returns 6.920 hits whereas `"x-forwarded-path"` returns 793 hits). Therefore, it would be more convenient to look up both names so that it will work out-of-the-box in all environments.

2. for this path rewriting step, if this header is absent, would it be simpler to just return the original buffer than always doing json-parse-serialize to create another buffer?

3. another question regarding this path rewriting step, originally I thought one would use a templated original data with a template engine like mustache to set the variables. I think that would probably be more efficient than doing the json-parse-serialize approach but would create a new dependency to another lib or else would need some custom template variable replacing code. So, I am raising this point but no strong opinion about which approach is better.

4. for this rewriting to be useful, the host property should either be removed from openapiv2.json so that the client can simply assume the host to be the current target host or be replaced to the correct host using x-forwarded-host. I think simply removing this host property from openapi spec is a better choice than adding the host replacing code.

scholzj:
@sknot-rh Could you please have a look at the conflict? @ppatierno Could you please review this?

ppatierno (approved):

PR #435, opened 2020-06-23
Embedding JMX configuration and making metrics to be enabled/disabled

ppatierno:
This PR removes the option to specify a JMX YAML configuration for the Kafka part.
It will be "opinioned" and the configuration embedded into the jar.
The PR also allows to enable/disable metrics exposed to Prometheus.


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -51,27 +55,38 @@
     private static final Logger log = LoggerFactory.getLogger(Application.class);
 
     private static final String EMBEDDED_HTTP_SERVER_PORT = "EMBEDDED_HTTP_SERVER_PORT";
+    private static final String BRIDGE_METRICS_ENABLED = "BRIDGE_METRICS_ENABLED";
 
     private static final int DEFAULT_EMBEDDED_HTTP_SERVER_PORT = 8080;
 
+
+
```
scholzj, line 63: Do we need 3 empty lines? Isn't that a bit too much? Maybe two would be sufficient :-P

ppatierno: even just one!

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -51,27 +55,36 @@
...
+                // setup Micrometer metrics options
+                vertxOptions.setMetricsOptions(
+                        new MicrometerMetricsOptions()
+                                .setPrometheusOptions(new VertxPrometheusOptions().setEnabled(true))
+                                // define the labels on the HTTP server related metrics
+                                .setLabels(EnumSet.of(Label.REMOTE, Label.LOCAL, Label.HTTP_PATH, Label.HTTP_METHOD, Label.HTTP_CODE))
+                                // disable metrics about pool and verticles
+                                .setDisabledMetricsCategories(EnumSet.of(MetricsDomain.NAMED_POOLS, MetricsDomain.VERTICLES))
+                                .setJvmMetricsEnabled(true)
+                                .setEnabled(true));
```
tombentley, line 78: Looks like this should be factored into its own `metricsOptions()` method.

PR #440, opened 2020-06-24
Fixed not working logging level for bridge related stuff

ppatierno:
After moving to log4j2, the configuration file is not working for the `io.strimzi.kafka.bridge` anymore.
This PR fixes this.
This PR is also needed to fix https://github.com/strimzi/strimzi-kafka-operator/issues/3232


ppatierno:
The one in `config` folder is packed for the release. The other one in `resource` is used as internal resource for debugging purpose (even if you could just run your bridge passing the log4j2 properties file as command line parameter)

PR #444, opened 2020-06-28
Allow s390x build to fail but fast finish the overall builds

ppatierno:
In order to handle the Travis build for s390x, which is quite unreliable and fails due to infrastructure problem, and in order to avoid blocking development for that, I had a chat with @nirmannarang who suggested to use "allow failures" and "fast finish" Travis features so that even if s390x fails, the overall build is considered to be ok.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: xenial
 sudo: required
 language: java
-matrix:
+jobs:
   fast_finish: true
   allow_failures:
   - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ jobs:
     jdk: openjdk8
   - os: linux
     jdk: openjdk11
-  - bridge: s390xbuild
+  - build: s390x # custom job identifier to be referenced in the allow_failures section
     os: linux
     arch: s390x
     jdk: openjdk11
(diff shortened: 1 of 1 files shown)


ppatierno:
@nirmannarang I cannot add you as a reviewer but it would be great to have feedback from you as well.

ppatierno:
I am doing different tries but it seems that the s390x job isn't executed anymore :-(

.travis.yml
```diff
@@ -2,18 +2,22 @@ dist: xenial
...
 language: java
 jobs:
+  fast_finish: true
   include:
   - os: linux
-    dist: bionic
+    jdk: openjdk8
+  - os: linux
+    jdk: openjdk11
+  - bridge: s390xbuild
```
scholzj, line 11: Out of curiosity ... what does this `bridge: s390xbuild` mean?

ppatierno: You need an identifier for the allow failures section to specify which job is allowed to fail.

samuel-hawker: I wonder if this information would be useful as a comment?

Always nice when an identifier string is documented, avoids people googling and being confused in the future.

ppatierno: Yes good idea, I added a comment.

nirmannarang:
@ppatierno LGTM, Thanks

PR #448, opened 2020-07-08
Change runtime java to 11

stanlyDoge:

### DIFF
--- a/bin/docker/kafka_bridge_run.sh
+++ b/bin/docker/kafka_bridge_run.sh
@@ -55,5 +55,11 @@ function get_gc_opts {
 
 export JAVA_OPTS="${JAVA_OPTS} $(get_gc_opts)"
 
+# Deny illegal access option is supported only on Java 9 and higher
+JAVA_MAJOR_VERSION=$(java -version 2>&1 | sed -E -n 's/.* version "([0-9]*).*$/\1/p')
+if [ "$JAVA_MAJOR_VERSION" -ge "9" ] ; then
+  JAVA_OPTS="${JAVA_OPTS} --illegal-access=deny"
+fi
+
 # starting Kafka Bridge with final configuration
 exec /usr/bin/tini -s -w -e 143 -- "${MYPATH}"/../kafka_bridge_run.sh --config-file=/tmp/kafka-bridge.properties "$@"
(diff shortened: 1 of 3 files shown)


scholzj:
Should this go through similar changes as the Java 11 runtime work in operators? E.g. make the Java 11 Travis build the main one etc.?

bin/docker/kafka_bridge_run.sh
```diff
@@ -55,5 +55,11 @@ function get_gc_opts {
 
 export JAVA_OPTS="${JAVA_OPTS} $(get_gc_opts)"
 
+# Deny illegal access option is supported only on Java 9 and higher
+JAVA_MAJOR_VERSION=$(java -version 2>&1 | sed -E -n 's/.* version "([0-9]*).*$/\1/p')
+if [ "$JAVA_MAJOR_VERSION" -ge "9" ] ; then
+  JAVA_OPTS="${JAVA_OPTS} --illegal-access=deny"
+fi
```
ppatierno, line 62: I see this addition on the @scholzj PR for doing the same on the operators image as well. I can understand the purpose of it but was it strictly related to the transition to Java 11? @scholzj ?

scholzj: @tombentley requested it there. But I think it will not hurt here either. At least we will know the code is ready for the future. WDYT @tombentley?

ppatierno: I am sure that it doesn't hurt here and there. I was just wondering the rationale behind doing the change during the upgrade to Java 11, because the flag seems to work even with Java 9 as before.

scholzj: The flag does not exist in Java 8. So you have to surround it with the `if` if that was your point.

tombentley: It's not strictly related to Java 11, but previously we were using Java 8 where it wasn't supported (because no modules), so this is the earliest opportunity to add it. It's worth having because while we might have a clue about what our own code is doing we can never to so sure about dependencies.

ppatierno: And ... right ... we were on Java 8 not 9 ;-)

PR #450, opened 2020-07-09
Renamed HTTP and Kafka metrics

ppatierno:
Currently, the HTTP metrics start with `vertx_` and they could be confused with the ones coming from the operators as well.
Regarding the Kafka related metrics, they start with `kafka_producer_` and `kafka_consumer_` which don't describe that they are related to consumer/producer in the bridge.
This PR changes the prefix for both, using `strimzi_bridge_`.


ppatierno:
When merged, this could be cherry-picked to `release-0.17.x` branch in order to have a 0.17.1 release including this change.

PR #455, opened 2020-07-20
add getOffsets operation using adminConsumer

elakito:
Signed-off-by: Aki Yoshida <elakito@gmail.com>

I would like to add the getOffsets operation to retrieve the offsets pair (i.e., first and last) of a specific partition. This needs a consumer instance that acts more like the admin consumer than a normal consumer. I created this consumer instance using a dedicated consumer ID and have it managed within the sinked endpoints with the other consumers. There may be several alternatives in how to manage this consumer instance and I am not sure if the approach taken is the best one.

Could you look at this PR and give me a feedback?
Thanks.

aki

### DIFF
--- a/documentation/book/api/definitions.adoc
+++ b/documentation/book/api/definitions.adoc
@@ -143,6 +143,19 @@ __optional__|< <<_offsetrecordsent,OffsetRecordSent>> > array
 |===
 
 
+[[_offsetssummary]]
+=== OffsetsSummary
+
+[options="header", cols=".^3a,.^4a"]
+|===
+|Name|Schema
+|**beginning_offset** +
+__optional__|integer (int64)
+|**end_offset** +
+__optional__|integer (int64)
+|===
+
+
 [[_partition]]
 === Partition
 
--- a/documentation/book/api/paths.adoc
+++ b/documentation/book/api/paths.adoc
@@ -1464,4 +1464,69 @@ __required__|Name of the topic to send records to or retrieve metadata from.|str
 ----
 
 
+[[_getoffsets]]
+=== GET /topics/{topicname}/partitions/{partitionid}/offsets
+
+==== Description
+Retrieves a summary of the offsets for the topic partition.
+
+
+==== Parameters
+
+[options="header", cols=".^2a,.^3a,.^9a,.^4a"]
+|===
+|Type|Name|Description|Schema
+|**Path**|**partitionid** +
+__required__|ID of the partition.|integer
+|**Path**|**topicname** +
+__required__|Name of the topic containing the partition.|string
+|===
+
+
+==== Responses
+
+[options="header", cols=".^2a,.^14a,.^4a"]
+|===
+|HTTP Code|Description|Schema
+|**200**|a summary of the offsets|<<_offsetssummary,OffsetsSummary>>
+|**404**|The specified topic partition was not found.|<<_error,Error>>
+|===
+
+
+==== Produces
+
+* `application/vnd.kafka.v2+json`
+
+
+==== Tags
+
+* Topics
+
+
+==== Example HTTP response
+
+===== Response 200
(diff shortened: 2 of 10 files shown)


ppatierno:
Are you sure having a consumer instance is the only available solution? I would always try to check that something is available in the Admin Client API. For this use case I found this:

https://kafka.apache.org/25/javadoc/org/apache/kafka/clients/admin/Admin.html#listOffsets-java.util.Map-

as stated in the documentation it says "List offset for the specified partitions and OffsetSpec. This operation enables to find the beginning offset, end offset as well as the offset matching a timestamp in partitions."
It seems to be pretty much what you are trying to achieve, or?

elakito:
@ppatierno thanks for the info. I didn't see this method. I'll look into it.

elakito:
@ppatierno The listOffsets in AdminClient seems came with https://issues.apache.org/jira/browse/KAFKA-7689 in kafka-2.5.0. What should we do? We wait for 2.5.0 to support this operation in Kafka-bridge?

ppatierno:
@elakito so the first step would be update to Kafka 2.5.0 in the Vert.x Kafka client and exposing these methods. Then waiting for a new Vert.x 3.9.x release to use in the bridge and then exposing the feature. I know it's a long way ...

elakito:
@ppatierno vertx-kafka-client's master(4.0.0-snapshot) still uses 2.4.0 and I don't know if vertx 3.9.x is ever allowed to switch to 2.5.0.  And not to mention, 2.5.0 is relatively new. As confluent's http gateway supports `GET /topics/{topicname}/partitions/{partitionid}/offsets`, I would like to have it also supported in strimzi's kafka-bridge soon.

ppatierno:
@elakito we can move to Kafka 2.5.0 for both, master and 3.9.x. I will try to find some time to do the upgrade. It's not trivial because I need to rebuild the vert.x stack project to check that there it no dependency to be broken. Other Vert.x projects have Kafka dependency.

elakito:
@ppatierno OK. Then we wait for the 2.5 API.

ppatierno:
@elakito I updated the Vert.x Kafka client to use Kafka 2.6 but now we have to wait for the Vert.x release.

elakito:
@ppatierno thanks for the update. I'll look at it.

ppatierno:
@elakito FYI, now the master is updated to Vert.x 3.9.3 with Kafka 2.6.0 so I guess you can update this PR as we discussed, right?

elakito:
@ppatierno thanks!  I'll update the PR...

elakito:
@ppatierno I need to expose listOffsets from vertx-kafka-client KafkaAdminClient so that strimizi can use it. I created this PR
https://github.com/vert-x3/vertx-kafka-client/pull/186 in master.
Could you review it? Thanks.

elakito:
@ppatierno thanks for merging the vertx PR. I prepared for the change using vertx-kafka-client-3.9.4-SNAPSHOT with https://github.com/elakito/strimzi-kafka-bridge/commit/ab2c469a9ab0d83ff49497678fbf98871b83c606

Should I wait for vertx-kafka-client-3.9.4 to be released and then create a PR from it or should I replace this PR with a new one using 3.9.4-SNAPSHOT and park it here until 3.9.4 is released?

ppatierno:
Well, you can anticipate the work on the PR using the 3.9.4-SNAPSHOT but then we have to wait for the final 3.9.4 release to get the PR merged. It's just anticipating the work, but if you want you can wait for 3.9.4 ... as you prefer.

elakito:
I saw 3.9.4 is released so I'll prepare for a PR...

elakito:
@ppatierno will you update pom.xml to upgrade vertx to 3.9.4?

elakito:
Replaced by the above PR.

PR #456, opened 2020-07-23
[MO] - [tests] -> fix

see-quick:
This PR fixing that tests did not execute on master's branch. 

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -204,7 +204,7 @@
 			<groupId>io.strimzi</groupId>
 			<artifactId>test-container</artifactId>
 			<version>${project.version}</version>
-		</dependency>
+		</dependency> 
 	</dependencies>
 
 	<pluginRepositories>
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
 		<junit.platform.version>1.5.2</junit.platform.version>
 		<hamcrest.version>2.2</hamcrest.version>
 		<maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
-		<maven-surefire-plugin.version>2.22.1</maven-surefire-plugin.version>
+		<maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
 		<swagger2markup.version>1.3.7</swagger2markup.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -202,29 +200,6 @@
 			<type>test-jar</type>
 			<scope>test</scope>
 		</dependency>
-		<dependency>
```
ppatierno, line 205: why all these dependencies are not needed anymore?

pom.xml
```diff
@@ -19,16 +19,14 @@
 		<junit.platform.version>1.5.2</junit.platform.version>
 		<hamcrest.version>2.2</hamcrest.version>
 		<maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
-		<maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
+		<maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
```
ppatierno, line 22: you mentioned you were using the "milestone" version in the operator project as well, are we sure that using this "old" version doesn't break anything?

pom.xml
see-quick: i have tried execute tests with `3.0.0-M5` but there is some missing dependency which is needed.  In the next PR i will investigate what is the problem with that version but for now we can let this lower version (which is stable)

pom.xml
see-quick: I have tried with `3.0.0-M5` but there is some dependency, which is missing. Anyway, in the next PR I can investigate the problem and change it to this `3.0.0-M5` but for now, we can let it on stable `2.22.2`.

pom.xml
see-quick: These dependencies were needed for that `M5` version

see-quick: I have tried with 3.0.0-M5 but there is some dependency, which is missing. Anyway, in the next PR I can investigate the problem and change it to this 3.0.0-M5 but for now, we can let it on stable 2.22.2.

pom.xml
see-quick: These dependencies were needed for that M5 version

PR #457, opened 2020-07-23
[MO] - [preparing for pipeline] ->  #1 external kafka + bridge

see-quick:
This PR adding the ability to run the tests with external Kafka and bridge. 


ppatierno:
Can you explain me better about this PR please? It seems to mix different things. Why some tests are disabled if the external bridge env var is true?

see-quick:
Ok, this PR preparing the whole test-suite to run against external Kafka and external Bridge. Moreover, Disabling some tests because I run the tests which all the same context (bridge config). For instance `ConsumerGeneratedNameTest` is ignored because it needs a different config of Bridge.

see-quick:
The points it that it is really hard to pass tests consistently because we have 3 types of env. 

 - In-memory bridge + Kafka = always passing
 - External bridge + Kafka = some fails, which has to be disabled I have tried to fix them (already fix like 5 of them) but if I modify the test to pass on external bridge + Kafka it will fail on in-memory which is really strange.
 -  Travis CI (in-memory bridge + Kafka) =  passing

ppatierno:
Would it be better to understand why some tests fail in a scenario and not the other? At the end, it's always bridge + Kafka and maybe the fact they are failing means something worst underneath?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -465,7 +466,7 @@ void createConsumerWithNotExistingParameter(VertxTestContext context) throws Int
 
     @Test
     void receiveSimpleMessage(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
-        String topic = "receiveSimpleMessage";
+        String topic = "receiveSimpleMessage-1";
```
ppatierno, line 469: why changing the name?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -711,7 +712,7 @@ void receiveFromTopicsWithPattern(VertxTestContext context) throws InterruptedEx
 
     @Test
     void receiveSimpleMessageFromPartition(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
-        String topic = "receiveSimpleMessageFromPartition";
+        String topic = "receiveSimpleMessageFromPartition-1";
```
ppatierno, line 715: ditto as above

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1576,6 +1555,17 @@ void createConsumerNameIsNotSetAndBridgeIdIsSet(VertxTestContext context) throws
...
+    @BeforeEach
+    void setUp() {
+        name = generateRandomConsumerName();
+        consumerWithEarliestReset.put("name", name);
+        consumerJson.put("name", name);
+    }
 
+    private String generateRandomConsumerName() {
+        int salt = new Random().nextInt(Integer.MAX_VALUE);
+        return "my-kafka-consumer" + salt + salt;
```
ppatierno, line 1569: why adding `salt` two times?

see-quick: two times less probability that new consumer-name will be the same :D

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
see-quick: I have changed it to default most likely old code... )

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -406,7 +408,7 @@ void createConsumerWithForwardedHeaderDefaultPort(VertxTestContext context) thro
     }
 
     @Test
-    void createConsumerWithForwardedHeaderWrongProto(VertxTestContext context) throws InterruptedException {
+    void createConsumerWithForwardedHeaderWrongProto(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
```
ppatierno, line 411: which part of the code raises these new two exceptions?

see-quick: old code sorry :(

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
ppatierno: tbh I don't think so ... you are generating the random salt just one time but concatenating twice. The fact that you would have less probability is related to the generation of the number not to the concatenation. If you generate salt = 23 for example you will just have "my-kafka-consumer2323" but the probability is the same as you just generate salt = 23 and have "my-kafka-consumer23". For having less probability you should generate two different salts and concatenating them.

see-quick:
Make sense, you are god Paolo!

src/test/java/io/strimzi/kafka/bridge/http/ProducerTest.java
```diff
@@ -191,7 +191,7 @@ void sendSimpleMessageWithKey(VertxTestContext context) throws InterruptedExcept
     }
 
     @Disabled("Will be check in the next PR, this is just external tests for Bridge")
-    @DisabledIfEnvironmentVariable(named = "BRIDGE_EXTERNAL_ENV", matches = "((?i)FALSE(?-i))")
+    @DisabledIfEnvironmentVariable(named = "EXTERNAL_BRIDGE", matches = "((?i)TRUE(?-i))")
```
ppatierno, line 194: wasn't disabling test supposed to be in a different PR?

see-quick: I have changed the logic and naming the env nothing more.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1576,6 +1555,17 @@ void createConsumerNameIsNotSetAndBridgeIdIsSet(VertxTestContext context) throws
         consumerService()
             .deleteConsumer(context, groupId, consumerInstanceId[0]);
         context.completeNow();
+    }
+
+    @BeforeEach
+    void setUp() {
+        name = generateRandomConsumerName();
+        consumerWithEarliestReset.put("name", name);
```
ppatierno, line 1563: `consumerWithEarliestResetJson` to highlight that it's JSON as you did for `consumerJson`

PR #458, opened 2020-07-27
[MO] - [preparing for pipeline] -> #2 Disabling tests for external

see-quick:
This PR is based on the https://github.com/strimzi/strimzi-kafka-bridge/pull/457 so after it will be merged I will do a rebase (it will remove the duplicate code). Moreover adding disabled tests, which does not work with external Kafka + bridge.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -412,7 +412,7 @@ public class ConsumerTest extends HttpBridgeTestBase {
     }
 
     @Test
-    void createConsumerWithForwardedHeaderWrongProto(VertxTestContext context) throws InterruptedException {
+    void createConsumerWithForwardedHeaderWrongProto(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
         // this test emulates a create consumer request coming from an API gateway/proxy
         String forwarded = "host=my-api-gateway-host;proto=mqtt";
 
@@ -1510,6 +1510,7 @@ public class ConsumerTest extends HttpBridgeTestBase {
         context.completeNow();
     }
 
+    @DisabledIfEnvironmentVariable(named = "EXTERNAL_BRIDGE", matches = "((?i)TRUE(?-i))")
     @Test
     void consumerDeletedAfterInactivity(VertxTestContext context) {
         CompletableFuture<Boolean> create = new CompletableFuture<>();
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
@@ -1510,7 +1510,7 @@ public class ConsumerTest extends HttpBridgeTestBase {
         context.completeNow();
     }
 
-    @DisabledIfEnvironmentVariable(named = "EXTERNAL_BRIDGE", matches = "((?i)TRUE(?-i))")
+    @DisabledIfEnvironmentVariable(named = "EXTERNAL_BRIDGE", matches = "((?i)TRUE(?-i))") 
     @Test
     void consumerDeletedAfterInactivity(VertxTestContext context) {
         CompletableFuture<Boolean> create = new CompletableFuture<>();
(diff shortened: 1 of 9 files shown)


im-konge:
@see-quick @stanlyDoge can you please take a look at this so it can be merged? This should fix several tests from the suite.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -1636,6 +1638,6 @@ void setUp() {
 
     private String generateRandomConsumerName() {
         int salt = new Random().nextInt(Integer.MAX_VALUE);
-        return "my-kafka-consumer" + salt;
+        return "my-kafka-consumer" + salt + salt;
```
ppatierno, line 1641: I had a discussion with @see-quick about this. Adding salt two times is useless.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -470,7 +471,7 @@ void createConsumerWithNotExistingParameter(VertxTestContext context) throws Int
 
     @Test
     void receiveSimpleMessage(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
-        String topic = "receiveSimpleMessage";
+        String topic = "receiveSimpleMessage-1";
```
ppatierno, line 474: why adding this "-1" to topic names (I see it's done for others) fixes problems?

ppatierno:
I a couple of comments but I there are no information why we are disabling some tests for external bridge or why they are failing. Can you provide more insights, please?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
see-quick: It should work without these suffixes but AFAIK when I deploy external Kafka and external Bridge it fixes problems otherwise I got an error with a topic already exists.

see-quick:
Tomorrow, I will take a closer look at these test cases.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
ppatierno: So it means that there is another problem somewhere, maybe not deleting topics at the end of the tests?

see-quick: Could be the issue, that `AdminClient` does not delete them properly.

see-quick:
@ppatierno So, the point of Disabling the tests cases for external `Kafka` and `KafkaBridge` for instance `ConsumerGeneratedNameTest` is because this test suite needs to change the of `KafkaBridge`.

stanlyDoge:
It seems `batchHandler` is buggy (https://github.com/vert-x3/vertx-kafka-client/issues/170). I tried to rewrite it to the sequential handling and the test (`sendPeriodicMessage`) is working now. Patch - https://pastebin.com/JBds8490

see-quick:
@stanlyDoge @ppatierno everything resolved. You can review it again... :)

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
```diff
@@ -782,7 +781,7 @@ void receiveFromTopicsWithPattern(VertxTestContext context) throws InterruptedEx
 
     @Test
     void receiveSimpleMessageFromPartition(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
-        String topic = "receiveSimpleMessageFromPartition";
+        String topic = "receiveSimpleMessageFromPartition-1";
```
ppatierno, line 784: still "-1" ? :-D

src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeTestBase.java
```diff
@@ -154,19 +157,11 @@ static void afterAll(VertxTestContext context) {
         }
     }
 
-    @AfterEach
-    void tearDown() throws ExecutionException, InterruptedException {
+    @BeforeEach
+    void tearDown() throws ExecutionException, InterruptedException, TimeoutException {
+        topic = "my-topic-" + new Random().nextInt(Integer.MAX_VALUE);
```
ppatierno, line 162: didn't we discuss on this? isn't it better having a meaningful name for each topic to understand more about a test failure?

see-quick: Hmmm, yes we do. I would rather use this generated topic on the topic. It is the same approach as we do in `strimzi-kafka-operator` repo.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerTest.java
see-quick: Sorry :D Removed ))

ppatierno (approved):

stanlyDoge (approved):

PR #469, opened 2020-09-17
Bumped to Vert.x 3.9.3

ppatierno:
Bumped to Vert.x 3.9.3 and removed Scala dependency already brought by Kafka dependency.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
@@ -14,6 +14,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.ext.web.codec.BodyCodec;
 import io.vertx.junit5.VertxTestContext;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.concurrent.CompletableFuture;
@@ -26,12 +27,13 @@ import static org.hamcrest.Matchers.is;
 
 public class ConsumerSubscriptionTest extends HttpBridgeTestBase {
 
+    String groupId = "my-group";
+
     @Test
     void unsubscribeConsumerNotFound(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
         adminClientFacade.createTopic(topic);
 
         String name = "my-kafka-consumer";
-        String groupId = "my-group";
 
         JsonArray topics = new JsonArray();
         topics.add(topic);
@@ -63,7 +65,6 @@ public class ConsumerSubscriptionTest extends HttpBridgeTestBase {
     @Test
     void subscribeExclusiveTopicAndPattern(VertxTestContext context) throws Throwable {
         String name = "my-kafka-consumer-exclusive";
-        String groupId = "my-group";
 
         JsonObject json = new JsonObject();
         json.put("name", name);
@@ -130,8 +131,6 @@ public class ConsumerSubscriptionTest extends HttpBridgeTestBase {
         String name = "my-kafka-consumer-does-not-exists-because-not-created";
         adminClientFacade.createTopic(topic);
 
-        String groupId = "my-group";
-
         JsonArray topics = new JsonArray();
         topics.add(topic);
 
@@ -161,7 +160,6 @@ public class ConsumerSubscriptionTest extends HttpBridgeTestBase {
         String topic = "subscriptionConsumerDoesNotExistBecauseAnotherGroup";
         adminClientFacade.createTopic(topic, 1, 1);
         String name = "my-kafka-consumer-does-not-exists-because-another-group";
-        String groupId = "my-group";
         String anotherGroupId = "anotherGroupId";
 
         JsonArray topics = new JsonArray();
@@ -211,7 +209,6 @@ public class ConsumerSubscriptionTest extends HttpBridgeTestBase {
         assertThat(adminClientFacade.listTopic().size(), is(2));
 
         String name = "my-kafka-consumer-list";
-        String groupId = "my-group";
 
         JsonObject json = new JsonObject();
         json.put("name", name);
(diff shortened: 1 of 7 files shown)


src/test/java/io/strimzi/kafka/bridge/http/HttpCorsTests.java
```diff
@@ -111,7 +111,7 @@ public void testCorsNotEnabled(VertxTestContext context) {
                         client.request(HttpMethod.POST, 8080, "localhost", "/consumers/1/instances/1/subscription")
                                 .putHeader("Origin", "https://evil.io")
                                 .send(ar2 -> context.verify(() -> {
-                                    assertThat(ar2.result().statusCode(), is(405));
+                                    assertThat(ar2.result().statusCode(), is(400));
```
stanlyDoge, line 114: Is this because of an empty subscription body?

ppatierno: It seems that with Vert.x 3.9.3 they changed the status code as Bad Request.

stanlyDoge (approved):
LGTM

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -244,7 +266,7 @@ void listConsumerSubscriptions(VertxTestContext context) throws InterruptedExcep
                 .send(ar -> {
                     context.verify(() -> {
                         assertThat(ar.succeeded(), is(true));
-
+                        listSubscriptions.complete(true);
```
ppatierno, line 269: you are just skipping all the assertions here I guess because the `listSubscriptions.get` is waiting and when it's completed there is no time for executing the assertions and the code after the get is executed and the test exits.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -226,16 +241,23 @@ void listConsumerSubscriptions(VertxTestContext context) throws InterruptedExcep
...
-        CompletableFuture<Boolean> consume = new CompletableFuture<>();
+                .subscribeConsumer(context, groupId, name, topic, topic2);
 
         // poll to subscribe
+        CountDownLatch latch = new CountDownLatch(2);
         consumerService()
                 .consumeRecordsRequest(groupId, name, BridgeContentType.KAFKA_JSON_JSON)
-                .as(BodyCodec.jsonObject())
-                .send(ar -> consume.complete(true));
+                .as(BodyCodec.jsonArray())
```
ppatierno, line 250: what's the array we are sending? consuming doesn't have any body

stanlyDoge: Good question. Leaving `jsonObject` causes `Future{cause=Invalid Json Object decoded as io.vertx.core.json.JsonArray}`

ppatierno: Ok ok, it's not about body sent, my bad. It's about decoding the response which is actually a JSON array.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -226,16 +241,23 @@ void listConsumerSubscriptions(VertxTestContext context) throws InterruptedExcep
...
+                .send(ar -> {
+                    latch.countDown();
+                });
+        consumerService()
+                .consumeRecordsRequest(groupId, name, BridgeContentType.KAFKA_JSON_JSON)
+                .as(BodyCodec.jsonArray())
+                .send(ar -> {
+                    latch.countDown();
+                });
+        latch.await(TEST_TIMEOUT, TimeUnit.SECONDS);
```
ppatierno, line 260: can you try to avoid this double request for polling and the latch. I think we are just missing the `consume.get` and using the CompletableFuture in the right way.
See the `subscriptionConsumerDoesNotExistBecauseAnotherGroup` for example. There is the `subscribe` future that of course is waited with `subscribe.get` and completed in the lamba.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
stanlyDoge: Yeah, I believe it is assembled in `io.strimzi.kafka.bridge.http.converter.HttpJsonMessageConverter#toMessages`

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -275,6 +253,7 @@ void listConsumerSubscriptions(VertxTestContext context) throws InterruptedExcep
                         assertThat(response.body().getJsonArray("partitions").size(), is(2));
                         assertThat(response.body().getJsonArray("partitions").getJsonObject(0).getJsonArray(topic2).size(), is(4));
                         assertThat(response.body().getJsonArray("partitions").getJsonObject(1).getJsonArray(topic).size(), is(1));
+                        listSubscriptions.complete(true);
```
ppatierno: try outside the `context.verify` and remove the `context.completeNow` it should be at the end.
Again, try to use the same pattern as `subscriptionConsumerDoesNotExistBecauseAnotherGroup` test.

src/test/java/io/strimzi/kafka/bridge/http/services/ConsumerService.java
```diff
@@ -33,6 +33,8 @@
 
     static final Logger LOGGER = LogManager.getLogger(ConsumerService.class);
 
+    private static final int POLL_TIMEOUT = 4000;
```
stanlyDoge, line 36: I guess you get this value from your investigation. Any idea why it is so high?

ppatierno: Tbh I just set a reasonable value higher than 1000ms. Anyway, it's not there to wait for 4 secs, the poll returns even before when rebalance is done and records are available. This is a max value, not a time to wait in each test.

stanlyDoge: I see. It is passing so far, well done!

PR #470, opened 2020-09-17
Modify versions to fix vulnerabilities

vbusch:
- Added dependencies for snakeyaml, commons-io, keycloak.  There were transitive dependencies before to older versions.
- bumped micrometer version
- changed jmx exporter version to match productized versions.

* Keycloak versions still need to be tested.

Updated scope of spotbugs-annotations, which doesn't need to be part of the final jar.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,13 @@
 			<groupId>io.strimzi</groupId>
 			<artifactId>kafka-oauth-client</artifactId>
 			<version>${strimzi-oauth.version}</version>
+              <exclusions>
+                    <exclusion>
+                        <groupId>org.bouncycastle</groupId>
+                        <artifactId>bcprov-jdk15on</artifactId>
+                    </exclusion>
+                </exclusions>
+
 		</dependency>
 		<dependency>
 			<groupId>io.jaegertracing</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -205,6 +213,43 @@
...
 		</dependency>
+
+		<!-- Transitive dependency version overrides for Vulnerabilities: -->
+		<!-- overriding version commons-io for vertex 3.9.3 - Vulnerability -->
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.5</version>
+		</dependency>
+		<!-- overriding version of snakeyaml for prometheus.jmx.collector 0.12.0: Vulnerability: DOS -->
```
scholzj, line 224: You changed the JMC Collector version to 0.10 ... so is this still needed?

pom.xml
```diff
@@ -27,8 +27,8 @@
 		<jaeger.version>1.1.0</jaeger.version>
 		<opentracing.version>0.33.0</opentracing.version>
 		<opentracing-kafka-client.version>0.1.12</opentracing-kafka-client.version>
-		<micrometer.version>1.3.1</micrometer.version>
-		<jmx-prometheus-collector.version>0.12.0</jmx-prometheus-collector.version>
+		<micrometer.version>1.3.9</micrometer.version>
+		<jmx-prometheus-collector.version>0.10</jmx-prometheus-collector.version>
```
scholzj, line 31: Are we downgrading to older version here? Why? Or is that again some weird numbering issue in JMX Exporter?

scholzj (commented):
Thanks for the PR. I left some comments.

pom.xml
```diff
@@ -205,6 +213,43 @@
...
+			<artifactId>commons-io</artifactId>
+			<version>2.5</version>
+		</dependency>
+		<!-- overriding version of snakeyaml for prometheus.jmx.collector 0.12.0: Vulnerability: DOS -->
+		<dependency>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.26</version>
+		</dependency>
+		<!-- overriding version of keycloak for kafka-oauth-client 0.5.0 : Vulnerability: Remote Code Execution -->
```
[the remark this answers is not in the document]
vbusch: 0.6.0 also uses keycloak 10.0 which has the CVE.  We need version 11.0

pom.xml
vbusch: Yes, it is bringing in version 2.4 which has a Remote Code Execution.

pom.xml
vbusch: To let it match a productized version.

scholzj: I don't think we normally do it ... If it is just dependency change, we can also do it directly in productisation and keep the newer version here I guess?

pom.xml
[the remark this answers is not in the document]
scholzj: So would it be better to update this first in the OAuth and do new 0.6.1 release of OAuth and then just use the new OAuth here? We will have the same problem in all projects, so doing it that way we fix it once and do not have to do it several times.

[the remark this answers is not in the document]
vbusch: Looks like master on https://github.com/strimzi/strimzi-kafka-oauth is at version 0.6.0.  I'll look at updating it for a productized version.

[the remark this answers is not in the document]
scholzj: No, we will need to productize that ;-)

pom.xml
vbusch: reverted this change, and we can do it in the productized version.

pom.xml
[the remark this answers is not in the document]
vbusch: Removed the keycloak & bouncy castle provider change.  They can be done in the oauth.

scholzj (approved):
Thanks for the PR.

pom.xml
```diff
@@ -205,6 +206,20 @@
...
 			<version>${project.version}</version>
 		</dependency>
+
+		<!-- Transitive dependency version overrides for Vulnerabilities: -->
+		<!-- overriding version commons-io for vertex 3.9.3 - Vulnerability -->
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.5</version>
+		</dependency>
```
ppatierno, line 216: should we open an issue to Vert.x raising the vulnerability they have and tracking the issue here?

vbusch:
@scholzj can you merge this PR?  I don't have write access.

ppatierno:
@vbusch what about my comment?

vbusch:
> @vbusch what about my comment?

@ppatierno thank you.  My apologies I completely missed it.  I should have refreshed first.  I'll open the issue now, and add it here.

ppatierno (approved):

Issue #388, opened 2020-02-21
make java_package doesn't work for clean install


morgangalpin:
To reproduce:
`make clean java_package`

The unit tests run and fail on `StaticTests.bridgeVersionDisplayedInStartupTest()`.

This is because the test phase relies on an artifact of the `maven-assembly-plugin:...:single` which runs after the `test` phase of maven.

To get the build to work, I needed to run:
```
MVN_ARGS=-DskipTests make clean java_package
make java_package
```

This is not ideal, as it takes longer, and the bridge version test isn't testing the current code, but an artifact of a previous build.

ppatierno:
Yeah we are aware of that. @granzoto you should try to think about a different test for that.

PR #477, opened 2020-09-30
move statictests to integration-test phase

elakito:
Signed-off-by: Aki Yoshida <elakito@gmail.com>

Moving StaticTests out of test phase and into integration-test phase.
As a result, mvn clean install will execute all tests except StaticTests and create the package jar and then execute StaticTests.
This should fix https://github.com/strimzi/strimzi-kafka-bridge/issues/388.

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -20,8 +20,8 @@ fi
 echo "Run docu check ..."
 make docu_check
 
-echo "Packaging ..."
-make java_package
+echo "Verifying ..."
+make java_verify
 
 echo "PULL_REQUEST=$PULL_REQUEST"
 echo "TAG=$TAG"
--- a/pom.xml
+++ b/pom.xml
@@ -275,6 +275,28 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>${maven-surefire-plugin.version}</version>
+				<configuration>
+					<excludes>
+						<exclude>**/*StaticTests.java</exclude>
+					</excludes>
+				</configuration>
+				<executions>
+					<execution>
+						<id>integration-test</id>
+						<goals>
+							<goal>test</goal>
+						</goals>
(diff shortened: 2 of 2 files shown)


ppatierno:
But in this way, it's never executed on Travis CI and we want it.

elakito:
@ppatierno is the travis test configured at `.travis/build.sh`? This seems to invoke `make java_package` which in turn invokes `mvn package `. In that case, we could instead invoke `mvn verify` to run also the integration-test during the travis test.

If the version info needs to be read from the packaged jar, the test has to wait for the packaging phase.  Or are you thinking of the version info to be read from a resource file (i.e., put release.version under src/main/resources)? In that case, we can read it before the packaging as well.

ppatierno:
@Frawless what do you think about it. I know that @see-quick was thinking about refactoring this test or doing it in a different way.

Frawless:
I am ok with this change. Guess the best option will be to have some test package as we have in operator repo and move all tests which needs jar instead of bridge in memory to there, but we have only one test atm. @see-quick  do you remember which changes you wanted to do?

see-quick:
I lost the idea that I had about the approach we should use but this change seems completely appropriate to me. One thing to consider is that we are going to have two builds in our CI:

1. In-memory Kafka with in-memory Kafka Bridge
2. Standalone Kafka with Standalonee Kafka Bridge

Moreover, on TravisCI we will have to execute `mvn verify` to invoke all tests with Static one?

ppatierno:
> Moreover, on TravisCI we will have to execute mvn verify to invoke all tests with Static one?

@see-quick this is what @elakito was suggesting.
@elakito can you make this change?

elakito:
@ppatierno OK. I'll make the travis change.

elakito:
I have made the change and also verified the travis job successfully executed StaticTests.

ppatierno (approved):
LGTM. Thanks!

Frawless (approved):

PR #479, opened 2020-10-04
Completed rate-limiting and content-based routing extensions

BorisRado:
Hi, this PR provides the bridge with rate limiting capabilities as well as with the possibility to dynamically determine the topic to which a bunch of messages should be directed to. 
You can read the extended description about how to manage these two functionalities in the `ADDED_README.md` file.
Implementational details about how did I implement the functionalities are discussed [in chapter 6 of my dissertation thesis, while in chapter 7](https://www.dropbox.com/s/5tnmdzd4eyr67i9/63170401_dopolnjena_diplomska.pdf?dl=0) it is presented a purely demonstrative use case for the two extensions.
Please let me know what do you think.
Cheers!

### DIFF
--- a/ADDED_README.md
+++ b/ADDED_README.md
@@ -1,0 +1,212 @@
+
+# 1 CONTENT-BASED ROUTING
+
+### 1.1 Purpose and overview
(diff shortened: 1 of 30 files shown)


ppatierno:
@BorisRado first of all thank you very much for this work! I haven't dug into the PR yet but I was thinking that having two different PRs (for rate limiting and custom routing) would be better in order to focus the review on each one. It could even speed up the process even understanding what kind of dependencies are important for one or the other.
Do you think it's reasonable?

BorisRado:
@ppatierno yes, I totally agree! In the coming days, I shall present two separate PR for rate limiting and content based routing.

PR #484, opened 2020-10-18
Try with resources is used which guarantees that the resource in question will be closed

kamaci:


ppatierno (approved):
LGTM. Thanks for the PR!

stanlyDoge (approved):

Issue #489, opened 2020-11-06
Assignment is broken


sknot-rh:
I wrote a test which is validating `assignments` endpoint. It is not returning 409 if the subscription is used together with assign endpoint. Also, there is an NPE.

PR #490, opened 2020-11-09
Assignment test + NPE fix

sknot-rh:

Assigment endpoint does not have any tests. 
Trying to access `subscribe` and `assign` endpoints ends up with NPE. After fix of NPE, it does not return 409 (Conflict).

Fixes https://github.com/strimzi/strimzi-kafka-bridge/issues/489

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
@@ -198,7 +198,7 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
     }
 
     /**
-     * Unubscribe all the topics which the consumer currently subscribes
+     * Unsubscribe all the topics which the consumer currently subscribes
      */
     protected void unsubscribe() {
         log.info("Unsubscribe from topics {}", this.topicSubscriptions);
@@ -360,6 +360,7 @@ public abstract class SinkBridgeEndpoint<K, V> implements BridgeEndpoint {
             Optional<PartitionInfo> requestedPartitionInfo =
                     availablePartitions.stream()
                             .filter(p -> p.getTopic().equals(topicSubscription.getTopic()) &&
+                                    topicSubscription.getPartition() != null &&
                                     p.getPartition() == topicSubscription.getPartition())
                             .findFirst();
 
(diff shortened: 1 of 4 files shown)


im-konge (approved):
LGTM, maybe there is something that @ppatierno will see :)

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -357,6 +357,7 @@ private void partitionsAssignmentAndSeek() {
             Optional<PartitionInfo> requestedPartitionInfo =
                     availablePartitions.stream()
                             .filter(p -> p.getTopic().equals(topicSubscription.getTopic()) &&
+                                    topicSubscription.getPartition() != null &&
```
ppatierno, line 360: can you explain in which case this can be `null`?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -356,6 +356,13 @@ private void doAssign(RoutingContext routingContext, JsonObject bodyAsJson) {
         this.setAssignHandler(assignResult -> {
             if (assignResult.succeeded()) {
                 HttpUtils.sendResponse(routingContext, HttpResponseStatus.NO_CONTENT.code(), null, null);
+            } else {
+                HttpBridgeError error = new HttpBridgeError(
```
ppatierno, line 360: why just returning CONFLICT? Couldn't it depend on the cause of error and why the assignment went wrong?
What kind of errors did you find?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionTest.java
```diff
@@ -264,6 +264,60 @@ void listConsumerSubscriptions(VertxTestContext context) throws InterruptedExcep
         context.completeNow();
     }
 
+    @Test
+    void assignTest(VertxTestContext context) throws InterruptedException, ExecutionException, TimeoutException {
+        String topicToSub = "subscribe-topic";
```
ppatierno, line 269: let's reflect the "assign" in the topic name

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
sknot-rh: If we do subscribe to a topic, it is stored into `topicSubscriptions`. Subscription request body might be for example 
```
{
  "topics": [
    "topic1",
    "topic2"
  ]
}
```
If we do assign after subscription (what is wrong because of exclusivity), we iterate through all topic subscriptions in `topicSubscriptions` and we are trying to get partitions. And because the subscription body does not have any partitions info, it is null. Then just comparing null causes NPE.

ppatierno (approved):

Issue #492, opened 2020-11-10
Poll without subscribe should return meaningfull body


sknot-rh:
Should return 
```
{
    "error_code": 500,
    "message": "Internal Server Error"
}
```
does return
```
Unexpected 'I'
```

It is probably trying to parse a JSON from String `Internal Server Error`.

PR #495, opened 2020-11-11
Poll without subscribe should return reasonable message

sknot-rh:

When the poll is requested, subscription list and the subsrciption patter is checked whether they are set up. If not, `Consumer is not subscribed to any topics or assigned any partitions` message is returned.

fixes https://github.com/strimzi/strimzi-kafka-bridge/issues/492

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -248,6 +248,16 @@ public class HttpSinkBridgeEndpoint<K, V> extends SinkBridgeEndpoint<K, V> {
     }
 
     private void doPoll(RoutingContext routingContext) {
+        if (!topicSubscriptionsPatternUsed && topicSubscriptions.isEmpty()) {
+            HttpBridgeError error = new HttpBridgeError(
+                    HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
+                    "Consumer is not subscribed to any topics or assigned any partitions"
+            );
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -248,6 +248,16 @@ private void doDeleteConsumer(RoutingContext routingContext) {
     }
 
     private void doPoll(RoutingContext routingContext) {
+        if (!topicSubscriptionsPatternUsed && topicSubscriptions.isEmpty()) {
```
ppatierno, line 251: not sure why we need this flag and it's not enough checking that the list is empty.
Subscribing or assigning is just filling some topics/partitions in the `topicSubscriptions` list.

sknot-rh: Look at https://github.com/strimzi/strimzi-kafka-bridge/blob/780dad5f5481b254a0b92ec6fe28e298be8ac30c/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java#L403

If we use `topics`, we do write topics into the `topicSubscriptions`. If we use `topic_pattern` we do not use `topicSubscriptions`.

ppatierno: right! So maybe it could make sense to save the pattern instead of having a boolean. Maybe the pattern used by the HTTP client could return useful if in the future we want to provide a way for the user to retrieve it. Wdyt?

ppatierno (approved):

scholzj:
@sknot-rh Why do you need another review when you have already approval from @ppatierno? Anything special to look for?

sknot-rh:
Force of habit of having 2 approvals O:)

Issue #502, opened 2021-03-02
 Add ppc64le support for travis


Abhijit-Mane:
ppc64le is IBM Power architecture.
We will submit a PR soon to enable strimzi-kafka-bridge for ppc64le as well.

im-konge:
Done in #507

PR #503, opened 2021-03-02
Update .travis.yml to add ppc64le support

Abhijit-Mane:
Added changes to support ppc64le arch.
Fixes issue #502

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,15 @@ jobs:
     jdk: openjdk8
   - os: linux
     jdk: openjdk11
+  - build: ppc64le
+    os: linux
+    arch: ppc64le
+    jdk: openjdk11
+    script: mvn install -DskipTests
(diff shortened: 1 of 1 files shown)


scholzj:
Can you please explain what is the expected value of this? It does not really produce anything, it does not even run the tests.

krishvoor:
@scholzj - we are working towards enabling AMQ Streams on ppc64le arch. and 
kafka-bridge is one of the dependencies, while the build is successful on ppc64le, the unit-tests are incomplete,
as they are dependent on some of the docker-images (that can be built via kafka-operator), we will submit a patch soon to
enable for ppc64le arch in kafka-operator.

This PR is to get it enabled/ build on ppc64le arch., once the docker-images of kafka-operator for ppc64le are in place,
we will enable the test cases for ppc64le.

PR #507, opened 2021-04-09
Add support for ppc64le arch

Abhijit-Mane:
Add changes to support ppc64le arch.
Fixes issue #502 
Signed-off-by: Abhijit Mane <abhijman@in.ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic
 sudo: required
 language: java
 jobs:
@@ -8,6 +8,15 @@ jobs:
     jdk: openjdk8
   - os: linux
     jdk: openjdk11
+  - build: ppc64le
+    os: linux
+    arch: ppc64le
+    jdk: openjdk11
+    script: mvn install -DskipTests
+    addons:
+      apt:
+        packages:
+        - maven
   - build: s390x
     os: linux
     arch: s390x
@@ -19,6 +28,7 @@ jobs:
         - maven
   allow_failures:
   - build: s390x
+  - build: ppc64le
 services:
 - docker
 before_install:


.travis.yml
```diff
@@ -8,6 +8,15 @@ jobs:
     jdk: openjdk8
   - os: linux
     jdk: openjdk11
+  - build: ppc64le
+    os: linux
+    arch: ppc64le
+    jdk: openjdk11
+    script: mvn install -DskipTests
```
Frawless, line 15: Tests are failing on ppc or why are skipped?

krishvoor: The dependent docker-images ( for these tests ) are supplied via strimzi-kafka-operator, we are working with community to get them enabled/ published for ppc64le architecture, currently as a work around, we are skipping the tests.

scholzj: I think that ideally the tests needing the other Docker Images should be marked as Integration Tests and we should use just `skipITs`. The Unit Tests should not need anything and should be able to pass even without it. 

If the ITs are not marked as ITs today, can you please open a separate issue for it?

scholzj (approved):
Looks good to me apart from the comment from @Frawless ... we should either have an issue about naming the tests properly or we should just skip the ITs.

.travis.yml
Frawless: I am afraid to even the unit tests use that image for Kafka. TBH I am not sure how to easily solve it. We can refactor the code and remove usage of it or build Kafka image for ppc...not sure which option is worse :/

scholzj: Well, as I said ... that is wrong. Unit tests should not depend on any other components such as Kafka. So IMHO anything depending on a Kafka container is IT and should be marked as such.

Frawless: Makes sense. @krishvoor issue for that would be nice :) thanks

Frawless (approved):

.travis.yml
scholzj: Ok, I opened https://github.com/strimzi/strimzi-kafka-bridge/issues/508 for this.

scholzj:
@ppatierno @sknot-rh Do you have any more comments on this? Or can it be merged?

sknot-rh (approved):

Issue #508, opened 2021-04-13
Mark integration tests as Integration tests and not Unit tests


scholzj:
It looks like Bridge has a lot of integration tests testing it directly with Kafka. But these seem to be marked as unit tests. This should be cleaned up and the tests should be named properly in order to make sure we can execute those separately (for example as `-DskipTests` or `-DskipITs`).

echernous:
The PR got merged, closing

PR #512, opened 2021-04-29
[Issue 508] Renamed IT tests

echernous:
Signed-off-by: Kate Chernousova <eachernous@gmail.com>

The change addresses https://github.com/strimzi/strimzi-kafka-bridge/issues/508, renamed some of tests as ITs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,5 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>io.strimzi</groupId>
 	<artifactId>kafka-bridge</artifactId>
@@ -19,7 +19,7 @@
 		<hamcrest.version>2.2</hamcrest.version>
 		<maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
 		<maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
-                <maven-failsafe-plugin.version>3.0.0-M5</maven-failsafe-plugin.version>
+		<maven-failsafe-plugin.version>3.0.0-M5</maven-failsafe-plugin.version>
 		<swagger2markup.version>1.3.7</swagger2markup.version>
 		<jackson-core.version>2.10.2</jackson-core.version>
 		<spotbugs.version>4.0.1</spotbugs.version>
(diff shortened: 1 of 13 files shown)


scholzj (commented):
Are the IT tests being executed? I think you migth need to add the failsafe plugin to the `pom.xml` to run them.

Frawless:
Yes, it looks that tests weren't executed. You can reuse the way how we handle ITs in operator repo with failsafe plugin.

echernous:
Hi @scholzj and @Frawless, thanks for checking! Added the failsafe plugin now

pom.xml
```diff
@@ -272,25 +276,53 @@
...
+                                        <execution>
+                                                <goals>
+                                                        <goal>integration-test</goal>
+                                                        <goal>verify</goal>
+                                                </goals>
+                                                <configuration>
+                                                        <includes>
+                                                                <include>**/IT*.java</include>
+                                                                <include>**/*IT.java</include>
+                                                                <include>**/*StaticTest.java</include>
```
Frawless, line 297: Why this one is not named as the others ITs?

Frawless (approved):
LGTM, I can see ITs executed in travis build for jdk8 and jdk11.

pom.xml
echernous: It should be `StaticTests` but I missed "s" in the end. I assume this class was created to put some static tests but currently there is only one test, should it be named as BridgeStartupIT or just StaticIT?

echernous: renamed it to StaticIT to be consistent

pom.xml
```diff
@@ -19,6 +19,7 @@
 		<hamcrest.version>2.2</hamcrest.version>
 		<maven-jar-plugin.version>3.1.2</maven-jar-plugin.version>
 		<maven-surefire-plugin.version>3.0.0-M5</maven-surefire-plugin.version>
+                <maven-failsafe-plugin.version>3.0.0-M5</maven-failsafe-plugin.version>
```
scholzj, line 22: Should this have the same alignment as the other lines? Or is it misaligned for some reason only in the diff?

pom.xml
```diff
@@ -30,6 +31,9 @@
 		<jmx-prometheus-collector.version>0.12.0</jmx-prometheus-collector.version>
 		<commons-cli.version>1.4</commons-cli.version>
 		<test-container.version>0.20.0</test-container.version>
+                <!-- property to skip surefire tests during failsafe execution -->
+                <!--suppress UnresolvedMavenProperty -->
+                <skip.surefire.tests>${skipTests}</skip.surefire.tests>
```
scholzj, line 36: Same as above.

scholzj (commented):
The code looks good to me. But the pom.xml seems to have a weird alignment. Not sure it is just the GitHub diff or also in reality. Could you please check it?

sknot-rh (approved):

pom.xml
echernous: yeah, something went wrong there, fixed now, thanks!

scholzj (approved):

PR #514, opened 2021-05-20
Enable DOCKER_BUILDKIT

krishvoor:
This PR will allow users to seamlessly build containers
on all the architectures without additionally passing build-args.

Signed-off-by: Krishna Harsha Voora <krishvoor@in.ibm.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,11 +23,20 @@ COPY target/kafka-bridge-${strimzi_kafka_bridge_version}/kafka-bridge-${strimzi_
 #####
 # Add Tini
 #####
-ENV TINI_VERSION v0.18.0
-ENV TINI_SHA256=12d20136605531b09a2c2dac02ccee85e1b874eb322ef6baf7561cd93f93c855
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /usr/bin/tini
-RUN echo "${TINI_SHA256} */usr/bin/tini" | sha256sum -c \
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Why not do this via `docker buildx` the same way as it is done in the operators repository?

krishvoor:
Hi Jakub,
Apologies for the delayed response.
> Why not do this via docker buildx the same way as it is done in the operator's repository?

Yes, for any one-off build on non-x86_64 architecture, this approach would work.

Currently, if we follow the instructions and try to build a container image i.e. via `make docker_build` 
the build fails on non-x86_64 architectures, hence these updates.

Please let me know, in case there are other workarounds.

scholzj:
> Currently, if we follow the instructions and try to build a container image i.e. via make docker_build
the build fails on non-x86_64 architectures, hence these updates.

Sorry, I'm not sure I follow. It does not fail for the operators.

krishvoor:
You are right, as you pointed out earlier, operators are handling this via `$DOCKER_BUILDX` in the Makefile
https://github.com/strimzi/strimzi-kafka-operator/blob/main/Makefile.docker#L21 
And this passes.

However, this is failing on strimzi-kafka-bridge, given the lack of it.

krishvoor:
Further to the above-given context, the problem is arising while downloading the architecture-specific tini binary..
If we do not introduce the above changes it results in picking up the wrong architecture-specific tini version.

krishvoor:
Moving the conversation to #521

PR #517, opened 2021-05-24
Use UBI8 as a new base image

scholzj:
This PR updates the base image used by Strimzi Bridge to Red Hat UBI8 as proposed by [SP-23](https://github.com/strimzi/proposals/blob/main/023-using-ubi8-as-base-image.md). It also adds support for ARM64 version of the `tini` utility.


sknot-rh (approved):
Are you looking at the failed tests?

ppatierno (approved):

scholzj:
@sknot-rh That seem to be some flaky tests

PR #522, opened 2021-06-01
Fix Vert.x cache directory

scholzj:
#504 updated Vert.x to version 4.0.3. But due to some changes, Vert.x 4 requires different configuration for the cache directory to work with Strimzi. This is not found int he tests because it does not show up when running outside of containers. This Pr fixes the cache path and makes the container image work.


ppatierno (approved):

sknot-rh (approved):
Good catch. Thank you

PR #535, opened 2021-08-03
Add fields to ProducerRecord and ProducerRecordToPartition in docs

LiamClarkeNZ:
Spotted some fields not in the documentation. 

Signed-off-by: Liam Clarke <liam@steelsky.co.nz>

### DIFF
--- a/documentation/book/api/definitions.adoc
+++ b/documentation/book/api/definitions.adoc
@@ -226,6 +226,10 @@ __optional__|< <<_partition,Partition>> > array
 __optional__|<<_kafkaheaderlist,KafkaHeaderList>>
 |**partition** +
 __optional__|integer (int32)
+|**key** +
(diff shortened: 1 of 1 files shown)


LiamClarkeNZ:
Was looking at the docs when assisting a colleague, saw some fields absent from ProducerRecord / ProducerRecordToPartition in the definitions, so added them.

scholzj:
The file you are changing is [generated in the Maven build](https://github.com/strimzi/strimzi-kafka-bridge/blob/main/pom.xml#L475-L494) from the OpenAPI spec. So you cannot just change it there. You would need to figure out why are these fields not generated in it

LiamClarkeNZ:
Ah, thanks @scholzj I did wonder, I'll dig further, and create a new PR.

PR #542, opened 2021-09-29
[MO] - completely removing Debezium Kafka broker dependency

see-quick:
This PR removing Debezium Kafka cluster dependency and replace it with our `strimzi-test-container`. Moreover, I have updated version from `0.23` to `0.25`.

Signed-off-by: morsak <xorsak02@stud.fit.vutbr.cz>

PR #543, opened 2021-09-30
Overrride the dependencies with CVE

ShubhamRwt:

This PR Overrides the dependencies which contains CVE.


ShubhamRwt:
We can override the Kafka dependency once PR #542  is merged

pom.xml
```diff
@@ -255,7 +255,24 @@
...
+			<groupId>io.netty</groupId>
+			<artifactId>netty-handler</artifactId>
+			<version>4.1.68.Final</version>
+		</dependency>
+		<dependency>
+			<!-- transitive dep; override here to avoid version with CVE-->
+			<groupId>org.apache.tomcat.embed</groupId>
+			<artifactId>tomcat-embed-core</artifactId>
+			<version>8.5.68</version>
+		</dependency>
```
scholzj, line 275: Please add to the comments what is it a transitive dependency of.

scholzj: You should also use a variable for the versions

tomncooper (approved):

pom.xml
```diff
@@ -99,6 +99,9 @@
 		<sonatype.nexus.staging>1.6.3</sonatype.nexus.staging>
 		<swagger2markup.version>1.3.7</swagger2markup.version>
 		<jackson-core.version>2.11.3</jackson-core.version>
+		<netty-codec.version>4.1.68.Final</netty-codec.version>
+		<netty-handler.version>4.1.68.Final</netty-handler.version>
```
scholzj, line 103: Do we need two options? Isn't that just `netty.version`?

ShubhamRwt:
@scholzj I have bumped the Kafka dependency to 3.0.0 now

scholzj (approved):
LGTM assuming tests pass.

sknot-rh (approved):

sknot-rh:
It seems that getting offsets from unknown topic has changed behavior in 2.8.x. Thus the `getOffsetsSummaryNotFoundTest` is failing. I am looking into that.

ShubhamRwt:
> It seems that getting offsets from unknown topic has changed behavior in 2.8.x. Thus the `getOffsetsSummaryNotFoundTest` is failing. I am looking into that.

Thankyou @sknot-rh

pom.xml
```diff
@@ -83,7 +83,7 @@
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<log4j.version>2.13.3</log4j.version>
 		<vertx.version>4.1.2</vertx.version>
-		<kafka.version>2.7.0</kafka.version>
+		<kafka.version>2.8.0</kafka.version>
```
scholzj, line 86: So, why can't we use 2.8.1 here? I understnd that 3.0.0 has issues with VErt.x. But 2.8.1 should be fine here, or?

ShubhamRwt: Ahh, I changed it from 2.8.1 to 2.8.0 to check if the test failure were present in this version also. I will bump the version now

tomncooper: 2.8.1 would be better as it has CVE fixes.

ShubhamRwt: I have bumped it to 2.8.1 now

Issue #253, opened 2019-06-14
Create a test case for crashing Kafka producer creation


ppatierno:
Currently we have a test in order to check the crash on Kafka consumer creation.
We should add one for the crash on the Kafka producer creation.
I wrote that one but then I deleted it. It's not so simple to test because actually, the only way to inject a crash on producer creation is to provide wrong producer configuration in the HTTP bridge properties file but the bridge is started before the test method running and we cannot change the configuration on fly.

PR #546, opened 2021-10-05
Invalid producer test

sknot-rh:

Create a test where we configure (kafka) producer incorrectly. We should get an error response from the bridge when it receives a request.

Fixes https://github.com/strimzi/strimzi-kafka-bridge/issues/253

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/InvalidProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/InvalidProducerIT.java
@@ -73,17 +73,13 @@ public class InvalidProducerIT extends HttpBridgeITAbstract {
 
         LOGGER.info("Environment variable EXTERNAL_BRIDGE:" + BRIDGE_EXTERNAL_ENV);
 
-        if ("FALSE".equals(BRIDGE_EXTERNAL_ENV)) {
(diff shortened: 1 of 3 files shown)


src/test/java/io/strimzi/kafka/bridge/http/InvalidProducerIT.java
```diff
@@ -0,0 +1,93 @@
...
+
+        if ("FALSE".equals(BRIDGE_EXTERNAL_ENV)) {
+            bridgeConfig = BridgeConfig.fromMap(tmpCfg);
+            httpBridge = new HttpBridge(bridgeConfig, new MetricsReporter(jmxCollectorRegistry, meterRegistry));
+            httpBridge.setHealthChecker(new HealthChecker());
+
+            LOGGER.info("Deploying in-memory bridge");
+            vertx.deployVerticle(httpBridge, context.succeeding(id -> context.completeNow()));
+        } else {
+            context.completeNow();
```
Frawless, line 84: Is this correct? Wouldn't that pass the test even in case of failure?

sknot-rh: Yea, that is a good catch. This test can run as non-external only. Thanks

PR #555, opened 2021-11-01
Add support for multi-arch image manifests and for AArch64 / ARM64 container builds

scholzj:
This PR adds support for multi-arch container builds and experimental support for the AArch64 / ARM64 platform.

Most of the logic is done inside the Azure Pipelines CI and in the Make build system. When building locally, the default build should work exactly as it worked up until now. It builds a regular image based on your platform and tags it as `:latest` by default. This way, there is minimal disruption for local development and the build is also faster without the multi-arch build. If you would want to build for another architecture and your Docker supports it, you can use the environment variable `DOCKER_ARCHITECTURE` ... for example `DOCKER_ARCHITECTURE=arm64 make all`.

But on the CI, it does the builds for both AMD64 and ARM64. These images are Since we use Java, these images use always the same java binaries (TAR.GZ with JARs), but differ in platform. They are pushed into the container registry with the platform as part of the tag - e.g. `:latest-amd64` and `:latest-arm64`. Afterwards, the build creates the multi-arch manifest which is tagged as `:latest` and pushed into the registry as well. That way, the image name will be still `quay.io/strimzi/kafka-bridge:latest` as before (technically, this would be the multi-arch manifest) and each platform will pull the right image based on it.

You can check how it will look in Quay on the example build I did and on this screenshot:
Screenshot 2021-11-01 at 23 21 30

### DIFF
--- a/.azure/build-pipeline.yaml
+++ b/.azure/build-pipeline.yaml
@@ -33,6 +33,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
+          architectures: ['amd64', 'arm64']
   - stage: container_publish
     displayName: Publish Container
     dependsOn:
@@ -48,6 +49,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
+          architectures: ['amd64', 'arm64']
   - stage: docs_publish
     displayName: Publish Docs
     dependsOn:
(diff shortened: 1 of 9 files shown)


scholzj:
This should be now ready for review.

ppatierno (approved):
It sounds good to me in relation of my understanding of Azure Pipelines ;-)

Frawless (approved):

PR #556, opened 2021-11-07
Add OpenTelemetry support.

alesj:
Signed-off-by: Ales Justin <ales.justin@gmail.com>


config/application.properties
```diff
@@ -1,7 +1,8 @@
 #Bridge related settings
 bridge.id=my-bridge
 # uncomment the following line to enable Jaeger tracing, check the documentation how to configure the tracer
-#bridge.tracing=jaeger
+bridge.tracing=jaeger-otel
+bridge.tracing.service-name=strimzi-kafka-bridge
```
ppatierno, line 5: Shouldn't the above be commented to not have tracing enabled by default? At least it's how works today with the comment above explaining to uncomment to enable tracing.

pom.xml
```diff
@@ -106,6 +106,9 @@
 		<jaeger.version>1.6.0</jaeger.version>
 		<opentracing.version>0.33.0</opentracing.version>
 		<opentracing-kafka-client.version>0.1.15</opentracing-kafka-client.version>
+		<opentelemetry.version>1.7.0-alpha</opentelemetry.version>
+		<opentelemetry-stable.version>1.7.0</opentelemetry-stable.version>
```
ppatierno, line 110: why there are two opentelemetry version properties (one with `-stable` and the other seems to be an alpha) using both in different places to import dependencies?

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -62,12 +61,13 @@
     private static final int DEFAULT_EMBEDDED_HTTP_SERVER_PORT = 8080;
 
     @SuppressWarnings({"checkstyle:NPathComplexity"})
+    @SuppressFBWarnings("REC_CATCH_EXCEPTION")
```
ppatierno, line 64: why this one?

src/main/java/io/strimzi/kafka/bridge/MetricsReporter.java
```diff
@@ -24,6 +25,7 @@
      * @param jmxCollectorRegistry JmxCollectorRegistry instance for scraping metrics from JMX endpoints
      * @param meterRegistry MeterRegistry instance for scraping metrics exposed through Vert.x
      */
+    @SuppressFBWarnings("SIC_INNER_SHOULD_BE_STATIC_ANON")
```
ppatierno, line 28: why this one?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,207 @@
...
+        return serviceName;
+    }
+
+    @Override
+    public void initialize() {
+        OpenTelemetrySdkAutoConfiguration.initialize();
+    }
+
+    private static Tracer get() {
+        return GlobalOpenTelemetry.getTracer(COMPONENT); // TODO OK?
```
ppatierno, line 77: what's this "TODO OK?" ?

config/application.properties
alesj: Ah, yes, forgot to change it back ... fixed.

pom.xml
alesj: "My" OpenTelemetry artifacts did yet made it to "stable" version, where Jaeger artifcat alfready did.
Hence the 2 diff versions.
But we agreed it's OK to use alpha for those -- as they are the initial ones, and haven't been changed.
I'll check if there is a stable version, but shouldn't make a diff.
(if I change it again now, I need to re-run the whole cycle, for the N-th time ...)

src/main/java/io/strimzi/kafka/bridge/Application.java
alesj: When I ran `mvn clean install -DskipTests spotbugs:check` it spit out this warning / error ...

src/main/java/io/strimzi/kafka/bridge/MetricsReporter.java
alesj: Same here, checkstyle / spotbugs complained ...

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: Just an early remark for me, removing it ...

src/main/java/io/strimzi/kafka/bridge/Application.java
ppatierno: this is not the reason, this is the detection of the problem ... I would like to know why there is now a problem that needs this suppress warning that wasn't there before.

src/main/java/io/strimzi/kafka/bridge/MetricsReporter.java
ppatierno: ditto as above

src/main/java/io/strimzi/kafka/bridge/Application.java
alesj: No idea, I was just following spotbugs error report.
Perhaps checkstyle plugin upgrade?

src/main/java/io/strimzi/kafka/bridge/MetricsReporter.java
alesj: Same ... as you can see, I didn't change anything else in this class.

src/main/java/io/strimzi/kafka/bridge/Application.java
ppatierno: I would try to understand what's the error and then at least opening an issue with the reason to track and doing a change later to avoid doing it in the current PR. If spotbugs is raising this maybe we don't see a potential bug ... it's worth investigating.

alesj: I think it's the issue of "eating" the exception:
* https://github.com/archived-codacy/codacy-findbugs/blob/master/src/main/resources/docs/description/REC_CATCH_EXCEPTION.md

Which in our case that's not the (completely the) case -- we log and exit the app ...

pom.xml
ppatierno: Did you check if there is a stable version? This "alpha" here bothers me a little bit :-)

src/main/java/io/strimzi/kafka/bridge/Application.java
ppatierno: The link you pasted seems to suggest that we should not catch just `Exception` as the main does but `RuntimeException` or being more specific. Would it be the solution instead of suppress the warning?

alesj: Imho, we can leave it as is ... as that's from main() and we log and exit.

pom.xml
alesj: Let me check ...

ppatierno (approved):

pom.xml
alesj: https://search.maven.org/artifact/io.opentelemetry.instrumentation/opentelemetry-kafka-clients-common

alesj: All alphas ... but that's OK, since, afaik / imho, the 1st didn't change at all ...

src/main/java/io/strimzi/kafka/bridge/tracing/TracingUtil.java
```diff
@@ -0,0 +1,107 @@
...
+    private static TracingHandle tracing = new NoopTracing();
+
+    public static TracingHandle getTracing() {
+        return tracing;
+    }
+
+    public static void initialize(BridgeConfig config) {
+        String tracingConfig = config.getTracing();
+        if (tracingConfig != null && tracingConfig.contains("jaeger")) {
+            boolean isOpenTelemetry = "jaeger-otel".equals(tracingConfig);
```
scholzj, line 31: Define as `static final String ...` ...

src/main/java/io/strimzi/kafka/bridge/tracing/TracingUtil.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final Logger log = LoggerFactory.getLogger(TracingUtil.class);
+    private static TracingHandle tracing = new NoopTracing();
+
+    public static TracingHandle getTracing() {
+        return tracing;
+    }
+
+    public static void initialize(BridgeConfig config) {
+        String tracingConfig = config.getTracing();
+        if (tracingConfig != null && tracingConfig.contains("jaeger")) {
```
scholzj, line 30: Would it be better to just say `...equals("jaeger") || ...equals("jarger-otel")`?

src/main/java/io/strimzi/kafka/bridge/tracing/TracingUtil.java
```diff
@@ -0,0 +1,107 @@
...
+            boolean isOpenTelemetry = "jaeger-otel".equals(tracingConfig);
+            TracingHandle instance = isOpenTelemetry ? new OpenTelemetryHandle() : new OpenTracingHandle();
+
+            String serviceName = instance.serviceName(config);
+            if (serviceName != null) {
+                log.info(
+                    "Initializing Jaeger ({}) tracingConfig with service name {}",
+                    isOpenTelemetry ? "OpenTelemetry" : "OpenTracing",
+                    serviceName
+                );
```
scholzj, line 40: Does this make sense? Doesn't `serviceName` ecist only for OTel? I.e. would this log line ever print for OpenTelemetry?

src/test/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryTest.java
```diff
@@ -0,0 +1,21 @@
...
+import io.vertx.tracing.opentelemetry.OpenTelemetryOptions;
+
+/**
+ * OpenTelemetry tests
+ */
+public class OpenTelemetryTest extends TracingTestBase {
+    @Override
+    protected TracingOptions tracingOptions() {
+        System.setProperty("otel.traces.exporter", "jaeger");
+        System.setProperty("otel.service.name", "strimzi-kafka-bridge-test");
```
scholzj, line 18: This is where you should use the constants defined elsewhere ;-)

scholzj (commented):
Did you test the backwards compatibility? I.e. that the old OpenTracing tracing works?

config/application.properties
```diff
@@ -1,7 +1,8 @@
 #Bridge related settings
 bridge.id=my-bridge
 # uncomment the following line to enable Jaeger tracing, check the documentation how to configure the tracer
-#bridge.tracing=jaeger
+#bridge.tracing=jaeger-otel
+#bridge.tracing.service-name=strimzi-kafka-bridge
```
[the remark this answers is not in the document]
alesj: OK, put back the old one as well + simple description.

pom.xml
```diff
@@ -106,6 +106,9 @@
...
+		<grpc.version>1.41.0</grpc.version>
```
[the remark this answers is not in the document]
alesj: Afair, gRPC is optional -- as it's just one of the transports that can be used (the most common one).
Hence it needs to be explicitly declared.

src/main/java/io/strimzi/kafka/bridge/Application.java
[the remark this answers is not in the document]
alesj: > If we do not know what and why are we catching, we should probably remove the whole try-catch instead of suppressing the warning.

And have main() throw Exception?

src/main/java/io/strimzi/kafka/bridge/converter/DefaultDeserializer.java
```diff
@@ -19,6 +20,7 @@ public void configure(Map<String, ?> configs, boolean isKey) {
     }
 
     @Override
+    @SuppressFBWarnings("REC_CATCH_EXCEPTION")
```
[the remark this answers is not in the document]
alesj: Same issue as within Application class ...

src/main/java/io/strimzi/kafka/bridge/tracing/TracingUtil.java
alesj: No, it exists for both -- this is the unique name for your traces / spans (e.g. when you check them in the Jaeger server UI).

src/main/java/io/strimzi/kafka/bridge/Application.java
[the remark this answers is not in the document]
scholzj: To have `main()` throw what exception? The error suggests there is no exception being thrown other than possible RuntimeException which can for sure be thrown out of `main()`.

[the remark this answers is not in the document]
alesj: Then the error is wrong, since there are 3 different non-runtime exceptions being thrown.
Hence probably why we had that all-catch in the first place.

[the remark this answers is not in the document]
scholzj: Then maybe what you had there was right, but you didn't explained it netiehr in the PR comments nor in the code comments.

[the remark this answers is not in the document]
alesj: I doubt I can explain it better then the error's original docs, from the link posted above, and this one:
* https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#rec-exception-is-caught-when-exception-is-not-thrown-rec-catch-exception

But I fail to see why this approach is better in our case (or for that matter in all cases ...).

src/main/java/io/strimzi/kafka/bridge/tracing/TracingConstants.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public interface TracingConstants {
+    String JAEGER = "jaeger";
+
+    String JAEGER_OPENTRACING = JAEGER;
+    String JAEGER_OPENTELEMETRY = JAEGER + "-otel";
+
+    String OPENTELEMETRY_SERVICE_NAME_ENV_KEY = "OTEL_SERVICE_NAME";
+    String OPENTELEMETRY_SERVICE_NAME_PROPERTY_KEY = "otel.service.name";
+    String OPENTELEMETRY_TRACES_EXPORTER_KEY = "otel.traces.exporter";
```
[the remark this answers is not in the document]
alesj: Moved to a class.

src/main/java/io/strimzi/kafka/bridge/Application.java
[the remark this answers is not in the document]
scholzj: I guess I need to look at the actual code.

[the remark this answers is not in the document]
scholzj: Ok, so, the code throws 3 different exceptions:
* `IOException`
* `MalformedObjectNameException`
* `ParseException`

So the error tells you to do either this:

```java
        } catch (MalformedObjectNameException|IOException|ParseException ex) {
            log.error("Error starting the bridge", ex);
            System.exit(1);
        }
```

eventually something like this if we want to catch the RuntimeException as well:

```java
        } catch (RuntimeException|MalformedObjectNameException|IOException|ParseException ex) {
            log.error("Error starting the bridge", ex);
            System.exit(1);
        }
```

So why not use that? This is what it suggests.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,212 @@
...
+    @Override
+    public String envName() {
+        return OPENTELEMETRY_SERVICE_NAME_ENV_KEY;
+    }
+
+    @Override
+    public String serviceName(BridgeConfig config) {
+        String serviceName = System.getenv(envName());
+        if (serviceName == null) {
+            serviceName = System.getProperty(OPENTELEMETRY_SERVICE_NAME_PROPERTY_KEY);
```
scholzj, line 61: What is the reason for getting it from the properties suddenly? Do we have 3 way s how to configure the service? In envvar., system propeerty and in the config file?

scholzj (commented):
The good news ... the OpenTracing still seems to work fine :-). But I didn't managed to get anything using OTel. When I try to follow the docs PR from @PaulRMellor, it always fails to connect to my Jaeger. Do you have some example configurations for Jaeger Operator and for the agent to play with?

config/application.properties
```diff
@@ -1,7 +1,11 @@
 #Bridge related settings
 bridge.id=my-bridge
-# uncomment the following line to enable Jaeger tracing, check the documentation how to configure the tracer
+# uncomment one the following lines (bridge.tracing) to enable Jaeger tracing, check the documentation how to configure the tracer
+# OpenTracing support
 #bridge.tracing=jaeger
+# OpenTelemetry support
+#bridge.tracing=jaeger-otel
+#bridge.tracing.service-name=strimzi-kafka-bridge
```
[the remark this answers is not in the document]
alesj: Dunno. @ppatierno ? ^

config/application.properties
```diff
@@ -1,7 +1,11 @@
 #Bridge related settings
 bridge.id=my-bridge
-# uncomment the following line to enable Jaeger tracing, check the documentation how to configure the tracer
+# uncomment one the following lines (bridge.tracing) to enable Jaeger tracing, check the documentation how to configure the tracer
+# OpenTracing support
 #bridge.tracing=jaeger
+# OpenTelemetry support
+#bridge.tracing=jaeger-otel
```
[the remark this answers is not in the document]
alesj: Not needed of course.
It was `jaeger` for some unknown reason -- as if there was ever any other option? -- so @ppatierno ? ^^
... and I just followed that unknown reason ...

pom.xml
```diff
@@ -203,6 +206,42 @@
...
+			<groupId>io.opentelemetry</groupId>
+			<artifactId>opentelemetry-exporter-jaeger</artifactId>
+			<version>${opentelemetry-stable.version}</version>
+		</dependency>
+		<!-- Use gRPC as the transport -->
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-netty-shaded</artifactId>
+			<version>${grpc.version}</version>
+		</dependency>
```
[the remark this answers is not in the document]
alesj: * https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#jaeger-exporter

`The [Jaeger](https://www.jaegertracing.io/docs/1.21/apis/#protobuf-via-grpc-stable) exporter. This exporter uses gRPC for its communications protocol.`

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: That's how OTel uses this -- env, sys -- but it can be left out.
@ppatierno ^

alesj:
> The good news ... the OpenTracing still seems to work fine :-). But I didn't managed to get anything using OTel. When I try to follow the docs PR from @PaulRMellor, it always fails to connect to my Jaeger. Do you have some example configurations for Jaeger Operator and for the agent to play with?

This is the env vars I have (plus enabled `jaeger-otel` tracing in config)

OTEL_TRACES_EXPORTER=jaeger;JAEGER_SERVICE_NAME=strimzi-kafka-bridge

and it works for me from IDE.

scholzj:
> This is the env vars I have (plus enabled `jaeger-otel` tracing in config)
> 
> OTEL_TRACES_EXPORTER=jaeger;JAEGER_SERVICE_NAME=strimzi-kafka-bridge
> 
> and it works for me from IDE.

Ok, so how do you do it on Kube with Jaeger operator?

config/application.properties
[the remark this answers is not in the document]
scholzj: Let me rephrase the question ... what it the exact relationship between OTel and Jaeger and how do things like Zipkin fit in?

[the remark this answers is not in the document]
alesj: They don't. :-)
Jaeger is a fixed choice, so imho not a good choice of config value.

alesj:
> > This is the env vars I have (plus enabled `jaeger-otel` tracing in config)
> > OTEL_TRACES_EXPORTER=jaeger;JAEGER_SERVICE_NAME=strimzi-kafka-bridge
> > and it works for me from IDE.
> 
> Ok, so how do you do it on Kube with Jaeger operator?

I think the only thing missing here is the OTEL_EXPORTER_JAEGER_ENDPOINT in k8s.

config/application.properties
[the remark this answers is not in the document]
scholzj: What does it mean that it is a fixed choice?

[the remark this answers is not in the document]
alesj: It's either `Jaeger` or nothing. No Zipkin, etc ...

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,212 @@
...
+import io.opentelemetry.instrumentation.kafkaclients.TracingConsumerInterceptor;
+import io.opentelemetry.instrumentation.kafkaclients.TracingProducerInterceptor;
+import io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import io.strimzi.kafka.bridge.config.BridgeConfig;
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.kafka.client.producer.KafkaProducerRecord;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.jetbrains.annotations.Nullable;
```
scholzj, line 30: I don't think we want to use this. And if we have to, it should be probably declared as dependnecy.

config/application.properties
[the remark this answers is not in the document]
alesj: So you could customize your spans / traces names.
It can be fixed to `strimzi-kafka-bridge`.

config/application.properties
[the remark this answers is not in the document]
scholzj: > It's either Jaeger or nothing. No Zipkin, etc ...

But why? The promise of OpenTelemetry is that it works with many different backends including Jaeger, Zipkin, etc.

config/application.properties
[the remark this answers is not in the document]
scholzj: I guess the point is, how does it differ from the environment variable and from the system property?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: Use what? Nullable?

config/application.properties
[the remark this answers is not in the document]
alesj: Yeah ... at the end it's that exporter config + right deps on the classpath.
I can add Zipkin, etc ... if you want.

config/application.properties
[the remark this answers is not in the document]
alesj: I think we follow the same order -- env, sys, config.
I'm happy to drop any or all of them.

config/application.properties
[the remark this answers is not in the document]
scholzj: You do not have to add Zipkin. But this clearly impacts the configuration and the initial question.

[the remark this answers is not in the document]
ppatierno: the reason was this https://github.com/strimzi/strimzi-kafka-bridge/blob/f96cb3f0958bfcad3c875d54ebb78cedc967a8a7/src/main/java/io/strimzi/kafka/bridge/Application.java#L138 ... this `bridge.tracing` is used to enable tracing and we used jaeger thinking about potential future value with other tracing systems. It's even reflected from the CO when configuring the tracing on the bridge through the STRIMZI_TRACING env var here https://github.com/strimzi/strimzi-kafka-bridge/blob/main/bin/docker/kafka_bridge_config_generator.sh#L4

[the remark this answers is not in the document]
alesj: Ah, OK ... this needs to be fixed then as well.

So I guess `jaeger-otel' is not a good name?
Or, actually, we need a new differentiator? By tracing type not impl ...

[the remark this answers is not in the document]
alesj: > Ah, OK ... this needs to be fixed then as well.

OK, this is fine, it takes the env var ...

config/application.properties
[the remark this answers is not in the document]
alesj: As how OTel lookups up config -- env, sys, custom config

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: Removed ...

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: Do we go with all 3 checks or not? Or just exact config property?

config/application.properties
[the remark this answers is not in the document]
ppatierno: So ... the bridge is using the [Vert.x Config](https://vertx.io/docs/vertx-config/java/) for getting the configuration.
As you can see [here ](https://github.com/strimzi/strimzi-kafka-bridge/blob/main/src/main/java/io/strimzi/kafka/bridge/Application.java#L95) we provides two stores for configuration: file (the application.properties) and env vars. Any env var value overrides the same parameter specified in the properties file (because how stores are added matters and in our case is `fileStore` first and then `envStore` as explained [here](https://vertx.io/docs/vertx-config/java/#_overloading_rules)).

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
ppatierno: @alesj I didn't get the question for me? Isn't it the code you added. If you want to know how bridge configuration works, take a look at the previous comment about using [Vert.x Config](https://vertx.io/docs/vertx-config/java/#_overloading_rules) so in our case (due to stores order) env vars first and application.properties file ... but no system properties at all.

alesj: OK, as "said" just before, I'll remove sys prop lookup.

config/application.properties
```diff
@@ -1,7 +1,11 @@
 #Bridge related settings
 bridge.id=my-bridge
-# uncomment the following line to enable Jaeger tracing, check the documentation how to configure the tracer
+# uncomment one the following lines (bridge.tracing) to enable Jaeger tracing, check the documentation how to configure the tracer
+# OpenTracing support
 #bridge.tracing=jaeger
+# OpenTelemetry support
+bridge.tracing=opentelemetry
+bridge.tracing.service-name=strimzi-kafka-bridge
```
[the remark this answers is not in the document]
alesj: Ah, testing ... and forgot to comment it back. Fixed.

bin/docker/kafka_bridge_run.sh
```diff
@@ -68,6 +68,11 @@ if [ "$FIPS_MODE" = "disabled" ]; then
     export JAVA_OPTS="${JAVA_OPTS} -Dcom.redhat.fips=false"
 fi
 
+# enabling OpenTelemetry with Jaeger by default
+if [ -n "$OTEL_SERVICE_NAME" ] && [ -z "$OTEL_TRACES_EXPORTER" ]; then
+  export OTEL_TRACES_EXPORTER="jaeger"
+fi
```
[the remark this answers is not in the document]
alesj: Imo no need for `STRIMZI_TRACING` gates.
I'll remove this snippet, and just leave similar one in bin/kafka_bridge_run.sh

bin/kafka_bridge_run.sh
```diff
@@ -13,4 +13,9 @@ fi
 # Make sure that we use /dev/urandom
 JAVA_OPTS="${JAVA_OPTS} -Dvertx.cacheDirBase=/tmp/vertx-cache -Djava.security.egd=file:/dev/./urandom"
 
+# enabling OpenTelemetry with Jaeger by default
+if [ -n "$OTEL_SERVICE_NAME" ] && [ -z "$OTEL_TRACES_EXPORTER" ]; then
```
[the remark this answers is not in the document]
alesj: Removed.

pom.xml
```diff
@@ -491,6 +546,11 @@
 									<ignoredUnusedDeclaredDependency>io.jaegertracing:jaeger-client</ignoredUnusedDeclaredDependency>
 									<ignoredUnusedDeclaredDependency>org.yaml:snakeyaml</ignoredUnusedDeclaredDependency> <!-- CVE override -->
 									<ignoredUnusedDeclaredDependency>org.apache.tomcat.embed:tomcat-embed-core</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
+									<!-- OpenTelemetry - used via classpath configuration -->
+									<ignoredUnusedDeclaredDependency>io.opentelemetry:opentelemetry-sdk-trace</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
+									<ignoredUnusedDeclaredDependency>io.opentelemetry.instrumentation:opentelemetry-kafka-clients-common</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
+									<ignoredUnusedDeclaredDependency>io.opentelemetry:opentelemetry-exporter-jaeger</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
+									<ignoredUnusedDeclaredDependency>io.grpc:grpc-netty-shaded</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
```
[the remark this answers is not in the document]
alesj: Ah, copy paste ...

alesj:
Closed in favor of a new PR: https://github.com/strimzi/strimzi-kafka-bridge/pull/633

PR #564, opened 2021-12-08
Add s390x support to multi-arch images on Azure CI

kun-lu20:
This PR is to add s390x support to Bridge images on Azure CI. 

Signed-off-by: Kun-Lu <kun.lu@ibm.com>

### DIFF
--- a/.azure/build-pipeline.yaml
+++ b/.azure/build-pipeline.yaml
@@ -33,7 +33,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
   - stage: container_publish
     displayName: Publish Container
     dependsOn:
@@ -49,7 +49,7 @@ stages:
           artifactPipeline: ''
           artifactRunVersion: ''
           artifactRunId: ''
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
   - stage: docs_publish
     displayName: Publish Docs
     dependsOn:
--- a/.azure/cve-pipeline.yaml
+++ b/.azure/cve-pipeline.yaml
@@ -36,7 +36,7 @@ stages:
           artifactPipeline: '${{ parameters.sourcePipelineId }}'
           artifactRunVersion: 'specific'
           artifactRunId: '${{ parameters.sourceBuildId }}'
-          architectures: ['amd64', 'arm64']
+          architectures: ['amd64', 'arm64', 's390x']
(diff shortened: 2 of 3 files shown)


scholzj:
Thanks for the PR. But I think we should have first some testing strategy before adding new platforms.

scholzj:
> Thanks @scholzj, for now we have tested the s390x Bridge image via the official quick start guide and through system tests of Strimzi Kafka Operator.

That is great, but I don't think we can be releasing images for some architecture based on some one-of tests.

kun-lu20:
Yeah, it would be great if we could have a strategy to test it more thoroughly.

kun-lu20:
Hi @scholzj , reg the testing strategy, I am wondering which tests are now running for Arm64 images? Can we follow their path to trigger the same tests on s390x images? We could also provide s390x VMs if you need them for testing or we could run the tests internally and report test results to your site. Please let me know what you think. Thanks!

scholzj:
The Arm64 tests are run manually right now, but are planned to be automated in the future. 

I think there are major differences between Arm64 and s390x:
* Arm64 infrastructure is widely available unlike s390x infrastructure. We have no available infratructure for s390x (TravisCI - aside from their recent security mishaps - does not offer enough resources to run the system tests on s390x).
* We have Arm64 knowledge, experience and interest in the core team. We have absolutely 0 experience with s390x.
* Arm64 is supported by Azure Pipelines which we use as CI. So it is easier to integrate the tests.

These differences are important because they mean that we can test the releases, bugs or issues when we need. We do not depend on anyone with Arm64 support.

> We could also provide s390x VMs if you need them for testing or we could run the tests internally and report test results to your site.

If you have s390x resources, that is a good start. But I'm not sure you can just provide the resources and expect the others to do the work.

The past experience with the different contributions with _s390x or ppc64le support_ (at least my personal view not representing other maintainers) is so far very bad. Someone comes, opens a PR which adds a line with `s390x` or `ppc64le` to some CI file, usually with disabled tests and JavaDoc. And that is it, support for _new architecture_ done. This is how it so far looks with both bridge and the OAuth library. But that is not the right approach and doesn't build trust. 

And I definitely do not want to end up in the same situation with the container images here or in the operators repo. We do not have the capacity to maintain support for these platforms in the Strimzi team. And to be completely honest I'm also not aware of any community interest in s390x (which once again is very different from Arm64).

I know it was not you personally who opened those PRs in the past. But that is the experience we have so far with support for IBM platforms and that is why this cannot be done without proper discussion, planning and proper commitments to do the work.

scholzj (approved):

ppatierno (approved):

kun-lu20:
Thanks @scholzj @ppatierno

PR #582, opened 2022-01-24
Fixes for running ITs with external Kafka and KafkaBridge

im-konge:

This PR fixes issues with some test suites when running ITs against external Kafka and KafkaBridge.

Currently, in `AmqpBridgeIT` we are getting the bootstrap address from `kafkaContainer.getBootstrapServers()` - if we are running Kafka externally, the `kafkaContainer` is empty (`null`), so execution of `kafkaContainer.getBootstrapServers()` will lead to NPE. `kafkaUri` contains bootstrap address of Kafka container or external Kafka, depending on env (if `EXTERNAL_KAFKA` is set to `TRUE`)

Second issue was with `InvalidProducerIT`, which was implemented only for container run -> so if we deployed external Kafka and Bridge, test just failed on `BeforeAll` as the addresses are already binded and the test checks weren't executed against the external components.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/amqp/AmqpBridgeIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/amqp/AmqpBridgeIT.java
@@ -104,13 +104,12 @@ class AmqpBridgeIT extends HttpBridgeITAbstract {
 
     @BeforeAll
     public static void setUp(VertxTestContext context) {
-        adminClientFacade = AdminClientFacade.create(kafkaContainer.getBootstrapServers());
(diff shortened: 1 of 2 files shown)


see-quick (approved):
LGTM...(assuming that it works 👀🦥)

scholzj:
You should probably better explain what it is and why is it needed.

im-konge:
@scholzj I updated the description, just forgot to save it.

@see-quick I ran all tests against containers and external components. Both worked.

ppatierno (approved):

PR #584, opened 2022-01-27
Added disable FIPS mode in OpenJDK to the CHANGELOG

ppatierno:
Just adding the "disable FIPS mode in OpenJDK" to the CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 0.22.0
+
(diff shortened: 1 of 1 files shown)


scholzj (approved):
I guess it is ok. But I intentionally didn't added it because I expect it to end in 0.21.4.

ppatierno:
Yeah but we don't track patch release in the CHANGELOG, so it's better having it at least in the next minor release 0.22.0.

PR #588, opened 2022-01-31
Update Config Providers, Vert.x and Jackson dependencies

scholzj:
This PR updates the Config Providers to the new 1.0.0 versions. It is currently using the RC1 versions. In order to make sure we use the same Jackson version accross all dependencies (Vert.x, direct Jackson dependency and Kubernetes Config Provider), it also updates the Vert.x and Jakcson versions. This aligns all of them on Jackson 2.13.1.

PR #589, opened 2022-02-01
Updated vertx and jackson dependency

ShubhamRwt:

Updated the vertx version to 4.2.4 and also updated jackson since previous jackson version 2.11.3 was giving test failures with vertx 4.2.4


scholzj:
Thsi is already done in #588

Issue #607, opened 2022-03-02
Asynchronous Publishing API


Anthuang:
For the publishing API (https://strimzi.io/docs/bridge/latest/#_send), the current behavior of the bridge is to block upon completion of all send requests, see https://github.com/strimzi/strimzi-kafka-bridge/blob/09a3868600503b4775d76a266b89a43fbf29350b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java#L157. When `acks` is set to `0`, the expected behavior is that the client returns immediately without waiting for any responses from the brokers, and we expect the same from the bridge itself. Since we want to publish events via the bridge on the main thread, latency is very important to us.

It'd be extremely helpful (difference of 100+ ms) if we could have a fully async option which responds immediately to the caller without resolving futures. We employed a hacky fix which just consisted of removing the future resolution and it helped address the latencies.

Let me know if this is a reasonable ask.

scholzj:
I guess that would be some super-unreliable mode which would not fail even when the Kafka cluster is not running. If that is the case, why do you wait for the response if you don't care about reliability? I'm sure your mainline can just fire the HTTP request and not wait for it to complete if you don't care about the reliability.

Anthuang:
Thanks for the quick response!

The biggest advantage I see to checking a response from the bridge is input validation. This allows us to figure out if there's an issue in the code, rather than any issue operation-wise (e.g. brokers are down).

We are currently just ignoring the response as you said as a workaround, but it's really hacky especially since we're stuck using Python 2 (due to legacy reasons).

chadlagore:
I think the issue here is that the behavior doesn't conform to the [documented behavior](https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html#producerconfigs_acks) of `acks`. Specifically:

> acks=0 If set to zero then the producer will not wait for any acknowledgment from the server at all. The record will be immediately added to the socket buffer and considered sent. No guarantee can be made that the server has received the record in this case, and the retries configuration will not take effect (as the client won't generally know of any failures). The offset given back for each record will always be set to -1.

scholzj:
The `acks` sematics applies to the Kafka producepart. Are you saying that when you send with acks=0 the producer is configured with `acks=1` or `acks=-1`?

Anthuang:
The bridge itself has the right configuration when it comes to `acks`, but we do not see the documented behavior:

```
[...]

PR #625, opened 2022-05-27
Add `async` query parameter to control send endpoint behavior

Anthuang:
Open to naming suggestions for the query parameter.

Closes: #607

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -66,6 +66,12 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
         super.open();
     }
 
+    public void maybeClose() {
+        if (this.closing) {
+            this.close();
+        }
+    }
+
     @Override
     @SuppressWarnings("checkstyle:NPathComplexity")
     public void handle(Endpoint<?> endpoint) {
@@ -87,6 +93,8 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
                 return;
             }
         }
+
+        boolean isAsync = Boolean.parseBoolean(routingContext.queryParams().get("async"));
 
         Tracer tracer = GlobalTracer.get();
 
@@ -147,6 +155,17 @@ public class HttpSourceBridgeEndpoint<K, V> extends SourceBridgeEndpoint<K, V> {
         List<HttpBridgeResult<?>> results = new ArrayList<>(records.size());
 
         // start sending records asynchronously
+        if (isAsync) {
+            // if async is specified, return immediately once records are sent
+            this.sendAsyncRecords(records);
+            Tags.HTTP_STATUS.set(span, HttpResponseStatus.NO_CONTENT.code());
+            span.finish();
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.NO_CONTENT.code(),
+                    BridgeContentType.KAFKA_JSON, null);
+            this.maybeClose();
+            return;
+        }
+
         List<Future> sendHandlers = new ArrayList<>(records.size());
         for (KafkaProducerRecord<K, V> record : records) {
             Promise<RecordMetadata> promise = Promise.promise();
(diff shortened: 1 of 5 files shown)


scholzj (commented):
LGTM assuming it works. But should be reviewed by @ppatierno as the Bridge guru.

So nits:
* Should something be added to the docs?
* Should we add a note about this to the CHANGELOG.md file?

Anthuang:
I will wait for @ppatierno's review, but I'm also testing it currently in a staging environment (have to also write the glue code).

I'll look into adding it to the docs/changelog as well

scholzj (commented):
I think you need to rebase it as well. There seems to be some conlfict as you probably created it from some older `main` branch.

bin/docker/docker_setup.sh
```diff
@@ -0,0 +1,7 @@
+java -version
```
scholzj, line 1: I don't think this should be here.

config/bridge.properties
```diff
@@ -0,0 +1,22 @@
+#Bridge related settings
```
scholzj, line 1: This file probably neither?

bin/docker/docker_setup.sh
Anthuang: Oops let me delete it. Messed up while rebasing

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -209,6 +225,12 @@ private JsonObject buildOffsets(List<HttpBridgeResult<?>> results) {
         return jsonResponse;
     }
 
+    private void sendAsyncRecords(List<KafkaProducerRecord<K, V>> records) {
```
Anthuang, line 228: This is to reduce `Cyclomatic Complexity`

Anthuang:
Not sure what's causing this error:

```
2022-05-31T16:07:13.3121820Z [ERROR] io.strimzi.kafka.bridge.http.AdminClientIT  Time elapsed: 30.916 s  <<< ERROR!
2022-05-31T16:07:13.3146277Z java.util.concurrent.TimeoutException: 
2022-05-31T16:07:13.3148525Z The test execution timed out. Make sure your asynchronous code includes calls to either VertxTestContext#completeNow(), VertxTestContext#failNow() or Checkpoint#flag()
```

Anthuang:
That seems to have fixed it (although I'm not sure why), can someone approve the rest of the workflows.

ppatierno:
@Anthuang I checked the last tests running with the error. I am not sure that your latest change (about closing) is the fix for the test not working even because the "async" related stuff is not used at all in the failing tests. I think you should revert back your latest change because as you said "That seems to have fixed it (although I'm not sure why)" and investigate the root cause of that test failure. It's possible there is some other subtle problem or the test is flaky.

ppatierno:
My concern is that we don't have these tests failing on our main branch so I was wondering if further investigation was needed.

scholzj:
Well, it seemed to have passed now without any changes to existing tests. So maybe it is fixed now? TBH, in #629 (where the change is actually not used during the unit tests at all) I had also quite a lot of failures and needed restarts :-/.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -66,6 +66,12 @@ public void open() {
         super.open();
     }
 
+    public void closeOnError() {
```
ppatierno, line 69: why this name? I mean, from the code inside, it doesn't show up that it's closing when there is an error.

Anthuang: Do you have a suggestion for the name? I'm just trying to keep the `Cyclomatic Complexity` within the bounds without changing too much code with this PR.

ppatierno: Let's try with something like `maybeClose` taking into account that the `closing` has to be validate.

ppatierno (changes requested):

ppatierno (approved):
LGTM. Thanks! @scholzj would you mind to have another pass please?

scholzj (approved):
Thanks for the PR.

PR #627, opened 2022-05-30
Bump the kafka version to 3.2.0 and vertx version to 4.3.1

ShubhamRwt:

This PR bumps the Kafka version in the bridge to 3.2.0 since vertx 4.3.1 is released and is compatible to work with kafka version >= 3.0.0 also.The PR replaces the deprecated `routingContext.getbody()` methods with the new alternatives and also other fixes

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpAdminClientEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpAdminClientEndpoint.java
@@ -156,7 +156,7 @@ public class HttpAdminClientEndpoint extends AdminClientEndpoint {
                 root.put("partitions", partitionsArray);
                 HttpUtils.sendResponse(routingContext, HttpResponseStatus.OK.code(), BridgeContentType.KAFKA_JSON, root.toBuffer());
 
-            } else if (done.cause() instanceof UnknownTopicOrPartitionException) {
+            } else if (done.cause() != null && done.cause().getCause() instanceof UnknownTopicOrPartitionException) {
                 HttpBridgeError error = new HttpBridgeError(
                         HttpResponseStatus.NOT_FOUND.code(),
                         done.cause().getMessage()
@@ -187,7 +187,7 @@ public class HttpAdminClientEndpoint extends AdminClientEndpoint {
                     });
                 }
                 HttpUtils.sendResponse(routingContext, HttpResponseStatus.OK.code(), BridgeContentType.KAFKA_JSON, root.toBuffer());
-            } else if (describeTopicsResult.cause() instanceof UnknownTopicOrPartitionException) {
+            } else if (describeTopicsResult.cause() != null && describeTopicsResult.cause().getCause() instanceof UnknownTopicOrPartitionException) {
                 HttpBridgeError error = new HttpBridgeError(
                         HttpResponseStatus.NOT_FOUND.code(),
                         describeTopicsResult.cause().getMessage()
@@ -233,7 +233,7 @@ public class HttpAdminClientEndpoint extends AdminClientEndpoint {
                     HttpUtils.sendResponse(routingContext, HttpResponseStatus.NOT_FOUND.code(),
                             BridgeContentType.KAFKA_JSON, error.toJson().toBuffer());
                 }
-            } else if (describeTopicsResult.cause() instanceof UnknownTopicOrPartitionException) {
+            } else if (describeTopicsResult.cause() != null && describeTopicsResult.cause().getCause() instanceof UnknownTopicOrPartitionException) {
                 HttpBridgeError error = new HttpBridgeError(
                         HttpResponseStatus.NOT_FOUND.code(),
                         describeTopicsResult.cause().getMessage()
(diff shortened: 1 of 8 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -276,7 +276,7 @@ private void createConsumer(RoutingContext routingContext) {
         this.httpBridgeContext.setOpenApiOperation(HttpOpenApiOperations.CREATE_CONSUMER);
 
         // check for an empty body
-        JsonObject body = routingContext.getBody() != null && routingContext.getBody().length() != 0 ? routingContext.getBodyAsJson() : new JsonObject();
+        JsonObject body = routingContext.body().buffer() != null && routingContext.body().buffer().length() != 0 ? routingContext.getBodyAsJson() : new JsonObject();
```
tomncooper, line 279: So has the new vertx version changed the inplementation of RoutingContext?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -502,8 +502,8 @@ public void handle(Endpoint<?> endpoint, Handler<?> handler) {
         JsonObject bodyAsJson = EMPTY_JSON;
         try {
             // check for an empty body
-            if (routingContext.getBody() != null && routingContext.getBody().length() != 0) {
-                bodyAsJson = routingContext.getBodyAsJson();
+            if (routingContext.body().buffer() != null && routingContext.body().buffer().length() != 0) {
+                bodyAsJson = routingContext.body().asJsonObject();
```
tomncooper, line 506: Everywhere else you are using `.body().buffer().....` why do you not need that here? Dos the json method handle that for you?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerIT.java
```diff
@@ -475,7 +475,7 @@ void createConsumerWithWrongFetchMinBytes(VertxTestContext context) throws Inter
     @Test
     void createConsumerWithNotExistingParameter(VertxTestContext context) throws InterruptedException, TimeoutException, ExecutionException {
         checkCreatingConsumer("foo", "bar", HttpResponseStatus.BAD_REQUEST,
-                "Validation error on:  - provided object should not contain additional properties", context);
+                "Validation error on:  - Provided object contains unexpected additional property: foo", context);
```
tomncooper, line 478: Not sure "foo" should still be in there?

src/test/java/io/strimzi/kafka/bridge/http/HttpCorsIT.java
```diff
@@ -110,6 +110,7 @@ public void testCorsNotEnabled(VertxTestContext context) {
                         client.request(HttpMethod.POST, 8080, "localhost", "/consumers/1/instances/1/subscription")
                                 .putHeader("Origin", "https://evil.io")
                                 .send(ar2 -> context.verify(() -> {
+                                    System.out.println(ar2.result().body());
```
tomncooper, line 113: You need to make sure you are removing these debug print lines

tomncooper (changes requested):
Just a few nits to fix

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ShubhamRwt: The alternative suggested for routingContext.getBodyasJson() was routingContext.body().asJsonObject().

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Yeah they have deprecated the `getBody` and other methods, everything seems to be under a `RequestBody` object returned by `body` method.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: As I can see from the current implementation here https://github.com/vert-x3/vertx-web/blob/master/vertx-web/src/main/java/io/vertx/ext/web/RequestBody.java#L131 you could change the if condition just checking `rounting.body().isEmpty()` instead of checking null and length on it.

src/test/java/io/strimzi/kafka/bridge/http/HttpCorsIT.java
```diff
@@ -180,8 +179,8 @@ public void testCorsOriginAllowed(VertxTestContext context) {
                                 .putHeader("Origin", "https://strimzi.io")
                                 .putHeader("content-type", BridgeContentType.KAFKA_JSON)
                                 .sendJsonObject(topicsRoot, ar2 -> context.verify(() -> {
-                                    //we are not creating a topic, so we will get a 404 status code
-                                    assertThat(ar2.result().statusCode(), is(404));
+                                    //we are not creating a topic, so we will get a 400 status code
+                                    assertThat(ar2.result().statusCode(), is(400));
```
ppatierno, line 183: Why this change?

src/main/java/io/strimzi/kafka/bridge/http/HttpAdminClientEndpoint.java
```diff
@@ -156,7 +156,7 @@ public void doGetTopic(RoutingContext routingContext) {
                 root.put("partitions", partitionsArray);
                 HttpUtils.sendResponse(routingContext, HttpResponseStatus.OK.code(), BridgeContentType.KAFKA_JSON, root.toBuffer());
 
-            } else if (done.cause() instanceof UnknownTopicOrPartitionException) {
+            } else if (done.cause() != null && done.cause().getCause() instanceof UnknownTopicOrPartitionException) {
```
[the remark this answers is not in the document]
ShubhamRwt: done.cause() was returning `CompletionException` instead of the `UnknownTopicOrPartitionException` or you can also say that with the latest vertx and kafka, the `UnknownTopicOrPartitionException` comes wrapped inside the `CompletionException`. So I had to use done.cause().getCause() get the `UnknownTopicOrPartitionException` out of the done.cause() which was `CompletionException`

ppatierno (approved):

PR #633, opened 2022-06-29
Add OpenTelemetry support

alesj:
This PR adds OpenTelemetry support.
It abstract previous OpenTracing / Jaeger usage -- adding thin abstraction layer to hide actual implementation usage.
Most of the old code was used as it is, only `poll` or `consume` was changed away from interceptor usage - as we couldn't model trace/spans as expected.
Async send needed some extra work for OTel, due to its ThreadLocal current span propagation.

Current `bridge.tracing` property must be set to `opentelemetry` to use OTel tracing.


src/main/java/io/strimzi/kafka/bridge/tracing/OpenTracingHandle.java
```diff
@@ -0,0 +1,190 @@
...
+        public SpanHandle<K, V> span(RoutingContext routingContext) {
+            return buildSpan(spanBuilder, routingContext);
+        }
+    }
+
+    @Override
+    public void kafkaConsumerConfig(Properties props) {
+        // ignore interceptor, we handle consuming + tracing
+        // props.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, TracingConsumerInterceptor.class.getName());
+    }
```
ppatierno, line 138: same question as for the OpenTelemetry handle implementation.

tombentley:
@alesj please could you provide a description for this PR, and an issue reference if there is one? Thanks.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,261 @@
...
+        public SpanHandle<K, V> span(RoutingContext routingContext) {
+            return buildSpan(spanBuilder, routingContext);
+        }
+    }
+
+    @Override
+    public void kafkaConsumerConfig(Properties props) {
+        // ignore interceptor, we handle consuming + tracing
+        //props.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, TracingConsumerInterceptor.class.getName());
+    }
```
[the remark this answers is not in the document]
alesj: This is not needed anymore, as we handle consume on our own.
I guess I can remove this.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Override
+    public <K, V> SpanBuilderHandle<K, V> builder(RoutingContext routingContext, String operationName) {
+        SpanBuilder spanBuilder = getSpanBuilder(routingContext, operationName);
+        return new OTelSpanBuilderHandle<>(spanBuilder);
+    }
+
+    @Override
+    public <K, V> void handleRecordSpan(SpanHandle<K, V> parentSpanHandle, KafkaConsumerRecord<K, V> record) {
+        String operationName = String.format("%s %s", record.topic(), MessageOperation.RECEIVE);
```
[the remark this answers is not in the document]
alesj: This is not part of `poll`. This is `consume`, child (or children) of `poll`.
In OpenTracing it's prefixed by "From_", where in OTel this `MessageOperation.RECEIVE` is used with Kafka consume OTel code (agent or interceptor ...).

alesj:
> @alesj please could you provide a description for this PR, and an issue reference if there is one? Thanks.

Short description provided.
No issue ref afaik.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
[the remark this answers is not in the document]
ppatierno: You are right but why we don't need that with OpenTracing?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,259 @@
...
+    public void initialize() {
+        AutoConfiguredOpenTelemetrySdk.initialize();
+    }
+
+    private static Tracer get() {
+        return GlobalOpenTelemetry.getTracer(COMPONENT);
+    }
+
+    private SpanBuilder getSpanBuilder(RoutingContext routingContext, String operationName) {
+        Tracer tracer = get();
```
ppatierno, line 97: Can't we get the `Tracer` instance just one time when we create the handle instead of each time we need a span? It's just for better readiness because AFAICS in the OpenTelemetry code, the instance is anyway created one time and then just returned.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,259 @@
...
+        SpanBuilder spanBuilder;
+        Context parentContext = propagator().extract(Context.current(), routingContext, RCG);
```
ppatierno, line 99: What `RCG` stands for?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTracingHandle.java
```diff
@@ -0,0 +1,188 @@
...
+    @Override
+    public <K, V> SpanHandle<K, V> span(RoutingContext routingContext, String operationName) {
+        Tracer.SpanBuilder spanBuilder = getSpanBuilder(routingContext, operationName);
+        return buildSpan(spanBuilder, routingContext);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Override
+    public <K, V> void handleRecordSpan(SpanHandle<K, V> parentSpanHandle, KafkaConsumerRecord<K, V> record) {
+        Tracer tracer = GlobalTracer.get();
```
ppatierno, line 80: Ditto as OpenTelemetry, getting the `Tracer` just one time on creation.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,259 @@
...
+        public OTelSpanHandle(Span span) {
+            this.span = span;
+            this.scope = span.makeCurrent();
+        }
+
+        @Override
+        public void prepare(KafkaProducerRecord<K, V> record) {
+            String uuid = UUID.randomUUID().toString();
+            SPANS.put(uuid, span);
+            record.addHeader(X_UUID, uuid);
```
ppatierno, line 210: I tried to follow the flow. I see that when we send a record, you are calling this `prepare` to generate a UUID and it's put in the header of the Kafka record but when the `onSend` of the interceptor is called, you are removing this UUID before sending the record. So what's the purpose of this UUID in the record header between the prepare and onSend?

ppatierno: My question was not related to the `SPANS` map but to the headers in the Kafka record. You are adding the UUID in the header of the record here, but then before sending the record (in the interceptor) you are removing the UUID ... so I don't understand the reason for having the UUID being added here (then removed, without any use, later).

ppatierno: Actually looking at the code even OpenTracing sets a `TracingProducerInterceptor` class in the configuration but there is no special implementation of it compared to OpenTelemetry (your `ContextAwareTracingProducerInterceptor`). Why?

ppatierno: I think I have got the reason. You are getting the HTTP bridge "send" span and you are propagate that span as parent on the interceptor send thread so that it will be parent of the topic send provided by the base `TracingProducerInterceptor` implementation. My only question is why all this stuff is not needed in OpenTracing ...

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -158,8 +121,7 @@ public Iterator<Map.Entry<String, String>> iterator() {
         if (isAsync) {
             // if async is specified, return immediately once records are sent
             this.sendAsyncRecords(records);
-            Tags.HTTP_STATUS.set(span, HttpResponseStatus.NO_CONTENT.code());
-            span.finish();
+            span.finish(HttpResponseStatus.NO_CONTENT.code());
             HttpUtils.sendResponse(routingContext, HttpResponseStatus.NO_CONTENT.code(),
```
tombentley, line 125: Why do we finish the span before sending the response here, but the other way around in the code above? I'd have expected the order should be consistent.

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
```diff
@@ -0,0 +1,98 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+/**
+ * Base for OpenTracing and OpenTelemetry (manual) tests.
+ * <p>
+ * Test will only run if the bridge AND tracing server are up-n-running.
+ */
+@ExtendWith(VertxExtension.class)
+public abstract class TracingTestBase {
```
tombentley, line 43: No test for the OpenTracing impl?

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
```diff
@@ -0,0 +1,98 @@
...
+                assertThat(response.statusCode(), is(HttpResponseStatus.OK.code()));
+            });
+            context.completeNow();
+        };
+    }
+
+    @BeforeEach
+    public void setUp() {
+        assumeServer(String.format("http://%s:%s", Urls.BRIDGE_HOST, Urls.BRIDGE_PORT)); // bridge
+        assumeServer("http://localhost:16686"); // jaeger
```
tombentley, line 69: I think it would be better if we could use testcontainers to spin up a server. I admit I've no idea how much work that would really, but maybe it's not so much?

tombentley (commented):
Thanks @alesj! I left a few comments and questions.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: > My only question is why all this stuff is not needed in OpenTracing

It looks like OpenTracing propagates current span differently - I guess not via ThreadLocal ...
I didn't go into details for OpenTracing, since it worked as expected.

This surprised me in OTel, and was actually tricky to find this async send at first - this resulted in spans not being related, hence the "hacky" interceptor workaround.

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -167,9 +168,9 @@ protected void initConsumer(boolean shouldAttachBatchHandler, Properties config)
         props.putAll(kafkaConfig.getConfig());
         props.putAll(kafkaConfig.getConsumerConfig().getConfig());
         props.put(ConsumerConfig.GROUP_ID_CONFIG, this.groupId);
-        if (this.bridgeConfig.getTracing() != null) {
-            props.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, TracingConsumerInterceptor.class.getName());
-        }
+
+        TracingHandle tracing = TracingUtil.getTracing();
+        tracing.kafkaConsumerConfig(props);
```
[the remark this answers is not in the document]
alesj: It used to do something.
I guess it can be removed now.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
alesj: Afaik, I didn't change the order ... but will check again.
But yeah, consistent order makes sense.
Which should it be -- finish span 1st and then response, or vice-versa?

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
alesj: There was none to start, so I just added OTel.
Let me add one for OpenTracing then as well ...

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
alesj: With testcontainers it's chicken-n-egg problem ... you would first need to build latest Bridge image, only to be used in tests then.
Perhaps I can easily make it work with running Application::main with the right args ...

alesj: Ah, now I know why I added that service.name property to Bridge config ... otherwise you can only enable OpenTracing via env vars ... which makes it hard (if not impossible) to test.

alesj: Something like this is then required - custom generic container:
* https://github.com/alesj/test-clients/tree/tests1/testsuite/src/test/java/io/strimzi/testclients/test/support

I can give it a shot if it's really needed ...

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
ppatierno: This is where OpenTracing works. The `onSend` of its default interceptor calls this `buildAndInjectSpan` which is able to extract the current span context and set it as child.

https://github.com/opentracing-contrib/java-kafka-client/blob/31ce5260279c2cadf3d69c0acbd50f024afe4660/opentracing-kafka-client/src/main/java/io/opentracing/contrib/kafka/TracingKafkaUtils.java#L80

So following the `TracingKafkaUtils.extraSpanContext`, it's doing some action through the record headers where the context is stored and is able to get it.

https://github.com/opentracing-contrib/java-kafka-client/blob/31ce5260279c2cadf3d69c0acbd50f024afe4660/opentracing-kafka-client/src/main/java/io/opentracing/contrib/kafka/TracingKafkaUtils.java#L43

In OpenTelemetry, the default interceptor has the `onSend` calling this `buildAndInjectSpan` which gets the current context this way:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/cd528e98fbbaa9bf6e48351d908d43e28a6e0073/instrumentation/kafka/kafka-clients/kafka-clients-2.6/library/src/main/java/io/opentelemetry/instrumentation/kafkaclients/KafkaTelemetry.java#L84

It's different from the OpenTracing one and record headers are not in the picture AFAICS.

Not an OpenTelemetry expert but it smells something is missing here for having it working out of box without any hack.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
tombentley: I don't know for certain, but I would finish the span first and send the response. That way the tracing should definitely be present to clients that have observed the HTTP response.

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
tombentley: Yeah, it realise it's not a _loss_ of coverage. Thanks for adding one!

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: > record headers are not in the picture

They are used / read here:
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/cd528e98fbbaa9bf6e48351d908d43e28a6e0073/instrumentation/kafka/kafka-clients/kafka-clients-2.6/library/src/main/java/io/opentelemetry/instrumentation/kafkaclients/KafkaTelemetry.java#L93

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
tombentley: Looking at Ales's branch I'm inclined to thiat bridge support for the test-clients would be a great addition. @ppatierno @scholzj wdyt?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
ppatierno: I spent a quite of time debugging both OpenTracing and OpenTelemetry internals.
So you are right that OpenTracing is heavily based on using the Kafka record headers through the `uber-trace-id`. Thanks to it to be set in the Vert.x event loop thread (where the HTTP send span is created), the interceptor in a different Vert.x worker thread is able to get the parent span from that `uber-trace-id` working out of box.
With OpenTelemetry, the interceptor doesn't work like this and the context which makes sense in the Vert.x event loop thread (where the HTTP send span is created) doesn't exist in the interceptor (in the Vert.x worker thread) because everything OTel uses is a ThreadLocal storage for context (which of course isn't shared across threads :-)).
I am going to start a discussion in the OTel community because it means that the Kafka support doesn't work out of box and always needs an hack (as you are currently done). I was wondering why they didn't use the same OpenTracing way.

ppatierno: I just opened a issue/discussion in the open-telemetry java related project https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6297

pom.xml
```diff
@@ -111,6 +111,9 @@
 		<jaeger.version>1.6.0</jaeger.version>
 		<opentracing.version>0.33.0</opentracing.version>
 		<opentracing-kafka-client.version>0.1.15</opentracing-kafka-client.version>
+		<opentelemetry.version>1.9.0-alpha</opentelemetry.version>
```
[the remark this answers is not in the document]
ppatierno: Would you mind to upgrade both the versions (stable and alpha) to 1.15.0.
At same time I would do the other way around, so having a `opentelemetry.version` property for the stable one (as we have for all the other properties) and then using a suffix, `opentelemenetry.alpha-version` property for the alpha one.

src/main/java/io/strimzi/kafka/bridge/tracing/TracingConstants.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public static final String JAEGER = "jaeger";
+
+    public static final String JAEGER_OPENTRACING = JAEGER;
+    public static final String OPENTELEMETRY = "opentelemetry";
+
+    public static final String OPENTELEMETRY_SERVICE_NAME_ENV_KEY = "OTEL_SERVICE_NAME";
+    public static final String OPENTELEMETRY_SERVICE_NAME_PROPERTY_KEY = "otel.service.name";
+    public static final String OPENTELEMETRY_TRACES_EXPORTER_ENV_KEY = "OTEL_TRACES_EXPORTER";
+    public static final String OPENTELEMETRY_TRACES_EXPORTER_PROPERTY_KEY = "otel.traces.exporter";
```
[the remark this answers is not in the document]
ppatierno: @alesj I don't see this resolved by removing the `XXXXX_PROPERTY_KEY` constants

src/main/java/io/strimzi/kafka/bridge/tracing/TracingConstants.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Tracing constants.
+ */
+public final class TracingConstants {
+    public static final String COMPONENT = "strimzi-kafka-bridge";
+    public static final String KAFKA_SERVICE = "kafka";
+
+    public static final String JAEGER = "jaeger";
+
+    public static final String JAEGER_OPENTRACING = JAEGER;
```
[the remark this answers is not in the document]
ppatierno: @alesj I don't see this fixed, still two constants?

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
```diff
@@ -0,0 +1,29 @@
...
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
+
+import java.util.Properties;
+
+/**
+ * Simple interface to abstract tracing between legacy OpenTracing and new OpenTelemetry.
+ */
+public interface TracingHandle {
+    String envName();
```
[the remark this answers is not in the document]
ppatierno: @alesj still not renamed?

src/main/java/io/strimzi/kafka/bridge/tracing/TracingUtil.java
```diff
@@ -0,0 +1,78 @@
...
+        for (KafkaHeader header : record.headers()) {
+            headers.put(header.key(), header.value().toString());
+        }
+        return headers;
+    }
+
+    static void addProperty(Properties props, String key, String value) {
+        String previous = props.getProperty(key);
+        if (previous != null) {
+            props.setProperty(key, previous + "," + value);
```
ppatierno, line 73: you are using this for setting just the producer interceptor in both the handles. We don't use this for some list of values assigned to a property. I would just remove this method from here and setting the property directly in the `addTracingPropsToProducerConfig` for each handle.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/tracing/TracingUtil.java
ppatierno: @alesj never mind just noticed the comment from Tom B. about a possible interceptor in the list. I didn't remember that it was possible to add more than one.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: @ppatierno this Kafka OTel interceptor support is actually my implementation :-), but I did it by the guidelines of existing integration implementations (via that Instrumenter API). And that API, as I mentioned, needs an explicit current context / span.
The propagation or/aka storage can be changed - as ContextStorage is an abstraction, but I didn't find any other useful impl, where ThreadLocal is the default one.

src/main/java/io/strimzi/kafka/bridge/tracing/TracingConstants.java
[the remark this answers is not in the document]
alesj: OTel already works in Quarkus-ish way, it also understands props.
And since we decided to support old JAEGER_SERVICE_NAME, this is the way to actually set OTel service name.
Either we drop both or none.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
ppatierno: @alesj at this point the question is how the other implementations deal with cross-thread context ... I don't think that Kafka is the only use case. The current implementation is not that useable in the end because any user would need your hack by creating a custom interceptor class, handling an map of spans and so on.

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
```diff
@@ -0,0 +1,29 @@
...
+    String serviceName(BridgeConfig config);
+    void initialize();
+
+    <K, V> SpanBuilderHandle<K, V> builder(RoutingContext routingContext, String operationName);
```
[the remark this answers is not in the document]
alesj: You're not building the span, you get back a `builder` with which you then build a span.
Usually such methods are named `builder`.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: Yeah, good question -- any async impl (not just Kafka, etc) would have problems with this.
I guess they manage to pass-on the span somehow, w/o that map hack, etc ... ?

alesj: I guess we need to find a way to wrap Vertx's executor in this Context::wrap .. any idea where this could be done?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,261 @@
...
+                System.setProperty(OPENTELEMETRY_SERVICE_NAME_PROPERTY_KEY, serviceName);
+            }
+        }
+        if (serviceName != null) {
+            if (System.getenv(OPENTELEMETRY_TRACES_EXPORTER_ENV_KEY) == null && System.getProperty(OPENTELEMETRY_TRACES_EXPORTER_PROPERTY_KEY) == null) {
+                System.setProperty(OPENTELEMETRY_TRACES_EXPORTER_PROPERTY_KEY, JAEGER); // it wasn't set in script
+            }
+        }
+        return serviceName;
+    }
```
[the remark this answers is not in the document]
ppatierno: But if the user sets `bridge.tracing=opentelemetry` in the application property and then forgets to set the `OTEL_SERVICE_NAME` env var, I don't think it makes much sense to fall back to the Jaeger service name.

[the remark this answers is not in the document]
alesj: It could have that env var set from before ... and just changes tracing impl type ...
But yeah, let me know what's put final decision, and I'll adjust the code - if needed.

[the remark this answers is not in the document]
ppatierno: > It could have that env var set from before ... and just changes tracing impl type ...

You are saying that the initial situation is:

```shell
JAEGER_SERVICE_NAME = servicename
bridge.tracing=jaeger
```

then the user just changes the `bridge.tracing=opentelemetry` but forgets to set `OTEL_SERVICE_NAME`.
Following the code, it's going to set the `otel.service.name` with the Jaeger env var.
I have tried but nothing is showed in the Jaeger UI. The service name doesn't show up in the drop down list at all.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,227 @@
...
+import static io.strimzi.kafka.bridge.tracing.TracingConstants.OPENTELEMETRY_TRACES_EXPORTER_ENV_KEY;
+import static io.strimzi.kafka.bridge.tracing.TracingConstants.OPENTELEMETRY_TRACES_EXPORTER_PROPERTY_KEY;
+
+/**
+ * OpenTelemetry implementation of Tracing.
+ */
+class OpenTelemetryHandle implements TracingHandle {
+
+    private Tracer tracer;
+    private ExecutorService service;
```
ppatierno, line 49: what's the reason for storing the `ExecutorService` instance here? I see it's never being used but just set.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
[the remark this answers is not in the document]
alesj: It should work?!
You're saying this `old school` doesn't work?

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
```diff
@@ -0,0 +1,88 @@
...
+    void initialize();
+
+    /**
+     * Get custom executor service if needed.
+     * Else return null.
+     *
+     * @param service current executor service
+     * @return custom executor service or null
+     */
+    default ExecutorService get(ExecutorService service) {
```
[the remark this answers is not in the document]
alesj: Yeah ... any good idea?
`getExecutorService` .. we all know what we're gonna get?!

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: It's been set, and used.

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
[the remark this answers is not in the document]
ppatierno: Tbh without checking the method signature it's not clear what you are getting. In Java you could also use the `var` keyword and you don't know what's on the other side, returned by the method. From a class design pov something like `TracingHandle.get` let me think we are returning an instance of `TracingHandle` ? or maybe it's not really clear what.
Tbh, yes I would name it better. But the code is "wrapping"? So it's a method getting an `ExecutorService` and returning an `ExecutorService`. It sounds like an "adapter"?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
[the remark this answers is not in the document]
ppatierno: Ok. It's working. I think that yesterday I didn't notice that the problem was on startup due to a missing dependency you have fixed in the evening.

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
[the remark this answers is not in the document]
alesj: Well, the only impl atm is wrapping, but it could be doing anything.
I guess `adapt` would be best?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
alesj: It's returned on every `BridgeExecutorServiceFactory::delegate()` call ...

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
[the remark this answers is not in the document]
ppatierno: +1

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
ppatierno: I see now but ... I don't think that the `ExecutorService` instance belongs to a telemetry handle.
It should be created one time in the application (based on tracing enabled or not and which one) and provided to Vert.x.
It would avoid to have this `delegate()` stuff called every time.
The main problem is that we know about if tracing is enabled through the `ConfigRetriever` ...

```java
ConfigRetriever retriever = ConfigRetriever.create(vertx, options);
```

where we already need the Vert.x instance.
We should know about telemetry setting before that and getting the `ExecutorService` accordingly. 
I am thinking how to get a good solution for that ... unless you already have one :-)

alesj: Yeah, that's the main problem and reason for that `delegate()`.
Otoh, those calls are not costly, they simply return existing instances, no additional computation.
No new idea for a better solution atm unfortunately. :-)

ppatierno: So ... this kind of solution to instrument Kafka could be fine with Vert.x but when using Kafka Java native client you don't have access to the executor, executor service or thread where the interceptor onSend is executed.
Thinking in this way, I had an idea about avoiding the executor service wrapping but doing a step back with a custom interceptor (which is anyway the only way you would have with Kafka Java native).
The implementation would be this one ...

```java
public static class ContextAwareTracingProducerInterceptor<K, V> extends TracingProducerInterceptor<K, V> {
        @Override
        public ProducerRecord<K, V> onSend(ProducerRecord<K, V> record) {
            return Context.current().wrapFunction(super::onSend).apply(record);
        }
    }
}
```

Just really simple. It wraps the `onSend` of the default producer interceptor and what the `wrapFunction` does is simple as well:

```java
default <T, U> Function<T, U> wrapFunction(Function<T, U> function) {
    return t -> {
      try (Scope ignored = makeCurrent()) {
        return function.apply(t);
      }
    };
  }
```

So it's kind of something you did in the previous implementation.
The only thing is that I have to debug more because it's not working in the sense that the HTTP span and Kafka span are not related together when sending :-(
Can you give it a try as well? Maybe there is something I am doing wrong and I am blind to see it?
This solution would avoid to inject an executor service and would be a kind of pattern for Kafka producer instrumentation (when the native Java client is used)

ppatierno: I guess it's not worked AS EXPECTED :-) ... because in the custom interceptor we are already in the other thread in the Kafka client so the context is different and empty ... it's too late :-(

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Adapt executor service if needed.
+     * Else return null.
+     *
+     * @param service current executor service
+     * @return adapted executor service or null
+     */
+    default ExecutorService adapt(ExecutorService service) {
+        return null;
```
ppatierno, line 47: not sure the logic is right. If not adapting it should return the executor service as it is. 
It would also mean simplifying the logic in the delagation we have in the  `BridgeExecutorServiceFactory` ...

```java
private ExecutorService delegate() {
     return TracingUtil.getTracing().adapt(original);
}
```

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
alesj: Ah yes, that equally works, let me simplify it.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
ppatierno: Vert.x seems to have a TP support for tracing here https://github.com/eclipse-vertx/vertx-tracing
Of course refactoring everything is not feasible and the feature is TP but what we should try is using the `VertxContextStorageProvider` instead of the default one which is based on thread local.

https://github.com/eclipse-vertx/vertx-tracing/blob/master/vertx-opentelemetry/src/main/java/io/vertx/tracing/opentelemetry/VertxContextStorageProvider.java

This Vert.x storage seems to be able to correlate Vert.x context with OpenTelemetry context and that could help.
Right now I am not able to set this storage yet :-(

alesj: > Right now I am not able to set this storage yet :-(

What's the problem?
Or should I have a look at it?

alesj: VCSP is automatically found -- via META-INF/services.
But ...

```
Exception in thread "vert.x-internal-blocking-2" java.lang.NullPointerException: Cannot invoke "io.vertx.core.Context.getLocal(Object)" because "vertxCtx" is null
	at io.vertx.tracing.opentelemetry.VertxContextStorageProvider$VertxContextStorage.attach(VertxContextStorageProvider.java:27)
	at io.vertx.tracing.opentelemetry.VertxContextStorageProvider$VertxContextStorage.attach(VertxContextStorageProvider.java:23)
	at io.opentelemetry.context.Context.makeCurrent(Context.java:202)
	at io.opentelemetry.context.Context.lambda$wrap$1(Context.java:211)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:831)
```

Missing something more ... ?

ppatierno: My problem was adding the `vertx-tracing` dep but not noticing that it was there already as `test` scope which was overriding mine. Now it seems to work :-)

src/test/java/io/strimzi/kafka/bridge/tracing/TracingTestBase.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    protected abstract TracingOptions tracingOptions();
+
+    @Test
+    public void testSmoke(VertxTestContext context) throws Exception {
```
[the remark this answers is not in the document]
ppatierno: @tombentley do you think we should keep it anyway or just remove? Maybe opening an issue about adding proper tracing tests later?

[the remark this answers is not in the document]
alesj: > Maybe opening an issue about adding proper tracing tests later?

+1

ppatierno:
With the current status of the PR we are avoiding the hack we had before about tracking spans with a UUID for using the right tracing context in the Kafka producer interceptor due to the async nature of Vert.x and different threads usage.
We are now using the `VertxContextStorageProvider` which an implementation of the context storage for OpenTelemetry in the Vert.x land which takes into account bringing OpenTelemetry context via Vert.x context.

Unluckily the current solution has a problem when an HTTP client uses the "keep-alive" connection enabled.
In this scenario, when the bridge is used to send a message, on the first request the HTTP send span and the underneath Kafka send message span are related correctly (the latter is CHILD OF the former).
All the subsequent HTTP send requests are not traced correctly because the above two spans are totally independent.
After a long investigation, it turned out that it is due to the Vert.x Kafka client producer on the bridge being the same and reused on all HTTP requests when "keep-alive" connection is enabled.
The issue is that the Vert.x Kafka client is taking the initial Vert.x context when it's created for the first time (on the first HTTP request and connection established), and then it uses this one for executing any other blocking call (see the Kafka send).
Of course, each incoming HTTP request has its own context (bringing tracing) instead.
So the first time everything works as expected because the two contexts are actually the same. From the second request, the Kafka client uses the first "old" context (with no tracing anymore) and not the right "current" context (bringing tracing).

Of course with no "keep-alive" everything works fine each time, because a new Kafka client is created for every HTTP request and the context is always the "right" one.
We are in touch with Vert.x people to get advice and about the behaviour of the Vert.x Kafka client.

Meanwhile we already have the "workaround" which is actually a middle solution (pushed by Ales during these days to avoid the above mentioned hack) about using an executor service which is wrapped through the Context (in OpenTelemetry) in order to use the right context in Vert.x.
This solution was suggested by OpenTelemetry people and it seems to be the "official" one when you don't have the right context storage (which we have but driving to the above problem).

This is more or less the status of the art.

If there is no trick for the Vert.x Kafka client or we would need a change on the client so waiting for a new release, I would move forward with this PR by using the last "workaround" and then tracking an issue to come back using the storage later.

Last consideration is that Vert.x seems to have an official support for tracing now, both for OpenTelemetry and OpenTracing but seems to be still in TP and anyway using it would need a more bigger refactoring on the bridge because such support handle the HTTP part of the tracing for us our of box, so we should remove our code. I am against this solution in the near future.

I would like to know what the other maintainers think about the overall picture of this. @tombentley @scholzj ?

bin/kafka_bridge_run.sh
```diff
@@ -13,4 +13,9 @@ fi
 # Make sure that we use /dev/urandom
 JAVA_OPTS="${JAVA_OPTS} -Dvertx.cacheDirBase=/tmp/vertx-cache -Djava.security.egd=file:/dev/./urandom"
 
+# enabling OpenTelemetry with Jaeger by default
+if [ -n "$OTEL_SERVICE_NAME" ] && [ -z "$OTEL_TRACES_EXPORTER" ]; then
+  export OTEL_TRACES_EXPORTER="jaeger"
+fi
```
ppatierno, line 19: aren't we doing this check in the `OpenTelemetryHandle` already (together with setting jaeger as default exporter)? Why here as well?

alesj: Just in case you don't run Bridge via this script ... which you don't if you use IDE.

ppatierno: Tbh the Strimzi bridge is supposed to run via the script. If you are not using it but the IDE, it's your responsibility to set the right env vars to make it working. I would remove the duplicated code and leave just the script imho.

alesj: So removing the dup code in code, but leave the script?

ppatierno: Yeah that was my suggestion

ppatierno:
Thanks to Julien from the Vert.x team, we have got a fix in the Vert.x Kafka client (https://github.com/vert-x3/vertx-kafka-client/pull/226) which I tested and seems to work. Of course, this will be available in the next Vert.x 4.3.3 release. An ETA seems to be August for having this release.
At this point we have two choices: updating this PR with the workaround of wrapping the Vert.x executor context while waiting for the Vert.x 4.3.3 release with a proper fix or ... waiting for the Vert.x 4.3.3 release and not merging this PR yet.
I have no strong opinion. Waiting for the Vert.x release could mean even another month, but I don't think we are going to have another bridge release soon.

@tombentley @scholzj what's your feeling on this?

pom.xml
```diff
@@ -498,6 +557,13 @@
 									<ignoredUnusedDeclaredDependency>org.apache.logging.log4j:log4j-slf4j-impl</ignoredUnusedDeclaredDependency>
 									<ignoredUnusedDeclaredDependency>io.jaegertracing:jaeger-client</ignoredUnusedDeclaredDependency>
 									<ignoredUnusedDeclaredDependency>org.yaml:snakeyaml</ignoredUnusedDeclaredDependency> <!-- CVE override -->
+									<ignoredUnusedDeclaredDependency>org.apache.tomcat.embed:tomcat-embed-core</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
+									<!-- OpenTelemetry - used via classpath configuration -->
+									<ignoredUnusedDeclaredDependency>io.opentelemetry:opentelemetry-sdk-trace</ignoredUnusedDeclaredDependency>
+									<ignoredUnusedDeclaredDependency>io.opentelemetry.instrumentation:opentelemetry-kafka-clients-common</ignoredUnusedDeclaredDependency>
+									<ignoredUnusedDeclaredDependency>io.opentelemetry.instrumentation:opentelemetry-instrumentation-api</ignoredUnusedDeclaredDependency>
+									<ignoredUnusedDeclaredDependency>io.opentelemetry:opentelemetry-exporter-jaeger</ignoredUnusedDeclaredDependency>
+									<ignoredUnusedDeclaredDependency>io.grpc:grpc-netty-shaded</ignoredUnusedDeclaredDependency>
```
ppatierno, line 566: you removed the grpc dependency but not this one

ppatierno (changes requested):

ppatierno:
Finally Vert.x 4.3.3 is out! @alesj could you update the PR by using this release and test again that everything works fine please?
I am ok with the PR but I will have another final pass. 
@scholzj @tombentley could you review please, it would be great to finally close this addition after so long time :-)

pom.xml
```diff
@@ -267,6 +270,56 @@
...
+			<version>${opentelemetry.alpha-version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.opentelemetry.instrumentation</groupId>
+			<artifactId>opentelemetry-instrumentation-api-semconv</artifactId>
+			<version>${opentelemetry.alpha-version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.opentelemetry.instrumentation</groupId>
+			<artifactId>opentelemetry-kafka-clients-common</artifactId>
```
ppatierno, line 310: To remove, it's transitive:

```shell
[INFO] +- io.opentelemetry.instrumentation:opentelemetry-kafka-clients-2.6:jar:1.16.0-alpha:compile
[INFO] |  +- io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:jar:1.16.0-alpha:compile
[INFO] |  \- io.opentelemetry.instrumentation:opentelemetry-kafka-clients-common:jar:1.16.0-alpha:runtime
```

pom.xml
```diff
@@ -498,6 +557,12 @@
 									<ignoredUnusedDeclaredDependency>org.apache.logging.log4j:log4j-slf4j-impl</ignoredUnusedDeclaredDependency>
 									<ignoredUnusedDeclaredDependency>io.jaegertracing:jaeger-client</ignoredUnusedDeclaredDependency>
 									<ignoredUnusedDeclaredDependency>org.yaml:snakeyaml</ignoredUnusedDeclaredDependency> <!-- CVE override -->
+									<ignoredUnusedDeclaredDependency>org.apache.tomcat.embed:tomcat-embed-core</ignoredUnusedDeclaredDependency>  <!-- CVE override -->
+									<!-- OpenTelemetry - used via classpath configuration -->
+									<ignoredUnusedDeclaredDependency>io.opentelemetry:opentelemetry-sdk-trace</ignoredUnusedDeclaredDependency>
+									<ignoredUnusedDeclaredDependency>io.opentelemetry.instrumentation:opentelemetry-kafka-clients-common</ignoredUnusedDeclaredDependency>
+									<ignoredUnusedDeclaredDependency>io.opentelemetry.instrumentation:opentelemetry-instrumentation-api</ignoredUnusedDeclaredDependency>
```
ppatierno, line 564: All the three above can be removed because they are not declared in the pom (so no warning about "unused declared")

ppatierno (changes requested):
@alesj I left some comments mostly around dependencies that can be safely removed

pom.xml
```diff
@@ -267,6 +270,56 @@
...
+			<version>${opentelemetry.alpha-version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.opentelemetry</groupId>
+			<artifactId>opentelemetry-api</artifactId>
+			<version>${opentelemetry.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.opentelemetry</groupId>
+			<artifactId>opentelemetry-context</artifactId>
```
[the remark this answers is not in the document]
alesj: [INFO] --- maven-dependency-plugin:3.1.1:analyze-only (analyze) @ kafka-bridge ---
[WARNING] Used undeclared dependencies found:
[WARNING]    io.opentelemetry:opentelemetry-context:jar:1.16.0:compile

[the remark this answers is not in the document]
ppatierno: I didn't have the warning but now noticed that the related classes are used. Don't know why I didn't get the error though :-(

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Override
+    public <K, V> SpanBuilderHandle<K, V> builder(RoutingContext routingContext, String operationName) {
+        SpanBuilder spanBuilder = getSpanBuilder(routingContext, operationName);
+        return new OTelSpanBuilderHandle<>(spanBuilder);
+    }
+
+    @Override
+    public <K, V> void handleRecordSpan(SpanHandle<K, V> parentSpanHandle, KafkaConsumerRecord<K, V> record) {
+        String operationName = String.format("%s %s", record.topic(), MessageOperation.RECEIVE);
```
tombentley, line 106: Can we change this to use string concat. `String.format` is surprisingly expensive, it's not providing much value here and it appears to be on a relatively hot path (the POLL request)

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Build span builder handle.
+     *
+     * @param <K> key type
+     * @param <V> value type
+     * @param routingContext Vert.x rounting context
+     * @param operationName current operation name
+     * @return span builder handle
+     */
+    <K, V> SpanBuilderHandle<K, V> builder(RoutingContext routingContext, String operationName);
```
tombentley, line 59: I don't understand what purpose this builder serves. It's used once in the `HttpSinkBridgeEndpoint` like this:

```java
            SpanBuilderHandle<K, V> builder = tracing.builder(routingContext, HttpOpenApiOperations.POLL.toString());
            SpanHandle<K, V> span = builder.span(routingContext);
```

So why can't the `builder` method of `TracingHandle` just return the `SpanHandle`? Basically push the existence of builders into the `TracingHandle` implementation as much as possible.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTracingHandle.java
```diff
@@ -0,0 +1,189 @@
...
+        return new OTSpanBuilderHandle<>(getSpanBuilder(routingContext, operationName));
+    }
+
+    @Override
+    public <K, V> SpanHandle<K, V> span(RoutingContext routingContext, String operationName) {
+        Tracer.SpanBuilder spanBuilder = getSpanBuilder(routingContext, operationName);
+        return buildSpan(spanBuilder, routingContext);
+    }
+
+    @SuppressWarnings("rawtypes")
```
tombentley, line 82: No need for this AFAICS. The cast can be written `((OTSpanHandle<?,?>) parentSpanHandle)`.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTracingHandle.java
```diff
@@ -0,0 +1,189 @@
...
+import static io.strimzi.kafka.bridge.tracing.TracingConstants.KAFKA_SERVICE;
+
+/**
+ * OpenTracing implementation of TracingHandle.
+ */
+class OpenTracingHandle implements TracingHandle {
+
+    private static Tracer tracer;
+
+    private static Tracer getTracer() {
```
tombentley, line 41: `GlobalTracer.get` is just returning "the constant GlobalTracer", so I don't think we need the static field and associated null check. Maybe we don't need this method either, unless we're trying to hide from the call sites of this method that it's the global trace which we're using.

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
```diff
@@ -0,0 +1,88 @@
...
+    void initialize();
+
+    /**
+     * Adapt executor service if needed.
+     * Else return service parameter instance.
+     *
+     * @param service current executor service
+     * @return adapted executor service or service parameter instance
+     */
+    default ExecutorService adapt(ExecutorService service) {
```
tombentley, line 46: This method appears to never be called.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -335,8 +307,7 @@ public Iterator<Map.Entry<String, String>> iterator() {
                         HttpUtils.sendResponse(routingContext, responseStatus.code(),
                             BridgeContentType.KAFKA_JSON, error.toJson().toBuffer());
                     }
-                    Tags.HTTP_STATUS.set(span, responseStatus.code());
-                    span.finish();
+                    span.finish(responseStatus.code());
```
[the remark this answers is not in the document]
alesj: OK, added `finally` and moved `finish` there.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -190,9 +155,7 @@ public Iterator<Map.Entry<String, String>> iterator() {
                     results.add(new HttpBridgeResult<>(new HttpBridgeError(code, msg)));
                 }
             }
-
-            Tags.HTTP_STATUS.set(span, HttpResponseStatus.OK.code());
-            span.finish();
+            span.finish(HttpResponseStatus.OK.code());
```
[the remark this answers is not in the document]
alesj: Which HttpResponseStatus should we return for failed `done` ?

src/main/java/io/strimzi/kafka/bridge/tracing/TracingHandle.java
alesj: It used to be used twice ... with re-impl of consume span handling it reduced to one.

alesj: OK, removed this mid-builder usage all together.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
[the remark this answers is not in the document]
alesj: Atm I used

```
            int code = done.succeeded() ? HttpResponseStatus.OK.code() : HttpResponseStatus.INTERNAL_SERVER_ERROR.code();
```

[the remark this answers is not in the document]
alesj: Although, I see `handleError` tries to be a bit smarter ... but that's per-result ... where this is for whole response.
Should we try to be smarter as well for a whole response?

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -169,7 +133,8 @@ public Iterator<Map.Entry<String, String>> iterator() {
         List<Future> sendHandlers = new ArrayList<>(records.size());
         for (KafkaProducerRecord<K, V> record : records) {
             Promise<RecordMetadata> promise = Promise.promise();
-            sendHandlers.add(promise.future());
+            Future<RecordMetadata> future = promise.future();
```
ppatierno, line 136: why this change about adding a `future` variable?

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.Properties;
+
+import static io.strimzi.kafka.bridge.tracing.TracingConstants.COMPONENT;
+import static io.strimzi.kafka.bridge.tracing.TracingConstants.KAFKA_SERVICE;
+import static io.strimzi.kafka.bridge.tracing.TracingConstants.OPENTELEMETRY_SERVICE_NAME_ENV_KEY;
+
+/**
+ * OpenTelemetry implementation of Tracing.
+ *
+ * Note: we use Vert.x OpenTelemetry extension to setup custom ContextStorageProvider:
```
ppatierno, line 40: we know that it's happening because the Vert.x OpenTelemetry library is on the classpath and the OpenTelemetry framework is able to detect and configure it automatically. Would you mind to add a comment here about how it works please? I am sure we'll forget :-)

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
alesj: I needed it before for that callback hack -- removed now.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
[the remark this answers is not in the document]
ppatierno: Thanks for reverting back the code! ;-)

ppatierno (approved):
LGTM! @alesj thanks for the effort!

tombentley (approved):
Thanks Ales!

PR #642, opened 2022-08-11
Updated CHANGELOG with OpenTelemetry support

ppatierno:


scholzj (commented):
Should we have the OpenTracing deprecation in a separate subchaper to make it more visible (see https://github.com/strimzi/strimzi-kafka-operator/blob/main/CHANGELOG.md#changes-deprecations-and-removals as example.)

scholzj (approved):

PR #649, opened 2022-08-23
docs(tracing): Add a procedure to set up tracing

PaulRMellor:
Signed-off-by: prmellor <pmellor@redhat.com>

*Documentation*

- Adds a new assembly for configuring the Kafka Bridge
- Adds a new procedure to describe how to set up distributed tracing for the Kafka Bridge
- Updates the quickstart -- removes the configuration procedure and describes default config of the `application.properties` file that are used in the quickstart


ShubhamRwt (approved):

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+[id='proc-configuring-kafka-bridge-tracing-{context}']
+= Configuring distributed tracing
+
+[role="_abstract"]
+Enable distributed tracing to trace messages consumed and produced by the Kafka Bridge, and HTTP requests from client applications.
+
+Properties to enable tracing are present in the `application.properties` file.
+To enable distributed tracing, do the following:
+
+* Uncomment the required tracing properties in the properties file
```
joaedwar, line 14:
```suggestion
* Uncomment the required tracing properties in the properties file.
```

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+* Set environment variables for tracing
```
joaedwar, line 15:
```suggestion
* Set environment variables for tracing.
```

joaedwar (approved):
a couple minor suggestions.

lgtm

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+. Save the configuration file.
+. Set the environment variables for tracing.
++
+.Environment variables for OpenTelemetry 
+[source,env]
+----
+OTEL_SERVICE_NAME=my-jaeger-service # <1>
+----
+<1> The name of the OpenTelemetry Jaeger tracer service.
```
ppatierno, line 63: It doesn't have any relation with Jaeger. Let's remove it from the text and the env var value.

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+If you want to use another tracing system with OpenTelemetry, do the following: 
+
+. Add the library of the tracing system to the Kafka classpath.
+. Add the name of the tracing system as an additional environment variable.
++
+.Additional environment variable when not using Jaeger 
+[source,env]
+----
+OTEL_SERVICE_NAME=my-jaeger-service
```
ppatierno, line 98: Again no Jaeger here

ppatierno (changes requested):

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+[id='proc-configuring-kafka-bridge-tracing-{context}']
+= Configuring distributed tracing
+
+[role="_abstract"]
+Enable distributed tracing to trace messages consumed and produced by the Kafka Bridge, and HTTP requests from client applications.
+
+Properties to enable tracing are present in the `application.properties` file.
+To enable distributed tracing, do the following:
+
+* Uncomment the required tracing properties in the properties file.
```
ppatierno, line 14: I would not say in that way.
I would more focusing on the `bridge.tracing` property that should be set in the file.
Something like "set the bridge.tracing property value enabling the tracing you want to use. Possible values are jaeger and opentelemetry".
The fact that we are providing the `application.properties` file with commented lines is for simplicity but I would not use it for explaining how it works.
Wdyt?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+* Set environment variables for tracing.
+
+With the default configuration, you can introduce a Jaeger tracing system based on OpenTelemetry or OpenTracing.
```
ppatierno, line 17: wdym by "you can introduce" ?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+OpenTelemetry and OpenTracing are API specifications for collecting tracing data as _spans_ of metrics data.
+Spans represent a specific operation.
+A trace is a collection of one or more spans.
+
+Traces are generated when:
+
+* The Kafka Bridge sends messages to HTTP clients and consumes messages from HTTP clients
+* HTTP clients send HTTP requests to send and receive messages through the Kafka Bridge
```
ppatierno, line 26: aren't these two sentence saying the same but from two different point of view? bridge and clients? or did you mean something different?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+Jaeger implements the required APIs and presents visualizations of the trace data in its user interface for analysis. 
+
+To have end-to-end tracing, you must configure tracing in your HTTP clients.
+
+NOTE: The OpenTracing project is now archived, so Strimzi has deprecated support for OpenTracing.
```
ppatierno, line 32: Isn't it better using a warning than a note?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+.Prerequisites
+
+* xref:proc-downloading-kafka-bridge-{context}[The Kafka Bridge installation archive is downloaded].
+
+.Procedure
+
+. Edit the `application.properties` file provided with the Kafka Bridge installation archive.
++
+Use the properties file to uncomment the relevant tracing property.
```
ppatierno, line 42: it's kind of the same point I made before. I would talk about setting the `bridge.tracing` property with the right value instead of uncommenting stuff.

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
++
+.Example configuration to enable OpeTelemetry
+[source,properties]
+----
+#bridge.tracing=jaeger # <1>
+bridge.tracing=opentelemetry # <2>
+----
+<1> The property for enabling OpenTracing (deprecated). Here left commented.
+<2> The property for enabling OpenTelemetry is uncommented.
```
ppatierno, line 51: Referred to my previous comments, we should just have the property set and not the commented one?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,103 @@
...
+The internal consumers and producers of the Kafka Bridge are now enabled for tracing.
+
+== Specifying tracing systems with OpenTelemetry
+
+Instead of the default Jaeger system, you can specify other tracing systems that are supported by OpenTelemetry.
+
+If you want to use another tracing system with OpenTelemetry, do the following: 
+
+. Add the library of the tracing system to the Kafka classpath.
+. Add the name of the tracing system as an additional environment variable.
```
ppatierno, line 93: Maybe having a link to the official OTel documentation about possible value for the `OTEL_TRACES_EXPORTER` env var could be useful. Here it is: https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#exporters

documentation/modules/proc-installing-kafka-bridge.adoc
```diff
@@ -6,12 +6,22 @@
 = Installing the Kafka Bridge
 
 [role="_abstract"]
-Follow this procedure to install the Strimzi Kafka Bridge.
+Use the script provided with the Kafka Bridge archive to install the Strimzi Kafka Bridge.
```
ppatierno, line 9: I see we removed "Strimzi" in other places.

ppatierno (changes requested):

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
PaulRMellor: Works for me. I was trying to keep things very simple in the intro. But I think this is better

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
PaulRMellor: Changed to: _implement a Jaeger tracing system_

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
PaulRMellor: Maybe CAUTION?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
PaulRMellor: I think it's useful to show the two.

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
ppatierno: As not a native speaker, it doesn't sound great to me as well. Of course, it could be just me.
I guess the concept you are trying to explain is something like: "with the default configuration, by using OpenTelemetry or OpenTracing, you will get tracing based on Jaeger implementation underneath". Does "you can introduce a Jaeger ..." or "implement a Jaeger ..." map to the above sentence? If yes ... then it's me not getting it :-) ... even if I would elaborate more.

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
ppatierno: This is still what I see ...

`HTTP clients send HTTP requests to the Kafka Bridge` IS EQUAL TO `The Kafka Bridge consumes messages from HTTP clients`
`HTTP clients receive messages from the Kafka Bridge` IS EQUAL TO `The Kafka Bridge sends messages to HTTP clients`

But I think what you mean. I guess you are referring to the fact that if tracing is enabled on the HTTP client, and for example you are sending to the bridge, you will get: a span from the HTTP client (about its send), a span in the bridge about sending to Kafka, then finally a span from the Kafka instrumented library about message sent to Kafka. The same could apply for consuming in the other way around.
Anyway the spans coming from an HTTP client are not related to the tracing enabled on the bridge, but the tracing enabled on the client itself.
Or maybe you want to say something different which I am not getting right :-D

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
ppatierno: Fine with me. Just something more than a NOTE.

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,106 @@
...
+== Specifying tracing systems with OpenTelemetry
+
+Instead of the default Jaeger system, you can specify other tracing systems that are supported by OpenTelemetry.
+
+If you want to use another tracing system with OpenTelemetry, do the following: 
+
+. Add the library of the tracing system to the Kafka classpath.
+. Add the name of the tracing system as an additional exporter environment variable.
++
+.Additional exporter environment variable when not using Jaeger
```
ppatierno, line 95: Actually they are not env var for the exporter only. They are used by the overall OpenTelemetry sdk. I think it was better before.

ppatierno (changes requested):

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
PaulRMellor: Let's just make it simpler. :-)

> Traces are generated when the Kafka Bridge does the following:
> 
> * Sends messages to HTTP clients
> * Consumes messages from HTTP clients

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -0,0 +1,106 @@
...
+With the default configuration, you can implement a Jaeger tracing system based on OpenTelemetry or OpenTracing.
+
+OpenTelemetry and OpenTracing are API specifications for collecting tracing data as _spans_ of metrics data.
+Spans represent a specific operation.
+A trace is a collection of one or more spans.
+
+Traces are generated when the Kafka Bridge does the following:
+
+* Sends messages to HTTP clients
+* Consumes messages from HTTP clients
```
ppatierno, line 26: Still not sure .... what if something like this?

"Sends message back from Apache Kafka to HTTP clients when they are consuming"
"Get messages from HTTP clients to send to Apache Kafka when they are producing"

PaulRMellor: Trying to make it more succinct:

* Sends messages from Kafka to consumer HTTP clients 
* Receives messages from producer HTTP clients to send to Kafka

ppatierno (approved):
LGTM

Issue #653, opened 2022-09-05
Tracing failed `poll` operation on HTTP consume request


ppatierno:
Currently, the tracing span for a `poll` operation is created only when the actual Kafka poll ends successfully:
This happens in the `HttpSInkBridgeEndpoint.pollHandler` method here:

https://github.com/strimzi/strimzi-kafka-bridge/blob/main/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java#L249

Kafka poll success means `records.succeeded()` is true.
It means that there is no way to trace a failed Kafka poll, while OpenTelemetry allows to set an ERROR status on a span.
I think we should move the `poll` span creation earlier to be used even when the Kafka poll operation fails.

ShubhamRwt:
Closing this since #655 is now merged

PR #655, opened 2022-09-09
Moved creation of span happen before the poll is succesful

ShubhamRwt:

This PR fixes issue mentioned in #653. Moved the creation of the span happen before the poll is succesful so that it can also be used in the case if the poll is unsuccessful

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,8 +79,8 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<maven.compiler.source>1.9</maven.compiler.source>
-		<maven.compiler.target>1.9</maven.compiler.target>
+		<maven.compiler.source>1.8</maven.compiler.source>
+		<maven.compiler.target>1.8</maven.compiler.target>
 		<log4j.version>2.17.2</log4j.version>
 		<slf4j.version>1.7.21</slf4j.version>
 		<vertx.version>4.3.3</vertx.version>
--- a/pom.xml
+++ b/pom.xml
@@ -622,6 +622,14 @@
 					<excludeFilterFile>${project.basedir}/.spotbugs/spotbugs-exclude.xml</excludeFilterFile>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<source>9</source>
+					<target>9</target>
+				</configuration>
+			</plugin>
 		</plugins>
 	</build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -79,8 +79,8 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<maven.compiler.source>1.8</maven.compiler.source>
-		<maven.compiler.target>1.8</maven.compiler.target>
+		<maven.compiler.source>1.9</maven.compiler.source>
+		<maven.compiler.target>1.9</maven.compiler.target>
 		<log4j.version>2.17.2</log4j.version>
 		<slf4j.version>1.7.21</slf4j.version>
 		<vertx.version>4.3.3</vertx.version>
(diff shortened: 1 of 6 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -242,11 +242,14 @@ private void doDeleteConsumer(RoutingContext routingContext) {
     }
 
     private Handler<AsyncResult<KafkaConsumerRecords<K, V>>> pollHandler(RoutingContext routingContext) {
+
         return records -> {
-            if (records.succeeded()) {
 
-                TracingHandle tracing = TracingUtil.getTracing();
-                SpanHandle<K, V> span = tracing.span(routingContext, HttpOpenApiOperations.POLL.toString());
+            TracingHandle tracing = TracingUtil.getTracing();
```
ShubhamRwt, line 248: @ppatierno Hi, My first doubt was regarding this... we want the creation of span inside the `pollHandler` method itself, right?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -292,6 +295,7 @@ private Handler<AsyncResult<KafkaConsumerRecords<K, V>>> pollHandler(RoutingCont
                 );
                 HttpUtils.sendResponse(routingContext, HttpResponseStatus.INTERNAL_SERVER_ERROR.code(),
                     BridgeContentType.KAFKA_JSON, error.toJson().toBuffer());
+                span.finish(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
```
ShubhamRwt, line 298: My second doubt was that we have done abstraction over the span implementations right, so I used the finish method to set the error code but do we need to inject the RoutingContext into the span also as we are doing in case of successful poll?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: Yes because anyway the `span` instance is used just inside it. It's useless to create outside and pass that in.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: I don't think we need because anyway the span is going to be finished.

ppatierno (commented):
I left a couple of comments.
Have you tested this trying to reproduce an error so that you can see the tracing with the error in a Jaeger UI for example?

ShubhamRwt:
> I left a couple of comments. Have you tested this trying to reproduce an error so that you can see the tracing with the error in a Jaeger UI for example?

No, I haven't. But I will do it now to see if the changes are working fine

alesj:
OK, looks like an easy fix, and already resolved.
I think I tried to mock the previous code as much as possible, not moving it around, just adding abstraction layer on top.

alesj (approved):

pom.xml
```diff
@@ -622,6 +622,14 @@
 					<excludeFilterFile>${project.basedir}/.spotbugs/spotbugs-exclude.xml</excludeFilterFile>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
```
ShubhamRwt, line 626: @ppatierno I had to add this plugin to make use `Map.of()`. It is causing the build to fail. So should I remove the use `Map.of()` and use a normal Hashmap();

ppatierno: @ShubhamRwt this is because `Map.of` was introduced in Java 9, so it's not about using the plugin but about the Java version. It should work even not using the plugin but updating:

```xml
<maven.compiler.source>1.9</maven.compiler.source>
<maven.compiler.target>1.9</maven.compiler.target>
```

can you try that?

Said that we should move to use the plugin and Java 11 as for the operator but this is for a different PR.

src/main/java/io/strimzi/kafka/bridge/tracing/SpanHandle.java
```diff
@@ -32,4 +32,13 @@
      * @param code response code
      */
     void finish(int code);
+
+    /**
+     * Finish underlying span.
+     *
+     * @param code response code
+     *
```
ppatierno, line 40: delete blank line

pom.xml
ShubhamRwt: Added the change

scholzj: You cannot change to Java 9 when we support Java 8+.

ppatierno: @scholzj wdym by "we support Java 8+"? Shouldn't we also move to Java 11 as the operator?

scholzj: I think that is a discussion beyond this PR. The Bridge today supports and tests Java 8 and 11. You cannot change it to Java 9 and 11 just like that within a PR which does some tracing span modification (plus it is anyway way more complicated as you can see form the failing CI).

ppatierno (approved):

scholzj (approved):

PR #659, opened 2022-09-16
Extract consumer client.rack from /opt/strimzi/init/rack.id if present

robobario:
Why:
To support rack awareness the strimzi kafka operator will mount in a file containing the rack.id, this can be appended to the consumer configuration if it is present

This has a complementaty PR in strimzi-kafka-operator to mount in the file and populate it via an init container https://github.com/strimzi/strimzi-kafka-operator/pull/7339


scholzj (approved):
LGTM. Thanks.

ppatierno (commented):
We should take into account that the bridge can be used even without the Strimzi operator so on installation where the Kafka cluster is deployed "by hand".
How is it going to work in this case if the user wants rack-awareness? I see this change is just related to the docker container configuration.

scholzj:
@ppatierno When users runs it without the operator, they have to configure the bridge them self. So they can set the `kafka.consumer.client.rack` to whatever they want directly in the config file. This change needs to be done only for the operator where we set it automatically.

ppatierno (approved):
LGTM

PR #660, opened 2022-09-19
Added Jaeger endpoint on OpenTelemetry related documentation

ppatierno:
The OpenTelemetry documentation was missing the env var related to the endpoint for reaching the tracing system which collects the spans.
This PR adds the env var specific for OpenTelemetry when using Jaeger as default and the corresponding env var if the user wants to use a different tracing system.


PaulRMellor (approved):
Looks good to me

PR #665, opened 2022-09-30
Bumped OpenTelemetry to 1.18.0

ppatierno:


pom.xml
```diff
@@ -110,8 +110,8 @@
 		<jaeger.version>1.8.1</jaeger.version>
 		<opentracing.version>0.33.0</opentracing.version>
 		<opentracing-kafka-client.version>0.1.15</opentracing-kafka-client.version>
-		<opentelemetry.alpha-version>1.16.0-alpha</opentelemetry.alpha-version>
-		<opentelemetry.version>1.16.0</opentelemetry.version>
+		<opentelemetry.alpha-version>1.18.0-alpha</opentelemetry.alpha-version>
+		<opentelemetry.version>1.18.0</opentelemetry.version>
```
Frawless, line 114: just out of curiosity, what does the alpha version contains, and regular one not? I mean why do we use alpha with kinda same version?

scholzj (approved):

pom.xml
ppatierno: It's just that opentelemetry has different components, some are already stable, others are in "alpha" since months but used everywhere :-)

see-quick (approved):

im-konge (approved):

PR #669, opened 2022-10-12
Install asciidoctor locally

srajmane:
This PR is to resolve gem install issue on s390x Travis build.

Signed-off-by: Snehal Rajmane srajmane@us.ibm.com


ppatierno:
@scholzj the commit seems to be signed but the DCO is red. Do you have any idea about it?

scholzj:
I think the format of the signature is wrong. It should be `Signed-off-by: Author Name <authoremail@example.com>` => the `<` and `>` seem to be missing. Maybe that is the problem here?

srajmane:
> I think the format of the signature is wrong. It should be `Signed-off-by: Author Name <authoremail@example.com>` => the `<` and `>` seem to be missing. Maybe that is the problem here?

@scholzj Thank you, update the signature.

ppatierno:
Thanks Jakub! I didn't notice that :-(

scholzj (approved):
LGTM assuming the tests pass.

ppatierno (approved):
LGTM. Thanks for fixing this! The tests and Travis build passed!

PR #686, opened 2022-11-08
Added doc about Ryuk in privileged mode when running IT tests

ppatierno:
I was having hard times to run IT tests on my laptop.
It turned out that it was Ryuk (used by test-containers for cleanup) not able to start in privileged mode as it should be.
This PR adds a little bit of documentation about how you can run Ryuk in privileged mode or disable it.

### DIFF
--- a/development-docs/TESTING.md
+++ b/development-docs/TESTING.md
@@ -25,6 +25,7 @@ In case of interest you can check our implementation of Strimzi `Kafka` containe
 > The test-containers project uses [Moby Ryuk](https://github.com/testcontainers/moby-ryuk) to automatic cleanup of containers after the execution. Ryuk must be started as a privileged container.
 Running in privileged mode could not apply for all environments by default. If you get any error running Ryuk, you can set `TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED=true`
 or disable Ryuk by setting `TESTCONTAINERS_RYUK_DISABLED=true`.
+One of the potential issues is about having `java.lang.IllegalArgumentException: Requested port (8080) is not mapped`.
 
 ## Package Structure
 
--- a/development-docs/TESTING.md
+++ b/development-docs/TESTING.md
@@ -22,9 +22,9 @@ in root directory of the project.
 The next requirement is to have [docker](https://docs.docker.com/get-docker/) installed, because we use [test-containers](https://www.testcontainers.org/). 
 In case of interest you can check our implementation of Strimzi `Kafka` container [here](https://github.com/strimzi/strimzi-kafka-operator/tree/main/test-container).
 
-> The test-containers project uses [Moby Ryuk](https://github.com/testcontainers/moby-ryuk) to automatic cleanup of containers after the execution. Ryuk must be started as a privileged container.
-Running in privileged mode could not apply for all environments by default.  If you get an error like this `java.lang.IllegalArgumentException: Requested port (8080) is not mapped` 
-when test-containers tries to start, then you can set `TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED=true` or disable Ryuk by setting `TESTCONTAINERS_RYUK_DISABLED=true`. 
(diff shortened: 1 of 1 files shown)


scholzj (commented):
Should you maybe show some examples of the errors to which it applies? As mentioned before, there can be different issues such as the DockerHub limits etc. So it would be good to make it clear when does it apply instead of providing it as a advice for everything.

ppatierno:
I can but it will be just my error, I don't know about others.

scholzj:
> I can but it will be just my error, I don't know about others.

**Maybe** there are other errors which it solves which we haven't seen yet. But we also know for **sure** that there are other errors which the privileged flag will not solve.

ppatierno:
so I don't get what you are suggesting and what I should add? Tbh having this hints would have helped be avoiding to waste 1 hour. It's at least a try that user can try if they got an error which could be related to Ryuk.

scholzj:
You had a problem and solved it. It is fine to document the solution, it can save someone some time. But you seem to expect that everyone has the same issue with the same solution. Which is not the case. So I think it should be put like this: _If you see an error like `this`, then maybe you should try to do `this`._ ...  Instead, you seem to put it as _If you have any problem, do `this` because it helped me._

ppatierno:
Ok. Does it sound any better?

scholzj (approved):

development-docs/TESTING.md
```diff
@@ -22,6 +22,10 @@ in root directory of the project.
 The next requirement is to have [docker](https://docs.docker.com/get-docker/) installed, because we use [test-containers](https://www.testcontainers.org/). 
 In case of interest you can check our implementation of Strimzi `Kafka` container [here](https://github.com/strimzi/strimzi-kafka-operator/tree/main/test-container).
 
+> The test-containers project uses [Moby Ryuk](https://github.com/testcontainers/moby-ryuk) to automatic cleanup of containers after the execution. Ryuk must be started as a privileged container.
+Running in privileged mode could not apply for all environments by default.  If you get an error like this `java.lang.IllegalArgumentException: Requested port (8080) is not mapped`
```
PaulRMellor, line 26:
```suggestion
Running in privileged mode might not apply for all environments by default.  If you get an error like `java.lang.IllegalArgumentException: Requested port (8080) is not mapped` 
```

development-docs/TESTING.md
```diff
@@ -22,6 +22,10 @@ in root directory of the project.
 The next requirement is to have [docker](https://docs.docker.com/get-docker/) installed, because we use [test-containers](https://www.testcontainers.org/). 
 In case of interest you can check our implementation of Strimzi `Kafka` container [here](https://github.com/strimzi/strimzi-kafka-operator/tree/main/test-container).
 
+> The test-containers project uses [Moby Ryuk](https://github.com/testcontainers/moby-ryuk) to automatic cleanup of containers after the execution. Ryuk must be started as a privileged container.
```
PaulRMellor, line 25:
```suggestion
> The `test-containers` project uses [Moby Ryuk](https://github.com/testcontainers/moby-ryuk) to automatic cleanup of containers after the execution. Ryuk must be started as a privileged container.
```

development-docs/TESTING.md
```diff
@@ -22,6 +22,10 @@ in root directory of the project.
...
+Running in privileged mode could not apply for all environments by default.  If you get an error like this `java.lang.IllegalArgumentException: Requested port (8080) is not mapped` 
+when test-containers tries to start, then you can set `TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED=true` or disable Ryuk by setting `TESTCONTAINERS_RYUK_DISABLED=true`.
```
PaulRMellor, line 27:
```suggestion
when `test-containers` tries to start, you can set `TESTCONTAINERS_RYUK_CONTAINER_PRIVILEGED=true` or disable Ryuk by setting `TESTCONTAINERS_RYUK_DISABLED=true`. 
```

PaulRMellor (approved):
:+1:

PR #687, opened 2022-11-08
Moving from Jaeger exporter to OTLP exporter usage by default

### LINKED ISSUE #685: Migrating from OpenTelemetry Jaeger exporter to OTLP exporter
As for the operator with https://github.com/strimzi/strimzi-kafka-operator/issues/7592, we should migrate from OpenTelemetry Jaeger exporter to the OTLP exporter.

Referring to https://github.com/strimzi/strimzi-kafka-operator/issues/7570, the current Jaeger Exporter, used by default when OpenTelemetry is enabled, doesn't have support for TLS via SDK environment variables.
At same time, the Jaeger Exporter is going to be deprecated (details https://github.com/open-telemetry/opentelemetry-specification/pull/2858) and the OpenTelemetry community is pushing for using the OTLP exporter which has TLS support and it's anyway usable with a Jaeger backend which has [support for OTLP protocol](https://www.jaegertracing.io/docs/1.35/apis/#opentelemetry-protocol-stable) since 1.35 version.
We should migrate to OTLP Exporter as soon as possible, which means:

* removing the Jaeger Exporter dependency `opentelemetry-exporter-jaeger` and replacing it with the `opentelemetry-exporter-otlp`
* making the new OTLP Exporter the default one by setting the `OTEL_TRACES_EXPORTER=otlp` in our images script or just leaving that env var empty because it will be the default behavior for OpenTelemetry

The migration should be done before the next release, so it also means that the #684 can be closed and documentation doesn't need any statement around TLS.

ppatierno:
This PR fixes #685 moving from the Jaeger exporter to the OLTP exporter usage by default when using OpenTelemetry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 0.23.0
+
+* Moved from using the Jaeger exporter to OTLP exporter by default
+
 ## 0.22.3
 
 * Strimzi OAuth updated to 0.11.0 bringing support for password grants and OAuth metrics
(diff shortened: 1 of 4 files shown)


im-konge (approved):

scholzj (approved):
LGTM.

PR #689, opened 2022-11-14
Updated Prometheus client and JMX collector to align with newer Micrometer version

ppatierno:
This PR fixes 688 by bumping Prometheus client and JMX collector versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,8 +113,8 @@
 		<opentelemetry.alpha-version>1.18.0-alpha</opentelemetry.alpha-version>
 		<opentelemetry.version>1.18.0</opentelemetry.version>
 		<micrometer.version>1.9.5</micrometer.version>
-		<jmx-prometheus-collector.version>0.12.0</jmx-prometheus-collector.version>
-		<prometheus-simpleclient.version>0.7.0</prometheus-simpleclient.version>
+		<jmx-prometheus-collector.version>0.17.2</jmx-prometheus-collector.version>
+		<prometheus-simpleclient.version>0.16.0</prometheus-simpleclient.version>
 		<commons-cli.version>1.4</commons-cli.version>
 		<test-container.version>0.102.0</test-container.version>
 		<jakarta.version>2.3.2</jakarta.version>


pom.xml
```diff
@@ -113,8 +113,8 @@
 		<opentelemetry.alpha-version>1.18.0-alpha</opentelemetry.alpha-version>
 		<opentelemetry.version>1.18.0</opentelemetry.version>
 		<micrometer.version>1.9.5</micrometer.version>
-		<jmx-prometheus-collector.version>0.12.0</jmx-prometheus-collector.version>
-		<prometheus-simpleclient.version>0.7.0</prometheus-simpleclient.version>
+		<jmx-prometheus-collector.version>0.17.2</jmx-prometheus-collector.version>
+		<prometheus-simpleclient.version>0.16.0</prometheus-simpleclient.version>
```
scholzj, line 117: Is there a reason why are these configured separately instead of just being used as transitive dependencies?

scholzj (approved):
LGTM. But please check the comment.

im-konge (approved):

pom.xml
ppatierno: @scholzj  I was wondering the same so I took a deeper look and:

* we use the JMX Prometheus collector for scraping the local Kafka producer/consumer metrics via JMX and expose them in Prometheus
* the Prometheus simpleclient seems not needed to us directly but it's a Micrometer and JMX Prometheus collector dependency. It could be removed but then we have something like this:

```shell
[INFO] +- io.micrometer:micrometer-registry-prometheus:jar:1.9.5:compile
[INFO] |  \- io.prometheus:simpleclient_common:jar:0.15.0:compile
[INFO] +- io.prometheus.jmx:collector:jar:0.17.2:compile
[INFO] |  \- io.prometheus:simpleclient:jar:0.16.0:compile
[INFO] |     +- io.prometheus:simpleclient_tracer_otel:jar:0.16.0:compile
[INFO] |     |  \- io.prometheus:simpleclient_tracer_common:jar:0.16.0:compile
[INFO] |     \- io.prometheus:simpleclient_tracer_otel_agent:jar:0.16.0:compile
```

The JMX collector is bringing a different version of the simpleclient compared to Micrometer. I think we decided to add direct dependency to avoid this situation.

PR #701, opened 2022-11-19
Fixed warning on unchecked cast on empty list

ppatierno:
Trivial PR to fix an IDE warning about an unchecked cast missing because `EMPTY_LIST` returns just a `List` while it's better using the `emptyList()` method which returns the proper generic expected List type.


scholzj (approved):
Fine with me. But isn't it better to use `List.of()`?

ppatierno:
I took a look and yes, the immutable collection returner by the `List.of()` sounds fine because the headers are not touched inside the tests. I will change it.

scholzj (approved):

see-quick (approved):

PR #709, opened 2022-11-22
Remove setting the handlers beforehand

ppatierno:
A lot of logic related to the AMQP 1.0 support was setting the handlers beforehand to be used asynchronously later (mostly for push-based mechanism).
They are not needed anymore. When an HTTP call comes to the bridge we need to use the underneath Kafka consumer and reply back to the HTTP client when the Kafka call is completed in a more "sync" (from an HTTP point of view) way. 
In this case, passing the handler on the call is enough.
This PR makes the change for subscribe, unsubscribe and assign operations.


scholzj (approved):
LGTM I guess ...

PR #712, opened 2022-11-23
Bumped SnakeYAML 1.33 to fix CVE

ppatierno:
This PR updates SnakeYAML to 1.33 to address the dependabot reports.


scholzj (approved):

scholzj:
TBH, I'm not sure how I feel about the CHANGELOG entry. If the update and CVE is worth changelog entry ... is it also worth a patch release with the CVE fix? ;-)

Anyway, I approved it.

ppatierno:
I added that line because we were doing that in the past. Regarding the patch, I would try to wait for 0.23.0 ... we are pretty close to it :-)

PR #679, opened 2022-10-31
added params to disable/enable consumer/producer

asagarshinde:
Signed-off-by: sagar <sagar.s@karza.in>

PR #713, opened 2022-11-24
Add tests about disabling producer or consumer

ppatierno:
This PR adds missing tests related to disable producer or consumer on the bridge (pushed via #679)

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/DisablingConsumerProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/DisablingConsumerProducerIT.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.strimzi.kafka.bridge.http;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.strimzi.kafka.bridge.BridgeContentType;
+import io.strimzi.kafka.bridge.MetricsReporter;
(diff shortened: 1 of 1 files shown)


im-konge (approved):

scholzj (approved):
Great, thanks.

src/test/java/io/strimzi/kafka/bridge/http/DisablingConsumerProducerIT.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @BeforeEach
+    void setup(VertxTestContext context) {
+        vertx = Vertx.vertx();
+        client = WebClient.create(vertx, new WebClientOptions()
+                .setDefaultHost(Urls.BRIDGE_HOST)
+                .setDefaultPort(Urls.BRIDGE_PORT)
+        );
+        context.completeNow();
```
Frawless, line 81: I am not vertx expert, but are you sure this won't make all tests pass immediately even if the test fails? I remember we hit some confusing behavior in the operator repo in the past because of this.

Also wouldn't be better to use pure Java for new things?

ppatierno: That was my understanding but it's not the case, if you see all the other tests in the bridge.

> Also wouldn't be better to use pure Java for new things?

We are going to move out from Vert.x so the tests will be refactored again. I would leave these ones as Vert.x for now, then there will be an entire work to rewrite them without the Vert.x support.

Frawless (approved):

see-quick (approved):

ppatierno:
Going to merge, because Travis CI seems to have problem looping the job but not starting it.

PR #716, opened 2022-11-25
Update Config Providers to 1.1.0

scholzj:
This PR updates the Kubernetes Config Provider and Env Var Config Provider to their new release 1.1.0 which brings Java 17 support and dependency updates.


ppatierno (approved):
LGTM. In one of the previous release we updated the changelog with this dependencies bump. Do you think it's not needed or just forgot for this PR?

scholzj:
@ppatierno TBH, there is not much new in there what would be worth mentioning, so it was intentionally. Its mainly the Java 17 stuff why we did the new release. But Java 17 support will be covered as part of the bridge itself 🤷

PR #718, opened 2022-11-28
Add oauth metrics into the `jmx_metrics_config.yaml`

im-konge:

In Oauth 0.11.0, the oauth metrics were introduced.
But, to scrape these Oauth metrics from Bridge, we need to add the correct patterns into the `jmx_metrics_config.yaml`.

This PR does it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fixed bug about missing OAuth password grants configuration   
 * Dependency updates (Vert.x 4.3.5)
 * Dependency updates (snakeYAML [CVE-2022-41854](https://nvd.nist.gov/vuln/detail/CVE-2022-41854))
+* Add Oauth metrics into the `jmx_metrics_config.yaml`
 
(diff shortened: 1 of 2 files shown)


scholzj (commented):
What happens when OAuth is not used?

im-konge:
@scholzj good question, I will try it

im-konge:
@scholzj I tested this with just enabled metrics, without Oauth, and it just prints metrics, that are available. There are no Oauth metrics at all (there is no NaN or nulls etc.)

ppatierno (approved):
LGTM. I don't know all the OAuth metrics, so I would assume all these rules for them are necessary and we have them in the operator as well I guess?

im-konge:
Those metrics, which I'm adding to the Bridge, I found in Oauth repository from @mstruk , so I assumed that's all needed.
We don't have it in operators repository, it should be introduced in [Strimzi#7694](https://github.com/strimzi/strimzi-kafka-operator/pull/7694)

PR #727, opened 2022-12-12
Fixed body handler missing when CORS is enabled

### LINKED ISSUE #726: Sending messages with CORS enabled raises a `400 Bad request` with `Null body` error
This issue is referred to discussion https://github.com/strimzi/strimzi-kafka-bridge/discussions/696

When CORS is enabled this way ...

```shell
#Enable CORS
http.cors.enabled=true
http.cors.allowedOrigins=*
http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH
```

and we send an HTTP request for producing messages, we get an error:

```shell
curl -X POST http://localhost:8080/topics/my-topic -H 'content-type: application/vnd.kafka.json.v2+json' -d '{"records": [{"key": "key-1","value": "value-1"},{"key": "key-2","value": "value-2"}]}'
{"error_code":400,"message":"[Bad Request] Body application/vnd.kafka.json.v2+json parsing error: Null body"}
```

Disabling CORS makes it working fine instead.

ppatierno:
This PR fixes #726 by adding the `BodyHandler` explicitly when CORS is enabled (so when the `CorsHandler` is added as root handler in the builder).
It also adds a test about producing messages with CORS enabled. It was missing and we didn't catch the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## 0.24.0
 
-* TBD
+* Fixed regression about producing messages not working when CORS is enabled
+* Dependency updates (Netty 4.1.86.Final)
 
 ### Changes, deprecations and removals
 
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -31,6 +31,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 
+import io.vertx.ext.web.handler.BodyHandler;
 import io.vertx.ext.web.handler.CorsHandler;
 import io.vertx.ext.web.openapi.RouterBuilder;
 import io.vertx.ext.web.validation.BodyProcessorException;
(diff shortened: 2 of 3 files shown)


ppatierno:
I preferred to open this PR as draft while waiting for Vert.x authors answering my question here https://github.com/vert-x3/vertx-web/issues/2320 because right now I don't see any better way to fix it in the bridge.

pmlopes:
Hi @ppatierno you're right here. In 4.3.0 we introduced some extra validation to ensure that routers are not constructed wrongly and this forced some behavior change in the route builder.

By default, the route builder would allow specifying a `BodyHandler` or any other handler, the issue was that the order does matter. For example, if we assume the list:

```
[
  BodyHandler,
  CorsHandler
]
```

This means that a request that would fail the CORS checks would still be parsed and inflated to memory, while this isn't a big problem you can imagine more odd setups could lead to bad behavior.

So, your fix is correct, what makes this not so obvious (and I'm also not happy with it) is that we tried to keep the old behavior to always have a `BodyHandler` by default so for the "general" case is just works, but for the more complex scenarios, given that we don't know which handlers will be added, we can't automatically add the body handler, and it's up to the developer to add it at the right time.

Sorry for the inconvenience!

scholzj (approved):

im-konge (approved):

PR #734, opened 2022-12-20
Replace Vert.x Kafka client with native Java Kafka client in the sink endpoint

ppatierno:
This PR is about replacing the usage of the Vert.x Kafka client on the sink endpoint (consumer side of the bridge) with the native Java Kafka client (provided by Apache Kafka upstream project).

It provides the following things:

* using `CompletableFuture`(s) based pattern to run consumer requests asynchronously (as today it's done by Vert.x)
* removing the Vert.x instance from the sink/consumer side
* replacing Vert.x Kafka related classes, i.e. KafkaConsumer, KafkaConsumerRecord, and more with the corresponding native ones from the Java Kafka client

As additional note, I decided to start the async operations (via `CompletableFuture.runAsync` and similar) in the higher level class `HttpSinkBridgeEndpoint` because, as already raised by Tom in the producer side, we could end at some point to remove the `SinkBridgeEndpoint` class, having just one. This is also the same behaviour as the producer/source endpoint.

Anyway, there are still some Vert.x structures in place like the ones handling JSON, Buffer(s) and Handler(s).


src/main/java/io/strimzi/kafka/bridge/NoopPartitionsRebalance.java
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+
+/**
+ * No operations implementation about handling partitions being assigned on revoked on rebalancing
+ * It just logs partitions if enabled
+ */
+public class NoopPartitionsRebalance implements ConsumerRebalanceListener {
```
tombentley, line 19: Should we call it `LoggingPartitionsRebalance`, since that's what it does?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -186,25 +178,24 @@ private void doSeekTo(RoutingContext routingContext, JsonObject bodyAsJson, Http
                 .map(json -> new TopicPartition(json.getString("topic"), json.getInteger("partition")))
                 .collect(Collectors.toSet());
 
-        Handler<AsyncResult<Void>> seekHandler = done -> {
-            if (done.succeeded()) {
+        BiConsumer<Void, Throwable> seekHandler = (v, ex) -> {
```
tombentley, line 181: Wouldn't it be more natural to just use a lambda on line 198 in the call to `whenComplete`, rather than declaring a `BiConsumer` variable here and passing that to `whenComplete`?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -462,13 +470,15 @@ public void doListSubscriptions(RoutingContext routingContext) {
      * @param routingContext the routing context
      */
     public void doUnsubscribe(RoutingContext routingContext) {
-        this.unsubscribe(unsubscribeResult -> {
-            if (unsubscribeResult.succeeded()) {
+        // fulfilling the request in a separate thread to free the Vert.x event loop still in place
+        CompletableFuture.runAsync(() -> this.unsubscribe()).whenComplete((v, ex) -> {
```
tombentley, line 474: Am I alone in finding these lines with both `runAsync()` and `whenComplete()` to be not as clear as putting them on separate lines?

src/main/java/io/strimzi/kafka/bridge/NoopPartitionsRebalance.java
ppatierno: I was wandering the same, but waiting for feedback on my feeling without spoilering it :-) I will change the name.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: Totally fine with my, I changed in another place as well and removed the `BiConsumer` import.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
ppatierno: I did just because the kind of "very few" code in the runAsync, but I am fine to move it on next line ;-)

scholzj (approved):
LGTM. But I only run through it quickly ... should be probably also reviewed by Tom ;-)

ppatierno:
Thanks @scholzj ... yeah Tom already had a first pass. @tombentley I addressed your comments, any more feedback on this one?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -232,86 +215,71 @@ protected void assign(Handler<AsyncResult<Void>> assignHandler) {
...
-        this.consumer.commit(offsetsData, commitOffsetsHandler);
+    /**
+     * Commit the offsets on topic partitions provided as parameter
+     *
+     * @param offsetsData map containing topic partitions and corresponding offsets to commit
+     * @return map containing topic partitions and corresponding committed offsets
+     */
+    protected Map<TopicPartition, OffsetAndMetadata> commit(Map<TopicPartition, OffsetAndMetadata> offsetsData) {
+        log.trace("Commit thread {}", Thread.currentThread());
+        // TODO: doesn't it make sense to change using the commitAsync?
```
tombentley, line 240: Do you want to open an issue, so you don't forget this?

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
```diff
@@ -232,86 +215,71 @@ protected void assign(Handler<AsyncResult<Void>> assignHandler) {
...
+        this.consumer.commitSync(offsetsData);
+        return offsetsData;
     }
 
-    protected void commit(Handler<AsyncResult<Void>> commitHandler) {
-        this.consumer.commit(commitHandler);
+    /**
+     * Commit offsets returned on the last poll() for all the subscribed list of topics and partitions
+     */
+    protected void commit() {
```
tombentley, line 249: Maybe it would be clearer if this one was called `commitLastPolledOffsets()`?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -362,11 +359,16 @@ private void doAssign(RoutingContext routingContext, JsonObject bodyAsJson) {
...
-            }
-        });
+        // fulfilling the request in a separate thread to free the Vert.x event loop still in place
+        CompletableFuture.runAsync(() -> this.assign())
+                .whenComplete((v, ex) -> {
+                    log.trace("Assign handler thread {}", Thread.currentThread());
+                    if (ex == null) {
+                        HttpUtils.sendResponse(routingContext, HttpResponseStatus.NO_CONTENT.code(), null, null);
+                    } else {
+                        // TODO: with Vert.x we did nothing, are we sure? when assign can fail?
```
tombentley, line 369: Do you want to fix this, or open an issue for it?

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -392,25 +394,29 @@ private void doSubscribe(RoutingContext routingContext, JsonObject bodyAsJson) {
...
+            } else if (bodyAsJson.containsKey("topic_pattern")) {
+                Pattern pattern = Pattern.compile(bodyAsJson.getString("topic_pattern"));
+                this.subscribe(pattern);
+            }
+        }).whenComplete((v, ex) -> {
+            log.trace("Subscribe handler thread {}", Thread.currentThread());
+            if (ex == null) {
                 HttpUtils.sendResponse(routingContext, HttpResponseStatus.NO_CONTENT.code(), null, null);
+            } else {
+                // TODO: with Vert.x we did nothing, are we sure? when subscribe can fail?
```
tombentley, line 417: Same comment

tombentley (approved):
Left a few nits, but otherwise LGTM.

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
ppatierno: no hurt to change that even if `commit` is just the same name as used in the Java Kafka client having the same behaviour.

src/main/java/io/strimzi/kafka/bridge/SinkBridgeEndpoint.java
ppatierno: I will open an issue. I want to think more about that, if it's really necessary.

ppatierno: Opened https://github.com/strimzi/strimzi-kafka-bridge/issues/735

PR #736, opened 2022-12-27
Replace Vert.x Kafka client with native Java Kafka client in the admin endpoint

ppatierno:
This PR is about replacing the usage of the Vert.x Kafka client on the admin endpoint with the native Java Kafka client (provided by Apache Kafka upstream project).

It provides the following things:

* using CompletableFuture(s) based pattern to run consumer requests asynchronously (as today it's done by Vert.x)
* removing the Vert.x instance from the admin side
* replacing Vert.x Kafka related classes, i.e. TopicPartition, TopicPartitionInfo, TopicDescription and more with the corresponding native ones from the Java Kafka client

With this last PR, the `vertx-kafka-client` dependency is not needed anymore in the bridge but I had to keep it but moving to `test` scope because the Vert.x Kafka client is still used in the tests (which should be re-written from scratch, but it's a different bigger work to do not in this PR).

Anyway, there are still some Vert.x structures in place like the ones handling JSON, Buffer(s) and Handler(s).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,11 +132,6 @@
 		</dependency>
 		<dependency>
 			<groupId>io.vertx</groupId>
-			<artifactId>vertx-kafka-client</artifactId>
-			<version>${vertx.version}</version>
-		</dependency>
-		<dependency>
-			<groupId>io.vertx</groupId>
 			<artifactId>vertx-web</artifactId>
 			<version>${vertx.version}</version>
 		</dependency>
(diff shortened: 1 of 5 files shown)


scholzj (approved):
LGTM. But the tests seem to be failing.

ppatierno:
I restarted them. They were not related to my changes. It looks like some problems to start the testcontainer, maybe temporary. Let's see.

scholzj:
Yeah, looks they passed now.

PR #737, opened 2022-12-30
Replaced Vert.x JSON handling

### LINKED ISSUE #410: Consume message with json as payload, but with simple text as key.
Hello,

I follow the getting started to produce and consume a json payload.
https://strimzi.io/blog/2019/11/05/exposing-http-bridge/

The event payload is a json object, and the key is a simple string.

The consequence is when I poll message, I have this error.
`{"error_code":406,"message":"Failed to decode: Unrecognized token ....`
because strimzi kafka bridge try to convert my text to json.

If I change the format to "binary", and the accept to "application/vnd.kafka.binary.v2+json", it works, I can get the base64 string, and then convert to text, and then to json.

Is it not possible to consume the payload as json, but with accepting string as key?
I should probably not bethe only one in this case, the key used in Kafka is simply a string and not a json object, a text key is more logic.

Thanks.

ppatierno:
This PR is about replacing the JSON handling happening inside the bridge with Vert.x classes by using GSON library instead.
It doesn't remove the usage of the Vert.x `Buffer` class which is still in place. That removal will happen in a different PR.

The PR fixes #410 as well. It is able to handle simple strings (in key or value) as pure strings and not JSON strings but at same time it is able to handle JSON object used as key or value in the messages.
This different strings handling is also the reason why some fixes in the tests, where the simple strings are not JSON strings anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,12 +119,18 @@
 		<test-container.version>0.102.0</test-container.version>
 		<jakarta.version>2.3.2</jakarta.version>
 		<snakeyaml.version>1.33</snakeyaml.version>
+		<gson.version>2.9.1</gson.version>
 		<!-- property to skip surefire tests during failsafe execution -->
 		<!--suppress UnresolvedMavenProperty -->
 		<skip.surefire.tests>${skipTests}</skip.surefire.tests>
 	</properties>
 
 	<dependencies>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+			<version>${gson.version}</version>
+		</dependency>
 		<dependency>
 			<groupId>io.vertx</groupId>
 			<artifactId>vertx-core</artifactId>
@@ -173,6 +179,11 @@
 		<dependency>
 			<groupId>io.netty</groupId>
 			<artifactId>netty-common</artifactId>
+			<version>${netty.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.netty</groupId>
+			<artifactId>netty-buffer</artifactId>
 			<version>${netty.version}</version>
 		</dependency>
 		<dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -119,18 +119,12 @@
 		<test-container.version>0.102.0</test-container.version>
 		<jakarta.version>2.3.2</jakarta.version>
 		<snakeyaml.version>1.33</snakeyaml.version>
-		<gson.version>2.9.1</gson.version>
 		<!-- property to skip surefire tests during failsafe execution -->
 		<!--suppress UnresolvedMavenProperty -->
 		<skip.surefire.tests>${skipTests}</skip.surefire.tests>
 	</properties>
 
 	<dependencies>
-		<dependency>
-			<groupId>com.google.code.gson</groupId>
-			<artifactId>gson</artifactId>
-			<version>${gson.version}</version>
-		</dependency>
 		<dependency>
 			<groupId>io.vertx</groupId>
 			<artifactId>vertx-core</artifactId>
--- a/src/main/java/io/strimzi/kafka/bridge/http/converter/JsonUtils.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/converter/JsonUtils.java
@@ -21,7 +21,7 @@ public class JsonUtils {
     /**
      * ObjectMapper instance used for JSON encoding/decoding
      */
-    public static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final ObjectMapper MAPPER = new ObjectMapper();
 
     /**
      * Get the JSON representation of the provided buffer content
--- a/src/main/java/io/strimzi/kafka/bridge/http/converter/JsonUtils.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/converter/JsonUtils.java
@@ -110,7 +110,7 @@ public class JsonUtils {
      * @param field name of the field representing the child node
      * @return int value of the child node or null if it doesn't exist
      */
-    public static int getInt(JsonNode json, String field) {
+    public static Integer getInt(JsonNode json, String field) {
         return json.path(field).isMissingNode() ? null : json.get(field).asInt();
     }
 
@@ -122,7 +122,7 @@ public class JsonUtils {
      * @param field name of the field representing the child node
      * @return long value of the child node or null if it doesn't exist
      */
-    public static long getLong(JsonNode json, String field) {
+    public static Long getLong(JsonNode json, String field) {
         return json.path(field).isMissingNode() ? null : json.get(field).asLong();
     }
 }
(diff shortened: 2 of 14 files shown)


pom.xml
```diff
@@ -175,6 +181,11 @@
 			<artifactId>netty-common</artifactId>
 			<version>${netty.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>io.netty</groupId>
+			<artifactId>netty-buffer</artifactId>
+			<version>${netty.version}</version>
+		</dependency>
```
scholzj, line 188: You seem to add this dependency but do not really use it. Is it needed as some GSON override? Or why?

scholzj:
In most Strimzi components, we are currently using Jackson for JSON (de)serialization. Jackson is also used by our key dependencies (Kafka, Fabric8, ...). So I wonder if using Jackson here over Gson would make more sense. We cannot get rid of Jackson and adding a new dependency would just increase our footprint for things such as CVEs etc. Jackson should allow you to do the same handling through the `JsonNode` or `ObjectNode` classes. So you should not need the POJOs to use it. I think you should be able to just create a Jackson `ObjectMapper` and use its `readTree` method to get it or something like that.

ppatierno:
It was my doubt, about adding one more dependency. Maybe I missed something around the Jackson usage and I didn't find it as simple as GSON. I will give it another try.

ppatierno:
I have made some changes locally, not pushed to this PR yet. It was a simple change but there is a thing which I am struggling with.
The bridge should allow to send key and value as simple String (not JSON string, so no quotes around) and as JSON objects. This is what we allow via the OpenAPI spec.

What I am struggling with is when an HTTP client tries to consume a Kafka message so we are dealing with array of bytes for the `key` and the `value` and we should be able to detect that it's a:

* simple String (with no quotes)
* a JSON object
* a malformed JSON object to return an error

Regarding the first point, the current solution is leveraging the `JsonPrimitive` type in GSON and the following few lines of code to detect that the bytes represent a simple String:

```java
JsonElement json = JsonParser.parseString(new String(bytes, StandardCharsets.UTF_8));
            if (!json.isJsonPrimitive()) {
                jsonObject.add(key, json);
            } else {
                jsonObject.addProperty(key, json.getAsString());
            }
```

I haven't found a good way with Jackson. It doesn't have a way to detect that passed data represents a JSON primitive type and not an actual JSON node. The GSON solution is very smart and handy imho. 
Any try of Jackson 'ObjectMapper` methods like `readTree`, `readValue` just fails but the same happens when the JSON is malformed so no way to distinguish the two different situations.
After more thinking the only hack-ish way I found is to assume that what we get is a simple String with this code:

```java
private boolean isPlainString(byte[] bytes) {
        String tmp = new String(bytes, StandardCharsets.UTF_8);
        return !tmp.startsWith("{") && !tmp.startsWith("[") && !tmp.startsWith("\"");
    }
```

Last but not the least, one of our tests sends this message as invalid JSON `"Hello": "World"-0` .... and surprise ... Jackson doesn't fail! It assumes the `"Hello"` as first and valid token :-(
I am not going to push the change for now, but any good feedback or idea about the plain String problem is welcome :-)

ppatierno:
After more time thinking about this, I am almost convinced that what we have today is right and #410 should not be fixed as well ... I mean ...

The bridge, similarly to the Confluent REST proxy, enables two "embedded" types (the way how keys and values are encoded); binary and JSON. Corresponding to them we have two message converters, `HttpBinaryMessageConverter` (expecting Strings which represent Base64 encoded data) and `HttpJsonMessageConverter` (expecting JSON).
This is reflected in the OpenAPI spec where a key or value can be `string` or `object`.

Referring to the `HttpJsonMessageConverter`, it's handling JSON by nature so a key or value as `key-0` or `value-0`, for example, are today encoded in the Kafka record as JSON strings `"key-0"` and `"value-0"`. This is also how the Confluent REST proxy behaves. For example, using the proxy with something like

```json
"key": "key-1",
"value": "value-1",
```

brings to an HTTP consumer the same result but to a Kafka native client (see kafka-console-consumer) something like this:

```shell
"key-1"	"value-1"
```

(see the added double quotes, by encoding the original `key-1` and `value-1` as JSON)

I think that what we are trying to achieve, so having the consumer getting just `key-1` and `value-1` (without quotes), is not right taking into account the nature of the JSON message converter which is in charge to handle the "json" embedded format.

Other than GSON which provides a "candy" about getting the JSON string primitive type as explained before, we cannot say that something which is not a JSON is then a malformed JSON or just a String. We can be opinionated and making a decision (if not good JSON, then pure string) but it's not a good way to go imho.

I think that we could move to use Jackson but leaving the current behaviour, so not fixing #410 in the end.
To allow #410, the way is using the `binary` embedded type (as the user is doing) or maybe adding a new `raw` (or a better name) message converter for not doing any specific handling and allowing raw strings?

@scholzj  @tombentley any opinions on this?

scholzj:
@ppatierno I think it would be valuable to have a _String_ mode as well in addition to the binary and JSON modes. But I agree that forcing it through the JSON mode might not be the best option.

ppatierno:
@scholzj yeah, I went that direction and pushed changes using Jackson. I will open an issue about the possibility to have a `string` mode as a valid embedded format.

src/main/java/io/strimzi/kafka/bridge/http/converter/JsonUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Provides some utility methods for JSON encoding/decoding
+ */
+public class JsonUtils {
+
+    /**
+     * ObjectMapper instance used for JSON encoding/decoding
+     */
+    public static final ObjectMapper MAPPER = new ObjectMapper();
```
scholzj, line 24: Does this need to be public? It seems to be used only here.

pom.xml
scholzj: I still don't follow why do you have this change here.

src/main/java/io/strimzi/kafka/bridge/http/converter/JsonUtils.java
ppatierno: Good catch, I will fix it.

pom.xml
ppatierno: Tbh I am confused as well Jakub.
Without adding this dependency I am getting:

```shell
[WARNING] Used undeclared dependencies found:
[WARNING]    io.netty:netty-buffer:jar:4.1.85.Final:compile
```

But I am not sure where, in the new code, I am using something from this JAR :-(
It was the same with GSON and now again with Jackson.

scholzj: Ok, mysterious way of Maven plugins ... but not much we can do about it other than add it. Thanks for the explanation.

ppatierno: Let's say that the next step will be removing the Vert.x `Buffer` class which is anyway using Buffer from Netty, and I will use it directly, so the jar will be needed at that point :-)

scholzj (approved):

scholzj (approved):

ppatierno:
This PR now completes the removal of Vert.x JSON usage.
Any other usage of Vert.x JSON related classes (i.e. `JsonObject`, `JsonArray`, ...) is related to tests and the Vert.x Config module (in the main) which gets a configuration via a `JsonObject`.
Rewriting tests and replacing the Vert.x Config module will be done in future PRs.

ppatierno:
So, I am going to merge this and I opened #739 in order to track the idea of a `string` embedded format.

PR #740, opened 2023-01-03
Removed usage of Vert.x Buffer

ppatierno:
This PR removes the usage of the Vert.x `Buffer` class, just moving bytes around.
It's still used: 

1. when writing on the HTTP routing context response
2. in the tests.

The above will be fixed in future PRs when we are not going to use the Vert.x HTTP server features anymore and rewriting tests without Vert.x

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,11 +176,6 @@
 			<version>${netty.version}</version>
 		</dependency>
 		<dependency>
-			<groupId>io.netty</groupId>
-			<artifactId>netty-buffer</artifactId>
-			<version>${netty.version}</version>
-		</dependency>
-		<dependency>
 			<groupId>org.slf4j</groupId>
 			<artifactId>slf4j-api</artifactId>
 			<version>${slf4j.version}</version>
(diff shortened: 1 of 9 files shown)


scholzj (approved):

tombentley (commented):
The code changes look fine to me. My only question is what the performance impact of this change is, since (I assume) it must now be allocating a `byte[]` for every response (in addition to the `Buffer`), where before it was just passing a `Buffer` around, i.e. transiently we need twice the memory for each response. Or will this get changed again soon with a future PR?

ppatierno:
@tombentley that's a good point. I took a look at the internal Vert.x code, I didn't realize that even the `getBytes` from the buffer instance is copying the data to a new one (I thought it was returning the internal array presentation).
Of course, about sending the response, it's now creating a new buffer from the bytes so allocating another array internally.

> Or will this get changed again soon with a future PR?

It won't happen soon because it's related to removing the Vert.x Web and OpenAPI dependencies which are the more complex (we need to identify a good "plain" Java replacement to avoid having all HTTP requests handling issues on our shoulders).

At this point, I would say, let's close this one and we'll get rid of the Vert.x `Buffer` at the same time we remove the Vert.x Web and OpenAPI components. What we'll chose could have something similar (i.e. netty `ByteBuf`) to move around.

@tombentley @scholzj do you agree?

tombentley:
That's fine with me @ppatierno

ppatierno:
Closing this one as agreed.

PR #741, opened 2023-01-03
Replaced Vert.x Handler class and endpoint abstractions

ppatierno:
This PR:

* add a one-to-one replacement for the Vert.x `Handler` class, in order to remove one more Vert.x related stuff but leaving same logic
* remove endpoint abstractions still in place after the AMQP 1.0 removal. Now it's all about HTTP and `RoutingContext` usage (that will go away at some point in the future as well, when removing the Vert.x Web dependency).


scholzj (approved):

src/main/java/io/strimzi/kafka/bridge/Handler.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package io.strimzi.kafka.bridge;
+
+/**
+ * Represents a functional interface for handling callback when an asynchronous operation ends
+ *
+ * @param <T> type of the data to be handled
+ */
+public interface Handler<T> {
```
tombentley, line 13: If it's intended to be a functional interface then you should add `@FunctionalInterface`.

tombentley (approved):

ppatierno:
Going to merge this. It seems to be a Travis CI problem with ppc64le platform not able to boot machine sometimes and install needed stuff. I tried a ton of times since yesterday.

PR #742, opened 2023-01-04
Replaced Vert.x Config component with plain Java config retriever

ppatierno:
This trivial PR replaces the Vert.x Config component with a simple `ConfigRetriever` class for getting bridge configuration from both properties file and environment variables.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/config/ConfigRetriever.java
+++ b/src/main/java/io/strimzi/kafka/bridge/config/ConfigRetriever.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+package io.strimzi.kafka.bridge.config;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+/**
+ * Retrieve the bridge configuration from properties file and environment variables
+ */
+public class ConfigRetriever {
+
+    /**
+     * Retrieve the bridge configuration from the properties file provided as parameter
+     * and adding environment variables
+     * If a parameter is defined in both properties file and environment variables, the latter wins
+     *
+     * @param path path to the properties file
+     * @return configuration as key-value pairs
+     * @throws IOException when not possible to get the properties file
+     */
+    public static Map<String, Object> getConfig(String path) throws IOException {
+        Map<String, Object> configuration;
+        try (InputStream is = new FileInputStream(path)) {
+            Properties props = new Properties();
+            props.load(is);
+            configuration =
+                    props.entrySet().stream().collect(
+                            Collectors.toMap(
+                                    e -> String.valueOf(e.getKey()),
+                                    e -> String.valueOf(e.getValue()),
+                                    (prev, next) -> next, HashMap::new
+                            ));
+        }
+        configuration.putAll(System.getenv());
+        return configuration;
+    }
+}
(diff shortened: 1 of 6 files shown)


src/main/java/io/strimzi/kafka/bridge/config/ConfigRetriever.java
```diff
@@ -0,0 +1,46 @@
...
+            props.load(is);
+            configuration =
+                    props.entrySet().stream().collect(
+                            Collectors.toMap(
+                                    e -> String.valueOf(e.getKey()),
+                                    e -> String.valueOf(e.getValue()),
+                                    (prev, next) -> next, HashMap::new
+                            ));
+        }
+        configuration.putAll(System.getenv());
```
scholzj, line 43: Should it do some filtering for some prefixes we support here? Or do we really want to accept everything? Also, should it follow some commonly used conversions between env. vars and properties? E.g. `HTTP_PORT` to `http.port`? Or do you expect the env. vars to be named the same as the properties?

scholzj (commented):
Should it have some unit test?

scholzj:
I also wonder ... if we move away to completely our stuff, would it make sense to have it more tightůy integrated into the `*Config` classes?

src/main/java/io/strimzi/kafka/bridge/config/ConfigRetriever.java
ppatierno: Replicating the same behaviour of Vert.x Config, it's just loading everything and there is no convention about names, you have to define the env var as the same name of the property to override (so no `HTTP_PORT` to `http.port`). The filtering happens later when we build the actual config.
Anyway, while doing this I was not sure to continue to support env vars even because we don't use them anyway and everything is configured via properties file in the bridge.

scholzj: > Anyway, while doing this I was not sure to continue to support env vars even because we don't use them anyway and everything is configured via properties file in the bridge.

Might deserve a wider discussion and checking with users if they use it or not. But in general yes, it might make sense.

ppatierno:
Our `*Config` classes needs a way to get the application properties file content anyway and then filtering each of them what they need (i.e. `http.`, `producer.` and so on). Loading configuration in a central place into a map and then filtering later makes sense to me instead of each `*Config` class accessing the properties file.

tombentley (commented):
Please can you add some unit tests.

ppatierno:
Good point, sorry for forgetting them. Just committed!

tombentley (approved):
Thanks!

im-konge (approved):

src/test/java/io/strimzi/kafka/bridge/ConfigRetrieverTest.java
```diff
@@ -0,0 +1,90 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Unit tests for the ConfigRetriever class
+ */
+public class ConfigRetrieverTest {
+
+    @Test
+    public void testApplicationPropertiesFile() throws IOException {
```
katheris, line 36: Is this test actually testing that the ConfigRetriever correctly reads the properties file and gets the right value out? If I changed the ConfigRetriever class to always return "foo" as the value for every key I think this test would still pass right? It's testing whether the BridgeConfig class is using the ConfigRetriever, rather than testing that the ConfigRetriever correctly reads config I think

ppatierno: Good point. I think you are right. I will change the assertions to test that the bridge configuration (got through the retriever) actually matches the values in the application.properties file (which means that config retriever was able to read the file correctly as well).

ppatierno: @katheris I updated the test. Does it look better now?

ppatierno:
@scholzj any more comments on this?

scholzj (approved):

src/test/java/io/strimzi/kafka/bridge/ConfigRetrieverTest.java
katheris: Yeah that looks better

katheris (approved):

PR #751, opened 2023-01-20
Use JVM in `dynamic_resources.sh` for calculating memory

### LINKED ISSUE #731: OOM of Kafka bridge container due to cgroupv2 (Debian11, Ubuntu 21+...)
By default, new distributions of linux use cgroupv2
Strimzi uses /sys/fs/cgroup/memory.limit_in_bytes in dyamic_resources.sh, but this stat does not exist in cgroupv2.
Therefore the script returns nothing, and the JVM uses default memory setting.
When JVM Heap exceeds configured pod limit, pod is killed.

Solution: Detect if cgroup is v2 (e.g. memory.limit_in_bytes does not exist), and use instead memory.max which is in cgroupv2.

The same bug applies also to Zookeeper and Kafka Pods launched by Strimzi operator, but is maybe present also in operator pods.

im-konge:

This PR changes the approach, which was used in `dynamic_resources.sh` until now. The script now uses JVM for calculating the memory - so it also covers cgroups v2.

Fixes #731

### DIFF
--- a/bin/docker/dynamic_resources.sh
+++ b/bin/docker/dynamic_resources.sh
@@ -1,7 +1,17 @@
 #!/usr/bin/env bash
 
+mem_file_cgroups_v1="/sys/fs/cgroup/memory/memory.limit_in_bytes"
+mem_file_cgroups_v2="/sys/fs/cgroup/memory.max"
+
 function get_heap_size {
-  CONTAINER_MEMORY_IN_BYTES=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+  CONTAINER_MEMORY_IN_BYTES=""
+
+  if [ -f "$mem_file_cgroups_v1" ]; then
+    CONTAINER_MEMORY_IN_BYTES = $(cat "$mem_file_cgroups_v1")
+  else
+    CONTAINER_MEMORY_IN_BYTES = $(cat "$mem_file_cgroups_v2")
+  fi
+
    # use max of 31G memory, java performs much better with Compressed Ordinary Object Pointers
   DEFAULT_MEMORY_CEILING=$((31 * 2**30))
   if [ "${CONTAINER_MEMORY_IN_BYTES}" -lt "${DEFAULT_MEMORY_CEILING}" ]; then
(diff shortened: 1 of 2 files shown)


scholzj:
I this the right approach? Shouldn't it use JVM like e.g. the Kafka containers?

im-konge:
Yeah, that makes sense. I will change it, thanks :)

scholzj (approved):
LGTM if it works. Should be reviewed by Paolo as an SME.

im-konge:
I tested the changes and it seems that it works. This is a few lines of log of bridge container.
```
Kafka Bridge configuration:
#Bridge configuration
bridge.id=http-bridge-cluster-name

#Kafka common properties
kafka.bootstrap.servers=http-bridge-cluster-name-kafka-bootstrap:9092
kafka.security.protocol=PLAINTEXT

#Apache Kafka AdminClient

#Apache Kafka Producer

#Apache Kafka Consumer
kafka.consumer.auto.offset.reset=earliest

#HTTP configuration
http.host=0.0.0.0
http.port=8080
http.cors.enabled=false
http.cors.allowedOrigins=
http.cors.allowedMethods=

Configuring Java heap: -Xms1244m -Xmx1244m
++ dirname /opt/strimzi/bin/docker/../kafka_bridge_run.sh
+ MYPATH=/opt/strimzi/bin/docker/..
+ '[' -z -Dlog4j2.configurationFile=file:/opt/strimzi/custom-config/log4j2.properties ']'
+ JAVA_OPTS='-Xms1244m -Xmx1244m   -Dvertx.cacheDirBase=/tmp/vertx-cache -Djava.security.egd=file:/dev/./urandom'
+ '[' -n '' ']'
+ exec java -Xms1244m -Xmx1244m -Dvertx.cacheDirBase=/tmp/vertx-cache -Djava.security.egd=file:/dev/./urandom -Dlog4j2.configurationFile=file:/opt/strimzi/custom-config/log4j2.properties -classpath '/opt/strimzi/bin/docker/../../libs/*' io.strimzi.kafka.bridge.Application --config-file=/tmp/kafka-bridge.properties
```

ppatierno (approved):
LGTM. I tried that as well, it seems to work :-)

PR #754, opened 2023-01-26
Refactoring HTTP endpoints

ppatierno:
This PR refactors the HTTP sink/source/admin endpoints part.
It moves from using inheritance to composition so that an HTTP endpoint doesn't extend the class containing the Kafka logic but uses it instead:

* the `HttpSinkBridgeEndpoint`, `HttpSourceBridgeEndpoint` and `HttpAdminBridgeEndpoint` derives from `HttpBridgeEndpoint` where most of the common behaviour was moved.
* the above endpoints now use `KafkaBrideConsumer`, `KafkaBridgeProducer` and `KafkaBridgeAdmin` which are the old sink/source/admin base endpoint but still thin wrappers around the Java clients API.

While everything could be still collapsed in one class per endpoint I think that it's still better having more separation of concerns: HTTP endpoint part vs Kafka related communication.
It also avoids to have just one bigger class per endpoint type.
I was able to remove some `@SupperssWarnings` for the checkstyle related to the complexity.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeProducer.java
+++ b/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeProducer.java
@@ -5,9 +5,7 @@
 
 package io.strimzi.kafka.bridge;
 
-import io.strimzi.kafka.bridge.config.BridgeConfig;
 import io.strimzi.kafka.bridge.config.KafkaConfig;
-import io.strimzi.kafka.bridge.http.HttpBridgeEndpoint;
 import io.strimzi.kafka.bridge.tracing.TracingHandle;
 import io.strimzi.kafka.bridge.tracing.TracingUtil;
 import org.apache.kafka.clients.producer.KafkaProducer;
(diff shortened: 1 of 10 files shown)


scholzj (approved):
LGTM. Might be good to get review from @tombentley as well.

src/main/java/io/strimzi/kafka/bridge/KafkaBridgeProducer.java
```diff
@@ -84,7 +53,7 @@ protected void handleClose() {
      * @param record Kafka record to send
      * @return a CompletionStage bringing the metadata
      */
-    protected CompletionStage<RecordMetadata> send(ProducerRecord<K, V> record) {
+    public CompletionStage<RecordMetadata> send(ProducerRecord<K, V> record) {
```
tombentley, line 56: `Producer#send` can block. That's not a problem for how this `KafkaBridgeProducer#send` is called in the `HttpSourceBridgeEndpoint`, but maybe it's worth mentioning in the doc?

tombentley (approved):
What's not to love about a diffstat like +295/−346

PR #757, opened 2023-01-31
Added async query parameter on send/sentToPartition in OpenAPI v3 and v2

ppatierno:
The `async` query parameter was totally missing in the OpenAPI v3 specification and missing in the `sentToPartition` operation for v2.
Because the bridge uses v3 to validate, there was no validation on the boolean value for that query parameter.
This PR fixes it and add a test as well.


scholzj (approved):
The changes look good. But I guess you need to update the docs?

PR #761, opened 2023-02-03
Added clarification about subscribe operation and no records on poll

ppatierno:
Quite often we got user asking why after a subscribe operation, the poll doesn't return any records.
This PR tries to clarify it, explaining that a few poll are needed at the beginning (no records returned) to run the partitions rebalance and join-group process.

### DIFF
--- a/documentation/modules/proc-bridge-subscribing-consumer-topics.adoc
+++ b/documentation/modules/proc-bridge-subscribing-consumer-topics.adoc
@@ -28,9 +28,9 @@ The `topics` array can contain a single topic (as shown here) or multiple topics
 +
 If the request is successful, the Kafka Bridge returns a `204` (No Content) code only.
 
-As it happens with using an Apache Kafka client, the HTTP subscribe operation just adds the topics in the local Kafka consumer subscriptions list.
-In order to join the consumer group and getting partitions assigned, the client has to run multiple poll operations for starting the partitions rebalance.
-It means that after the HTTP subscribe operation, the first few HTTP poll operations could return no records but will start the partitions rebalance and the join-group.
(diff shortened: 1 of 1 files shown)


scholzj (approved):

documentation/modules/proc-bridge-subscribing-consumer-topics.adoc
```diff
@@ -28,6 +28,10 @@ The `topics` array can contain a single topic (as shown here) or multiple topics
 +
 If the request is successful, the Kafka Bridge returns a `204` (No Content) code only.
 
+As it happens with using an Apache Kafka client, the HTTP subscribe operation just adds the topics in the local Kafka consumer subscriptions list.
+In order to join the consumer group and getting partitions assigned, the client has to run multiple poll operations for starting the partitions rebalance.
+It means that after the HTTP subscribe operation, the first few HTTP poll operations could return no records but will start the partitions rebalance and the join-group.
```
PaulRMellor, line 33:
```suggestion
When using Apache Kafka, the HTTP subscribe operation adds topics to the local consumer's subscriptions. 
Joining a consumer group and obtaining partition assignments occur after running multiple HTTP poll operations, starting the partition rebalance and join-group process. 
It's important to note that the initial HTTP poll operations may not return any records.
```

PaulRMellor (approved):
LGTM :+1:

documentation/modules/proc-bridge-subscribing-consumer-topics.adoc
ppatierno: I pushed changes, but for the first sentence I used "When using an Apache Kafka client, ..."
It's important to highlight that I am talking about a client talking the Apache Kafka protocol. 
"When using Apache Kafka" ... doesn't say much, it's crystal clear we are using Apache Kafka :-D

Issue #756, opened 2023-01-30
Subscribe and assign with empty topics/partitions list should be an unsubscribe operation


ppatierno:
The Java Apache Kafka consumer API provides the `subscribe` and `assign` methods to subscribe to topics or being assigned partitions. The same kind of behaviour is exposed by the bridge via HTTP.
By the way, when using Java Kafka API, providing an empty list of topics to `subscribe` or an empty list of partitions to the `assign` acts as an `unsubscribe` operation. See Kafka consumer code as reference:

https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L960

https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1110

Actually, in the bridge, when you send a subscribe or assign request with empty `topics` or `partitions` list you get the following error:

```json
{
    "error_code": 500,
    "message": "java.lang.IllegalArgumentException: At least one topic to subscribe has to be specified!"
}
```

We should fix this and providing the same behaviour of the Java Kafka API by doing an unsubscribe.
NOTE: of course, this doesn't apply to the subscription with pattern.

haijun2022:
Hi @ppatierno , is this issue still valid and the solution finalized? I'd like to fix this issue based on the requirement? Thank you!

ppatierno:
@haijun2022 yes it is still in place. Happy to know you are interested on working on it. I will assign it to you. Let me know if you need any guidance.

haijun2022:
thank you, @ppatierno. I will reach out to you in case I need your further support!

ppatierno:
Closed by #763

PR #763, opened 2023-02-07
Update subscribe and assign method to follow same processing logic as apache kafka

haijun2022:
This PR fixed the bug described in issue https://github.com/strimzi/strimzi-kafka-bridge/issues/756

it update subscribe and assign method to follow same processing logic as apache kafka.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
+++ b/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
@@ -99,6 +99,7 @@ public class KafkaBridgeConsumer<K, V> {
 
         if (topicSubscriptions.isEmpty()) {
             this.unsubscribe();
+            return;
         }
 
         log.info("Subscribe to topics {}", topicSubscriptions);
@@ -157,6 +158,7 @@ public class KafkaBridgeConsumer<K, V> {
 
         if (topicPartitions.isEmpty()) {
             this.unsubscribe();
+            return;
         }
 
         log.trace("Assign thread {}", Thread.currentThread());
--- a/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
+++ b/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
@@ -93,8 +93,12 @@ public class KafkaBridgeConsumer<K, V> {
      * @param topicSubscriptions topics to subscribe to
      */
     public void subscribe(List<SinkTopicSubscription> topicSubscriptions) {
+        if (topicSubscriptions == null) {
+            throw new IllegalArgumentException("Topic subscriptions cannot be null");
+        }
+
         if (topicSubscriptions.isEmpty()) {
-            throw new IllegalArgumentException("At least one topic to subscribe has to be specified!");
+            this.unsubscribe();
         }
 
         log.info("Subscribe to topics {}", topicSubscriptions);
@@ -140,8 +144,8 @@ public class KafkaBridgeConsumer<K, V> {
      * @param topicSubscriptions topics to be assigned
      */
     public void assign(List<SinkTopicSubscription> topicSubscriptions) {
-        if (topicSubscriptions.isEmpty()) {
-            throw new IllegalArgumentException("At least one topic to subscribe has to be specified!");
+        if (topicSubscriptions == null) {
+            throw new IllegalArgumentException("Topic subscriptions cannot be null");
         }
 
         log.info("Assigning to topics partitions {}", topicSubscriptions);
@@ -149,6 +153,10 @@ public class KafkaBridgeConsumer<K, V> {
         Set<TopicPartition> topicPartitions = new HashSet<>();
         for (SinkTopicSubscription topicSubscription : topicSubscriptions) {
             topicPartitions.add(new TopicPartition(topicSubscription.getTopic(), topicSubscription.getPartition()));
+        }
+
+        if (topicPartitions.isEmpty()) {
+            this.unsubscribe();
         }
 
         log.trace("Assign thread {}", Thread.currentThread());
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
@@ -49,7 +49,7 @@ public class ConsumerSubscriptionIT extends HttpBridgeITAbstract {
 
         CompletableFuture<Boolean> unsubscribe = new CompletableFuture<>();
         consumerService()
-            .deleteRequest(Urls.consumerInstanceSubscription(groupId, name + "consumer-invalidation"))
+                .deleteRequest(Urls.consumerInstanceSubscription(groupId, name + "consumer-invalidation"))
                 .putHeader("Content-length", String.valueOf(topicsRoot.toBuffer().length()))
                 .as(BodyCodec.jsonObject())
                 .sendJsonObject(topicsRoot, ar -> {
@@ -78,7 +78,7 @@ public class ConsumerSubscriptionIT extends HttpBridgeITAbstract {
 
         // create consumer
         consumerService()
-            .createConsumer(context, groupId, json);
+                .createConsumer(context, groupId, json);
 
         // cannot subscribe setting both topics list and topic_pattern
         JsonArray topics = new JsonArray();
@@ -90,7 +90,7 @@ public class ConsumerSubscriptionIT extends HttpBridgeITAbstract {
 
         CompletableFuture<Boolean> subscribeConflict = new CompletableFuture<>();
         consumerService()
-            .subscribeConsumerRequest(groupId, name, topicsRoot)
+                .subscribeConsumerRequest(groupId, name, topicsRoot)
                 .sendJsonObject(topicsRoot, ar -> {
                     context.verify(() -> {
                         assertThat(ar.succeeded(), is(true));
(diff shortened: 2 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
```diff
@@ -93,8 +93,12 @@ public void create(Properties config, String groupId) {
      * @param topicSubscriptions topics to subscribe to
      */
     public void subscribe(List<SinkTopicSubscription> topicSubscriptions) {
+        if (topicSubscriptions == null) {
+            throw new IllegalArgumentException("Topic subscriptions cannot be null");
+        }
+
         if (topicSubscriptions.isEmpty()) {
-            throw new IllegalArgumentException("At least one topic to subscribe has to be specified!");
+            this.unsubscribe();
```
ppatierno, line 101: so after unsubscribing you are actually not returning and the subscribe operation will run anyway. It should not cause a problem just because the Kafka subscribe with empty topic list will behave as an unsubscribe.
For this reason I would say that this addition about calling unsubscribe is not needed, we should just allow an empty topics list and then allowing the code to run the subscribe.

src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
```diff
@@ -151,6 +155,10 @@ public void assign(List<SinkTopicSubscription> topicSubscriptions) {
             topicPartitions.add(new TopicPartition(topicSubscription.getTopic(), topicSubscription.getPartition()));
         }
 
+        if (topicPartitions.isEmpty()) {
+            this.unsubscribe();
+        }
```
ppatierno, line 160: same as described before. Run the unsubscribe here, not returning and then the code runs the assign.
Just running the assign with an empty topics list should work.

ppatierno (changes requested):

src/main/java/io/strimzi/kafka/bridge/KafkaBridgeConsumer.java
haijun2022: Thanks for the review, exactly, will update the code!

haijun2022:
@ppatierno @scholzj I have updated the code and added test code based on your suggestions. Could you reivew it again and let me know if there is any suggestions. thank you!

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -160,6 +161,62 @@ void subscriptionConsumerDoesNotExistBecauseNotCreated(VertxTestContext context)
...
+        JsonObject topicsRoot = new JsonObject();
+        topicsRoot.put("topics", topics);
+
+        // Create consumer and subscribe to the topic
+        JsonObject consumerJson = new JsonObject();
+        consumerJson.put("name", name);
+        consumerJson.put("format", "json");
+        consumerService()
+                .createConsumer(context, groupId, consumerJson)
+                .subscribeConsumer(context, groupId, name, topicsRoot);
```
ppatierno, line 181: after subscribing you should verify that it was ok by listing the subscription, so that after unsubscribe you can do the same and checking it's empty.
Anyway, for how the Kafka client works, after the subscribe operation you need to execute some poll/consume in order to allow the consumer to be really subscribed (joining the consumer group). The subscribe operation itself just fill a list inside the Kafka consumer nothing more.
See the `listConsumerSubscriptions` test which does something like that after subscribe and before listing the subscriptions.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -355,6 +412,60 @@ void assignAfterSubscriptionTest(VertxTestContext context) throws InterruptedExc
...
+        String consumerName = "my-kafka-consumer-assign-empty";
+
+        JsonObject consumerJson = new JsonObject();
+        consumerJson.put("name", consumerName);
+        consumerJson.put("format", "json");
+
+        JsonObject partitionsRoot = new JsonObject();
+        JsonArray partitions = new JsonArray();
+
+        partitionsRoot.put("partitions", partitions);
```
ppatierno, line 431: compared to the previous test, there is no assignment to a partition and then assignment with empty list to validate the feature.

ppatierno (changes requested):

ppatierno:
@haijun2022 I see you are on the right way! I noticed you added the check I suggested on the subscription test only but it should be on the assign test as well.

haijun2022:
> @haijun2022 I see you are on the right way! I noticed you added the check I suggested on the subscription test only but it should be on the assign test as well.

Thank you for the review, I tried but did not find a way to go to that logic branch as you suggested so I removed that test code! Will double check, meanwhile please let me know if that is doable!

ppatierno:
> Thank you for the review, I tried but did not find a way to go to that logic branch as you suggested so I removed that test code! Will double check, meanwhile please let me know if that is doable!

I am not sure what you mean by this :-(
You added the logic to unsubscribe when someone use `assign` but with empty partitions.
You added a test for it. But it was missing the suggestions I made to you on the other test (the one about unsubscribe via a `subscribe` with empty topics collection).

ppatierno:
@haijun2022 do you have any idea when this PR will be ready? We are in the process to start a 0.25.0 release and having your contribution included would be great :-)

haijun2022:
Hi @ppatierno, I am working on it now and will update by the end of today, thank you!

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
haijun2022: just added the section, thank you!

haijun2022:
> Hi @ppatierno, I am working on it now and will update by the end of today, thank you!

Hi @ppatierno , I just updated the code and please help review on it.  My IDE added some whitespaces which is not my intension, please ignore them during your review. Sorry for that!

ppatierno:
@haijun2022 it would be great you remove those whitespaces. They are noise in the PR and we don't want to merge the PR with them, thanks!

haijun2022:
> 

Yes, let me know if others are good now? I will update them later

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -160,6 +161,92 @@ void subscriptionConsumerDoesNotExistBecauseNotCreated(VertxTestContext context)
...
+                .as(BodyCodec.jsonArray())
+                .send(ar -> {
+                    if (ar.succeeded()) {
+                        LOGGER.info("Request result: {}", ar.result().body());
+                        consume.complete(true);
+                    }
+                });
+        consume.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+
+        CompletableFuture<Boolean> subscribe = new CompletableFuture<>();
```
ppatierno, line 197: can you create this later when it is really used just before the subscription to an empty topics list? Maybe even calling `subscribeEmpty` would be better. Thanks!

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -160,6 +161,92 @@ void subscriptionConsumerDoesNotExistBecauseNotCreated(VertxTestContext context)
...
+                        HttpResponse<JsonObject> response = ar.result();
+                        assertThat(response.statusCode(), is(HttpResponseStatus.NO_CONTENT.code()));
+                        assertThat(response.body(), is(nullValue()));
+                    });
+                    subscribe.complete(true);
+                });
+
+        subscribe.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+        // Validate the existing consumer list
+        CompletableFuture<Boolean> listSubscriptionsAfter = new CompletableFuture<>();
```
ppatierno, line 231: you can reuse the `listSubscriptions` variable as before just creating a `new CompletableFuture<>();`

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -355,6 +442,105 @@ void assignAfterSubscriptionTest(VertxTestContext context) throws InterruptedExc
...
+                        HttpResponse<JsonObject> response = ar.result();
+                        assertThat(response.statusCode(), is(HttpResponseStatus.NO_CONTENT.code()));
+                    });
+                    assignEmptyCF.complete(true);
+                });
+
+        assignEmptyCF.get(TEST_TIMEOUT, TimeUnit.SECONDS);
+
+        // Validate the existing consumer list
+        CompletableFuture<Boolean> listSubscriptionsAfter = new CompletableFuture<>();
```
ppatierno, line 524: same as before you should be able to use just one `listSubscriptions` completable future for before and after, just creating a new instance assigning to the same variable.

ppatierno:
@haijun2022 I think these are last comments and we are good to go.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
haijun2022: Hi @ppatierno , that does not work here. it shows "Local variable listSubscriptionsBefore defined in an enclosing scope must be final or effectively finalJava(536871575)"

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
haijun2022: updated, thanks!

ppatierno (approved):
LGTM. Thanks a lot for the contribution!

haijun2022:
> LGTM. Thanks a lot for the contribution!

Thanks a lot for your reivew and suggestions. @ppatierno

scholzj (approved):

PR #781, opened 2023-03-19
docs: fix example for POST request to /topics

avinashupadhya99:
Very minor update to the example for POST request to /topics

- Changes the position of the commas so that Bridge does not give errors when the example request is made as is with copy-paste


scholzj (approved):
Thanks for the PR. LGTM.

ppatierno (approved):
Thanks for the PR!

PR #782, opened 2023-03-22
Update Kafka Kubernetes Config Provider to 1.1.1

scholzj:
This PR updates the version of the Kafka Kubernetes Config Provider to 1.1.1 which brings updated dependencies and used the Java HTTP client. This helps us to address some CVEs in transitive dependencies such as SnakeYAML and OkHttp.


scholzj:
_Note: The TravisCI failure seems unrelated to this PR._

ppatierno:
@scholzj AFAIU this new version also brings snakeYAML 2.0 to fix related CVEs, is that right?
I don't think it's going to work with current bridge, where we are still forcing to snakeYAML 1.33 because the Vert.x OpenAPI component (in 4.3.8) doesn't work with snakeYAML 2.0 due to some broken backward compatibility.
We didn't update the JMX exporter to 1.8 yet for the same reason.
AFAIK, snakeYAML 2.0 will be available in Vert.x with version 4.4.x so we should wait for it before merging this one.

ppatierno (approved):

scholzj:
@ppatierno Thanks.

PR #784, opened 2023-04-13
Remove "remote" and "local" labels from HTTP server related metrics

### LINKED ISSUE #459: Remove "local" and "remote" as labels from HTTP bridge metrics
Because of the way most of the HTTP client works, every time an action is taken against the bridge, a new connection is open.
It happens, for example, for the readiness and liveness probes in Kubernetes but anyway can happen with any client (while others could leave the connection opened).
Currently, there are "local" and "remote" labels on each HTTP server metrics which specify the local and remote address and port for that request. The "local" is always the same while "remote" is different for each connection (at least it differs in the port part if the remote address is the same). This increases a lot the generated time series without any useful gain.
We should remove these labels from HTTP related metrics.

ppatierno:
This PR fixes #459 by removing "remote" and "local" labels from HTTP server related metrics.
It avoids a huge growth in the related time series samples on scraping.
I also tested this change on the Grafana dashboard we provide in Strimzi and there is no impact on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 0.26.0
+
+* Removed "remote" and "local" labels from HTTP server related metrics to avoid a big growth of time series samples.
+
 ## 0.25.0
 
 * Fixed printing operation log at the proper logging level.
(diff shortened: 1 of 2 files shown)


scholzj (approved):

ShubhamRwt (approved):

im-konge (approved):

ppatierno:
Going to merge, Travis failure doesn't depend on this change.

PR #785, opened 2023-04-13
Removed metrics counters for the /metrics endpoint itself

ppatierno:
The bridge provides metrics counters (related to HTTP requests) for all the HTTP endpoints including the `/metrics` endpoint which will be scraped by Prometheus to get the ...  metrics.
Imho having metrics for the `/metrics` endpoint isn't that useful.
This PR use a `MeterFilter` to not expose such counters.
In relation to getting rid of Vert.x and moving to use Quarkus, this is what the metrics support in Quarkus does as well. Metrics on the default `/q/metrics` endpoint are not exposed.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -8,7 +8,6 @@ package io.strimzi.kafka.bridge.http;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.config.MeterFilter;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.strimzi.kafka.bridge.Application;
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -178,6 +182,13 @@ public void start(Promise<Void> startPromise) {
                 this.router.errorHandler(HttpResponseStatus.NOT_FOUND.code(), this::errorHandler);
 
                 this.router.route("/metrics").handler(this::metricsHandler);
+                MeterRegistry meterRegistry = BackendRegistries.getDefaultNow();
```
ShubhamRwt, line 185: @ppatierno Aren't we doing this step already in `Application.java` class? Shouldn' we just directly use 
```
    if (metricsReporter.getMeterRegistry() != null) {
          ......
     }
```

scholzj (approved):

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: good catch @ShubhamRwt ! I have just pushed the change.

ShubhamRwt (approved):

ppatierno:
Travis failure doesn't depend on this change, going to merge anyway.

PR #800, opened 2023-05-15
Fixed HTTP status code for ready and healthy endpoints

ppatierno:
The you hit the `/ready` and `/healthy` endpoints, no content is returned and the body is empty.
This mean that the more appropriate HTTP status code has to be `204` and not `200` as we have today.
This PR fixes this issue.
It also adds the `404` code in the OpenAPI specification, already returned by the bridge (but missing in the spec).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Exposed the `/metrics` endpoint through the OpenAPI specification.
 * Fixed OpenAPI 3.0 `OffsetRecordSentList` component schema returning proper record offsets or error.
 * Fixed OpenAPI `ConsumerRecord` component schema returning key and value not only as (JSON) string but even as object.
+* Fixed OpenAPI HTTP status code returned by `/ready` and `/healthy`, from 200 to 204 because no content in the response.
 
 ## 0.25.0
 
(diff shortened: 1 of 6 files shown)


scholzj (commented):
If we are already changing the return codes ... is 404 really appropriate for not ready? The URL is correct and the answer is there. So _Not Found_ seems a bit weird. In the USer Operator for example we are using 500.

ppatierno:
> If we are already changing the return codes ... is 404 really appropriate for not ready? The URL is correct and the answer is there. So Not Found seems a bit weird. In the USer Operator for example we are using 500.

That was another option I had in mind. If we are using 500 for the UO, it's better to align with it then.

scholzj (approved):

ppatierno:
Current Java17 failure is about error on installing it in the Azure pipeline. Moving forward for now, by merging the PR.

scholzj:
> Current Java17 failure is about error on installing it in the Azure pipeline. Moving forward for now, by merging the PR.

Isn't the Azure pipeline the main CI for this? You should probably fix the Java installation to use the pre-installed Java instead of installing a custom one. We did that on the operators for example.

ppatierno:
yeah I opened this PR for it https://github.com/strimzi/strimzi-kafka-bridge/pull/802

PR #807, opened 2023-06-08
Setup right bootstrap strap server in the `ConsumerGeneratedNameIT` test

### LINKED ISSUE #747: The `ConsumerGeneratedNameIT` tests set up a wrong bootstrap server on the bridge
When running the `ConsumerGeneratedNameIT` tests, the configuration used for the bridge has the following hard coded bootstrap server:

```java
config.put(KafkaConfig.KAFKA_CONFIG_PREFIX + ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
```

This is wrong because the test starts a test-container which get a port and define listeners for Kafka and there is no `localhost:9092` but something like this:

```shell
2023-01-15T16:03:05.0909501Z [2023-01-15 16:03:04,466] INFO  <fkaContainer:161> [main        ] Mapped port: 32768
2023-01-15T16:03:05.0912073Z [2023-01-15 16:03:04,466] INFO  <fkaContainer:181> [main        ] This is all advertised listeners for Kafka PLAINTEXT://localhost:32768,BROKER1://172.18.0.2:9093
```

Anyway the admin client on the bridge is set up with the wrong hard coded bootstrap server:

```shell
2023-01-15T16:03:08.5646609Z [2023-01-15 16:03:08,564] INFO  <ClientConfig:376> [oop-thread-0] AdminClientConfig values: 
2023-01-15T16:03:08.5647472Z 	bootstrap.servers = [localhost:9092]
```

And it's not able to connect.

The tests pass anyway because they are not related on having a working connection from bridge to the Kafka cluster.
Anyway, we need to fix the set up which doesn't make sense.

ShubhamRwt:
This PR fixes #747. This PR aims to setup the Kafka Admin client with the correct bootstrap server coming from the test containers

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameIT.java
@@ -50,14 +50,23 @@ import static org.hamcrest.Matchers.is;
 @DisabledIfEnvironmentVariable(named = "EXTERNAL_BRIDGE", matches = "((?i)TRUE(?-i))")
 public class ConsumerGeneratedNameIT {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConsumerGeneratedNameIT.class);
-
+    static String kafkaUri;
     private static Map<String, Object> config = new HashMap<>();
     private static StrimziKafkaContainer kafkaContainer;
     private static final String BRIDGE_EXTERNAL_ENV = System.getenv().getOrDefault("EXTERNAL_BRIDGE", "FALSE");
     private static final String KAFKA_EXTERNAL_ENV = System.getenv().getOrDefault("EXTERNAL_KAFKA", "FALSE");
 
     static {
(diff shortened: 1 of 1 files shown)


scholzj (approved):

see-quick (approved):
LGTM. Thanks

src/test/java/io/strimzi/kafka/bridge/http/ConsumerGeneratedNameIT.java
```diff
@@ -50,14 +50,23 @@
...
+        if ("FALSE".equals(KAFKA_EXTERNAL_ENV)) {
+            kafkaContainer = new StrimziKafkaContainer();
+            kafkaContainer.start();
+            kafkaUri = kafkaContainer.getBootstrapServers();
+        } else {
+            // else use external kafka
+            kafkaUri = "localhost:9092";
+        }
+
+        config.put(KafkaConfig.KAFKA_CONFIG_PREFIX + ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaContainer.getBootstrapServers());
```
ppatierno, line 69: Not sure about this. You first set the right address into the `kafkaUri` and then use the `kafkaContaiber.getBootstrapServers()` to configure the admin client?

ppatierno (approved):

PR #823, opened 2023-07-20
Fixed java environment issue for s390x on Travis

dandotimujahid:


scholzj (commented):
Should the `jdk: openjdk11` be removed from the build matrix if we install the Java in the `before_install` build step?

ppatierno:
@dandotimujahid thanks for the PR but the `ppc64le` is still failing with the same error.

> Should the jdk: openjdk11 be removed from the build matrix if we install the Java in the before_install build step?

Anyway, I don't understand why we need to re-install the JDK if it is specified in the matrix so it should be already available there. Are we sure you are trying to fix something which is a Travis bug instead? We opened a support ticket but not got any answer yet.

dandotimujahid:
> Should the `jdk: openjdk11` be removed from the build matrix if we install Java in the `before_install` build step?

Yes, @scholzj  we can remove jdk: openjdk11 and maven from packages, from build matrix.

dandotimujahid:
> @dandotimujahid thanks for the PR but the `ppc64le` is still failing with the same error.
> 
> > Should the jdk: openjdk11 be removed from the build matrix if we install the Java in the before_install build step?
> 
> Anyway, I don't understand why we need to re-install the JDK if it is specified in the matrix so it should be already available there. Are we sure you are trying to fix something which is a Travis bug instead? We opened a support ticket but not got any answer yet.



> @dandotimujahid thanks for the PR but the `ppc64le` is still failing with the same error.
> 
> > Should the jdk: openjdk11 be removed from the build matrix if we install the Java in the before_install build step?
> 
> Anyway, I don't understand why we need to re-install the JDK if it is specified in the matrix so it should be already available there. Are we sure you are trying to fix something which is a Travis bug instead? We opened a support ticket but not got any answer yet.

The value of ${TRAVIS_CPU_ARCH} for ppc is 'ppc64le' but the directory name for java is 'ppc64el' hence it is creating the conflict for ppc. 

Is is Okay to add a hardcoded java path for ppc in its local env?

dandotimujahid:
@ppatierno, I tested the builds on Travis's environment with jdk-11 and maven being installed in the build matrix but still builds were failing with the same issue.  It worked after installing jdk and maven in before_install.

ppatierno:
This is the reason why I think it's more something which needs a fix on Travis side not on the project side.

dandotimujahid:
> This is the reason why I think it's more something which needs a fix on Travis side not on the project side.

Yes right, we will check with Travis team. Thanks.

dandotimujahid:
> did you get any insights from the Travis CI team

@ppatierno No response from Travis team, yet.

scholzj:
BTW: Did Travis fix something? It seemed to have passed fine in https://github.com/strimzi/strimzi-kafka-bridge/pull/824

ppatierno:
@scholzj it seems they fixed it see changelog https://changelog.travis-ci.com/
I also ran pipeline on `main` successfully.
@dandotimujahid if you are ok, we can close this PR.

dandotimujahid:
@ppatierno @scholzj, Thanks for the cooperation, The Issue is fixed by Travis team and the build has passed on main branch. We can close this PR.

PR #836, opened 2023-09-20
feat: added a new text embedded format

### LINKED ISSUE #739: Add a new `string` embedded format
Currently, we have the `binary` and `json` embedded format.
The JSON one is used for handling messages key and message as JSON, while the binary one allows to set them a Base64 encoded String from corresponding binary data.
As per #410, users need to transfer key and value which are simple raw strings and not JSON strings, so without quotes.
This is not possible with the `json` embedded format because it's not simple detect that a key or value is a simple String or just a malformed JSON (see #737 for more details).
Using the `binary` mode to transfer simple strings works but it's an hack.

In order to allow the simple strings use case, we could add a new `string` embedded format in the bridge.

antonio-pedro99:
This PR adds a new text embedded format.
This would fix #739


src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
     /**
      * Constructor
      *
-     * @param bridgeConfig the bridge configuration
-     * @param context the HTTP bridge context
-     * @param format the embedded format for consumed messages
-     * @param keyDeserializer key deserializer for consumed messages
+     * @param bridgeConfig      the bridge configuration
```
antonio-pedro99, line 78:
```suggestion
     * @param bridgeConfig  the bridge configuration
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
+     * @param context           the HTTP bridge context
```
antonio-pedro99, line 79:
```suggestion
     * @param context  the HTTP bridge context
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
+     * @param format            the embedded format for consumed messages
```
antonio-pedro99, line 80:
```suggestion
     * @param format  the embedded format for consumed messages
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
+     * @param keyDeserializer   key deserializer for consumed messages
```
antonio-pedro99, line 81:
```suggestion
     * @param keyDeserializer  key deserializer for consumed messages
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -111,8 +112,8 @@ public void close() {
      * Create a Kafka consumer
      *
      * @param routingContext the routing context
-     * @param bodyAsJson HTTP request body bringing consumer settings
-     * @param handler handler for the request
+     * @param bodyAsJson     HTTP request body bringing consumer settings
+     * @param handler        handler for the request
```
antonio-pedro99, line 116:
```suggestion
     * @param handler  handler for the request
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -528,7 +529,7 @@ public void doUnsubscribe(RoutingContext routingContext) {
     /**
      * Add a configuration parameter with key and value to the provided Properties bag
      *
-     * @param key key of the configuration parameter
+     * @param key   key of the configuration parameter
```
antonio-pedro99, line 532:
```suggestion
     * @param key  key of the configuration parameter
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
     /**
      * Constructor
      *
-     * @param bridgeConfig the bridge configuration
-     * @param context the HTTP bridge context
-     * @param format the embedded format for consumed messages
-     * @param keyDeserializer key deserializer for consumed messages
+     * @param bridgeConfig  the bridge configuration
```
antonio-pedro99, line 78:
```suggestion
     * @param bridgeConfig the bridge configuration
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
+     * @param context  the HTTP bridge context
```
antonio-pedro99, line 79:
```suggestion
     * @param context the HTTP bridge context
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
+     * @param format  the embedded format for consumed messages
```
antonio-pedro99, line 80:
```suggestion
     * @param format the embedded format for consumed messages
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -74,10 +75,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
+     * @param keyDeserializer  key deserializer for consumed messages
```
antonio-pedro99, line 81:
```suggestion
     * @param keyDeserializer key deserializer for consumed messages
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -111,8 +112,8 @@ public void close() {
      * Create a Kafka consumer
      *
      * @param routingContext the routing context
-     * @param bodyAsJson HTTP request body bringing consumer settings
-     * @param handler handler for the request
+     * @param bodyAsJson  HTTP request body bringing consumer settings
```
antonio-pedro99, line 115:
```suggestion
     * @param bodyAsJson HTTP request body bringing consumer settings
```

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -114,7 +115,7 @@ public void handle(RoutingContext routingContext, Handler<HttpBridgeEndpoint> ha
             }
             records = messageConverter.toKafkaRecords(topic, partition, routingContext.body().buffer().getByteBuf().array());
 
-            for (ProducerRecord<K, V> record :records)   {
+            for (ProducerRecord<K, V> record : records) {
```
antonio-pedro99, line 118:
```suggestion
            for (ProducerRecord<K, V> record :records)   {
```

ppatierno:
@antonio-pedro99 why are you suggesting changes reviewing your own PR instead of just making the changes in the PR itself? All these suggestions spam our inbox :-P

antonio-pedro99:
> @antonio-pedro99 why are you suggesting changes reviewing your own PR instead of just making the changes in the PR itself? All these suggestions spam our inbox :-P I also see they are about formatting on Javadoc. I would take them out, focusing on the gist of the issue imho.

my bad, sorry for that.

IntelliJ IDEA doing its job :) 
sure I will take it out.

antonio-pedro99:
@ppatierno I think this is the first step!

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
```diff
@@ -16,9 +16,16 @@ public class BridgeContentType {
     /** JSON encoding with BINARY embedded format */
     public static final String KAFKA_JSON_BINARY = "application/vnd.kafka.binary.v2+json";
 
+
+    /** JSON encoding with Plain text embedded format */
+    public static final String KAFKA_JSON_TEXT = "application/vnd.kafka.string.v2+json";
```
ppatierno, line 21: I think we agreed on using `text` instead of `string` on the issue.

ppatierno:
@antonio-pedro99 are you still working on this? I see it's draft and actually the converter does nothing :-)

@scholzj I am not sure we want to target this for 0.27.0.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99: Aren't they the same thing? I couldn't find any reference for `text`.

antonio-pedro99:
> @antonio-pedro99 are you still working on this? I see it's draft and actually the converter does nothing :-)
> 
> @scholzj I am not sure we want to target this for 0.27.0.

Yes, I am still on it. Will push some changes by the end of this week. 
Is the PR in the right direction?

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
ppatierno: We were discussing here to use `text` instead of `string` https://github.com/strimzi/strimzi-kafka-bridge/issues/739#issuecomment-1727337478 so it should be `application/vnd.kafka.text.v2+json`

ppatierno:
> Is the PR in the right direction?

Hard to say, the converted does nothing right now :-)

antonio-pedro99:
I meant,  apart from the converter implementation.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99: Yes, you are right. 
I just didn't know that `application/vnd.kafka.text.v2+json` works, as I didn't find any reading material online.

ppatierno: `application/vnd.kafka.text.v2+json` this is not something you could find online.
This is pretty custom like the others already in place and it is not a well defined standards in the HTTP land.

antonio-pedro99:
Hi @ppatierno could you send some examples of request body the user will send?

ppatierno:
You can just get an example from the official documentation.
 https://strimzi.io/docs/bridge/latest/#proc-producing-messages-from-bridge-topics-partitions-bridge

The content type would be the "text" one you are introducing but the "value" field in the record should be interpreted as string in your code and not as JSON string.

antonio-pedro99:
> You can just get an example from the official documentation. https://strimzi.io/docs/bridge/latest/#proc-producing-messages-from-bridge-topics-partitions-bridge
> 
> The content type would be the "text" one you are introducing but the "value" field in the record should be interpreted as string in your code and not as JSON string.

The value or the body?? Sorry it is confusing. 

Are we talking about this?
```
curl -X POST \
  http://localhost:8080/topics/bridge-quickstart-topic \
  -H 'content-type: application/vnd.kafka.json.v2+json' \
  -d 'my body here'
```

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99: > application/vnd.kafka.text.v2+json

It seems impossible, but I did not find anything on the internet tbh. That is why I was thinking it's an invalid content type. Anyway, now I am sure it is valid.

ppatierno: HTTP Content-Type (bringing the Media type) is something that you can define for your own application. There are some standard ones like `text/plain`, `application/json` and so on but it doesn't mean that you can't create your own.
The Strimzi bridge has its own content type. You can understand that because of the `vnd.` prefix which means to be vendor specific (this case Strimzi specific) and it defines a "Subtype".
Also the `+json` suffix specifies the structure of the media type to be always JSON.
So you will never find information about `application/vnd.kafka.text.v2+json` out of Strimzi project.
More details here: https://en.wikipedia.org/wiki/Media_type

ppatierno:
> The value or the body?? Sorry it is confusing.
Are we talking about this?

```shell
curl -X POST \
  http://localhost:8080/topics/bridge-quickstart-topic \
  -H 'content-type: application/vnd.kafka.text.v2+json' \
  -d 'my body here'
```

as I said before `+json` means that the media type will be always JSON, so your "my bode here" has to be a JSON.
And it's defined in the OpenAPI spec for the bridge.
The "value" field in that JSON has to be what you handle in the media type, in your case just a `text`.

antonio-pedro99:
> > The value or the body?? Sorry it is confusing.
> > Are we talking about this?
> 
> ```shell
> curl -X POST \
>   http://localhost:8080/topics/bridge-quickstart-topic \
>   -H 'content-type: application/vnd.kafka.text.v2+json' \
>   -d 'my body here'
> ```
> 
> as I said before `+json` means that the media type will be always JSON, so your "my bode here" has to be a JSON. And it's defined in the OpenAPI spec for the bridge. See the example here https://github.com/strimzi/strimzi-kafka-bridge/blob/main/src/main/resources/openapi.json#L1823 which is the same in the doc I shared before. The "value" field in that JSON has to be what you handle in the media type, in your case just a `text`.

Got it now. Should I care about the value field only? Or the key can also receive a `text`?

ppatierno:
> Got it now. Should I care about the value field only? Or the key can also receive a text?

what do you mean? the "value" field in the JSON will bring the text.

antonio-pedro99:
I mean
```
{
     "key": text,
     "value": text
}
```

ppatierno:
but "value" is bringing the value of the text, not sure what you mean by text there, so for example ...

```json
{
     "key": "my-key",
     "value": "my-value"
}
```

where "my-value" should be not handled as a JSON string (as it happens today) but as a string.
To be clear and explain better ... try to use the HTTP bridge as it is today and sending the above record using an HTTP client.
If you run on the other side a Kafka consumer (so not connected to the bridge but directly to Kafka), it will get `"my-value"` as value. Notice the quotes around the value. This is a JSON String.
What we want with the new `text` MIME-Type is that the Kafka consumer will get just `my-value`. Notice no quotes around the value. That is a plain string.

antonio-pedro99:
> where "my-value" should



> but "value" is bringing the value of the text, not sure what you mean by text there, so for example ...
> 
> ```json
> {
>      "key": "my-key",
>      "value": "my-value"
> }
> ```
> 
> where "my-value" should be not handled as a JSON string (as it happens today) but as a string. To be clear and explain better ... try to use the HTTP bridge as it is today and sending the above record using an HTTP client. If you run on the other side a Kafka consumer (so not connected to the bridge but directly to Kafka), it will get `"my-value"` as value. Notice the quotes around the value. This is a JSON String. What we want with the new `text` MIME-Type is that the Kafka consumer will get just `my-value`. Notice no quotes around the value. That is a plain string.

Thanks @ppatierno for explaining, it's now clearer. It's not about how we send the records, but in which format the consumers will receive the data produced.

ppatierno:
> It's not about how we send the records, but in which format the consumers will receive the data produced.

Well that's not true. What the consumer receives is what the producer sends, because Kafka itself doesn't alter payload.
And what the producer (in the bridge) sends depends on how the value in the HTTP request is "converted". So it's all about the producer anyway, not the consumer.

ppatierno:
@antonio-pedro99 do you have any news on this? still working on it?

antonio-pedro99:
> @antonio-pedro99 do you have any news on this? still working on it?

Yes :)

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java
antonio-pedro99, line 1: @ppatierno 
I have added a new embedded format, but the `HttpTextMessageConverter` is using the `HttpJsonMessageConverter` implementation.

For debugging purposes, the idea is to use the text embedded format, but the output should be as if we were using the `json embedded format`. But nothing is being retuned, not even an empty list `[]`.

To reproduce
[x] Create a new Consumer with `format` as `text`.
[x] subscribe the consumer to any topic
[x] produce messages to the topic
[x] retrieve the messages using the consumer: `no matter how many messages I produced, I just get no output`.

ppatierno: > [x] produce messages to the topic

Produce by still using HTTP bridge or a Kafka client?

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99, line 21: @ppatierno just having a look at this, I think it can be one of the reason.
while retrieving the messages, I was using `application/vnd.kafka.text.v2+json`.

ppatierno: so you should change here from `string` to `text` then.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java
antonio-pedro99: I am using HTTP Bridge to produce the messages.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99: Yes,I have changed that!
I am getting a new error now, 
``json
{"offsets":[{"partition":0,"offset":0},{"error_code":404,"message":"Topic bridge-quickstart-topic not present in metadata after 60000 ms."},{"partition":0,"offset":1}]}
```
I was already getting this, and I had to use kRafta instead of Zookeeper. Now even using kRafta, I am still facing the issue while producing messages!
any thought?

Comment (ppatierno) [reply src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java:21]: Maybe you should create the topic first? Anyway I think you have more problems with this PR ... I played with it and going to leave some feedback soon.

Comment (antonio-pedro99) [reply src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java:21]: > Maybe you should create the topic first? Anyway I think you have more problems with this PR ... I played with it and going to leave some feedback soon.

I have created the topic!

**Comment by ppatierno** [comment]
As the current status of the PR I see different issues.
I just cloned your branch and needed to apply some changes to make it "working" somehow:

As we said in a previous comment you need `public static final String KAFKA_JSON_TEXT = "application/vnd.kafka.text.v2+json";` in the `BridgeContentType` (using the `text` instead of the `string` part).


I don't think you are able to produce properly through the bridge as it stands in your PR. Your producer via HTTP should have `Content-Type: application/vnd.kafka.text.v2+json` in order to use the `text` as embedded type but if you do so now, you will get an exception as no supported type. You need a change in the `HttpBridge.contentTypeToFormat` adding:

```java
case BridgeContentType.KAFKA_JSON_TEXT:
                return EmbeddedFormat.TEXT;
```

Only this way the `text` becomes and acceptable type on producting.
I think right now you are still producing with json as embedded format.

Finally, I think saying you are not receiving is not right ... pay attention you could have get 406 NOT ACCEPTABLE from the bridge with an empty body of course. This is because you need to update the OpenAPI definition adding the following to the poll operation declaration:

```json
"application/vnd.kafka.text.v2+json": {
                                "schema": {
                                    "$ref": "#/components/schemas/ConsumerRecordList"
                                }
                             }
```

This is because the OpenAPI based server is refusing your consumer requests right now because it didn't recognize the above as a possible content type.

Try to fix the above, push to the PR and you can see you will be able to produce and consume but .... not in the right format because your text converter is still using too much JSON :-) So it will behave just like the JSON coverter.

antonio-pedro99:
> 
> Finally, I think saying you are not receiving is not right ... pay attention you could have get 406 NOT ACCEPTABLE from the bridge with an empty body of course. This is because you need to update the OpenAPI definition adding the following to the poll operation declaration:
> 
> ```json
> "application/vnd.kafka.text.v2+json": {
>                                 "schema": {
>                                     "$ref": "#/components/schemas/ConsumerRecordList"
>                                 }
>                              }
> ```
> 
> This is because the OpenAPI based server is refusing your consumer requests right now because it didn't recognize the above as a possible content type.

I just noticed this now.
When I use `text`, nothing was shown, but if you use `json` I was getting consumer embedded format mismatch.

antonio-pedro99:
> I don't think you are able to produce properly through the bridge as it stands in your PR. Your producer via HTTP should have `Content-Type: application/vnd.kafka.text.v2+json` in order to use the `text` as embedded type but if you do so now, you will get an exception as no supported type. You need a change in the `HttpBridge.contentTypeToFormat` adding:
> 
> ```java
> case BridgeContentType.KAFKA_JSON_TEXT:
>                 return EmbeddedFormat.TEXT;
> ```
> 
> Only this way the `text` becomes and acceptable type on producting. I think right now you are still producing with json as embedded format.

Good catch, seems that I forgot this.

antonio-pedro99:
@ppatierno thanks a lot for the guidance, it is working fine now.
I will add the new content type wherever necessary in the OpenAPI definition, and then implement the text converter.

ppatierno:
@antonio-pedro99 I have it already working but won't share the solution ;-)
Btw please pay attention to this ...
When a producer uses `json` or `text`, you should also see how they are received by a plain Kafka client (not via the HTTP client) because it really shows the difference between `json` and `text`:

```shell
"my-value2"
my-value2
```

The above is the output of a Kafka console consumer where .... `"my-value"` was sent by an HTTP producer using `json` format while `my-value2` was sent by an HTTP producer using `text`.

antonio-pedro99:
> @antonio-pedro99 I have it already working but won't share the solution ;-) Btw please pay attention to this ... When a producer uses `json` or `text`, you should also see how they are received by a plain Kafka client (not via the HTTP client) because it really shows the difference between `json` and `text`:
> 
> ```shell
> bin/kafka-console-consumer.sh --bootstrap-server localhost:9092 --topic my-topic --from-beginning
> "my-value2"
> my-value2
> ```
> 
> The above is the output of a Kafka console consumer where .... `"my-value"` was sent by an HTTP producer using `json` format while `my-value2` was sent by an HTTP producer using `text`.
> 
> The HTTP consumer output would be similar to the JSON one because in the end your putting a string in a JSON field but pay attention when you decode the "value" as text received by the producer. It won't be quoted so the JSON decode will fail ... so you don't need to have a JSON decode (and this is a suggestion :-P)

Hi @ppatierno, I have taken this into consideration.

I am not using any decoder, thus the HTTP consumer output is sometime giving "\"some value"\" and sometime “some value”.
On the another hand, from the consumer console, the output is what we expect.

I am not really sure if we want the backslashes `\`.
Anyway, I will push the changes.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99: @ppatierno this error persists. Seems that the data got produced, but I don't get the expected response from the HTTP bridge.

antonio-pedro99:
Some buildings are failing exactly because of what I said above. I added some tests.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java
```diff
@@ -0,0 +1,111 @@
...
+            jsonObject.put("partition", record.partition());
+            jsonObject.put("offset", record.offset());
+
+            ArrayNode headers = JsonUtils.createArrayNode();
+
+            for (Header kafkaHeader : record.headers()) {
+                ObjectNode header = JsonUtils.createObjectNode();
+
+                header.put("key", kafkaHeader.key());
+                header.put("value", new String(kafkaHeader.value()));
```
ppatierno, line 101: This should be as before ...

```java
header.put("value", DatatypeConverter.printBase64Binary(kafkaHeader.value()));
```

ppatierno (changes requested):

ppatierno:
> I am not using any decoder, thus the HTTP consumer output is sometime giving "\"some value"\" and sometime “some value”.
On the another hand, from the consumer console, the output is what we expect. I am not really sure if we want the backslashes \.

I am not sure from where you are getting the `\` which is something we don't want of course. I haven't got such a problem on my side. Can you provide a reproducer or at least explaining all the steps you are following? (how to produce, how you are consuming, and so on).

ppatierno:
@antonio-pedro99 you also need to update the openapiv2.json file. It's the old Swagger 2.0 format but we still support it.

ppatierno:
@antonio-pedro99 I think I have got the cause you are getting `\"` ... you are sending as `text` but receiving as `json` and it's normal. When a message is produced with `text`, it's expected to be consumed as `text` not as `json` otherwise the different converter is encoding differently.

antonio-pedro99:
> @antonio-pedro99 I think I have got the cause you are getting `\"` ... you are sending as `text` but receiving as `json` and it's normal. When a message is produced with `text`, it's expected to be consumed as `text` not as `json` otherwise the different converter is encoding differently.

Exactly what I was thinking over the weekend.

antonio-pedro99:
Another thing I could observe is that not all the output has unescaped backslashes. 
Take a look at this output:
```json
[
    {
        "topic": "bridge-quickstart-topic",
        "key": "my-key",
        "value": "This is some text again",
        "partition": 0,
        "offset": 0
    },
    {
        "topic": "bridge-quickstart-topic",
        "key": null,
        "value": "Hellow rold",
        "partition": 0,
        "offset": 1
    },
    {
        "topic": "bridge-quickstart-topic",
        "key": "\"my-key\"",
        "value": "\"sales-lead-0001\"",
        "partition": 0,
        "offset": 2
    },
    {
        "topic": "bridge-quickstart-topic",
        "key": null,
        "value": "\"sales-lead-0003\"",
        "partition": 0,
        "offset": 3
    },
    {
        "topic": "bridge-quickstart-topic",
        "key": "my-key",
        "value": "This is some text again",
        "partition": 0,
        "offset": 4
    }
]
```

ppatierno:
Again it depends how you send the messages by producing with text format or JSON format. If you have a reproducer that would be helpful otherwise I don't see any reasons for them, it's not happening on my side.

src/main/java/io/strimzi/kafka/bridge/BridgeContentType.java
antonio-pedro99: @ppatierno I finally got rid of this.
The problem happens if I try to produce messages to a partition that does not exist in the topic.

My topic is configured like

```
Topic: bridge-quickstart-topic	TopicId: MY83JIVaReuXx2-xvKdcWg	PartitionCount: 1	ReplicationFactor: 1	Configs: segment.bytes=1073741824
	Topic: bridge-quickstart-topic	Partition: 0	Leader: 1	Replicas: 1	Isr: 1
```
We have `partition:0`, if I try to produce message like

```json
  {
      "value": "sales-lead-0002",
      "partition": 1
  }
```
the following error will occur
```json
  {
            "error_code": 404,
            "message": "Topic bridge-quickstart-topic not present in metadata after 60000 ms."
 }
```

After changing to partition:0 or not mentioning the partition, the error fade away.

antonio-pedro99:
> Again it depends how you send the messages by producing with text format or JSON format. If you have a reproducer that would be helpful otherwise I don't see any reasons for them, it's not happening on my side.

Fr, it is not happing no longer from my side.

The same thing is happing with the ConsumerIT.java. While adding tests for the new embedded format, I was producing message like this
```java
  String sentBody = "Simple message as string";
        basicKafkaClient.sendJsonMessagesPlain(topic, 1, sentBody, 0, true);
```
And then try to read as `Text`, and of course I am getting the backslashes.

I am trying to use now `sendStringMessagesPlain`, but getting some time exceeded exceptions. I don't know why yet, but I am investigating further.

And also, I observed that there is no test for the binary format, and I am supposing it is because you weren't able to produce a binary message. Found this instead.
```
//        kafkaCluster.produce(topic, sentBody.getBytes(), 1, 0);
 // TODO: make client to producer binary data..
```

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -245,6 +245,7 @@ void sendBinaryMessageWithKey(VertxTestContext context) throws InterruptedExcept
         });
     }
 
+    @Disabled("Will be check in the next PR, this is just external tests for Bridge")
```
ppatierno, line 248: why?

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -274,7 +275,7 @@ void sendTextMessageWithKey(VertxTestContext context) throws InterruptedExceptio
                 new ByteArrayDeserializer(), new ByteArrayDeserializer());
         consumer.handler(record -> {
             context.verify(() -> {
-                assertThat(new String(record.value()), is(value));
+                assertThat(new String(record.value()).replace("\\", ""), is(value));
```
ppatierno, line 278: why?

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
antonio-pedro99: I was just checking the issue with backslashes :)

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
antonio-pedro99: I was trying to see if the build will pass, as it was not passing due to tests.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
antonio-pedro99: @ppatierno is it okay if we use a decoder in tests in order to get rid of `\`?

ppatierno: I don't see why you want to do this replace. There is no `\`.
The `text` converter is used and the test just works fine by asserting:

```java
assertThat(new String(record.value()), is(value));
```

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -245,6 +245,56 @@ void sendBinaryMessageWithKey(VertxTestContext context) throws InterruptedExcept
...
+            .sendJsonObject(root, verifyOK(context));
+
+        Properties consumerProperties = Consumer.fillDefaultProperties();
+        consumerProperties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaUri);
+
+        KafkaConsumer<byte[], byte[]> consumer = KafkaConsumer.create(vertx, consumerProperties,
+                new ByteArrayDeserializer(), new ByteArrayDeserializer());
+        consumer.handler(record -> {
+            context.verify(() -> {
+                assertThat(new String(record.value()).replace("\\", ""), is(value));
```
ppatierno, line 278:
```suggestion
                assertThat(new String(record.value()), is(value));
```

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
antonio-pedro99: As I said before, it was a trying. 

There is \ though. 

Did you write any test in your fork?

ppatierno: > There is \ though.

Nope. There is not. I am not receiving any `\' when running the test and it works fine. There is no need to use the `replace` because `\` is not in the received message.

> Did you write any test in your fork?

No I am not. The PR is yours :-)

antonio-pedro99: Why there is no test for the binary format?

ppatierno: I don't know, maybe was planned and not done, but it's not subject for this PR :-)

antonio-pedro99: > I don't know, maybe was planned and not done, but it's not subject for this PR :-)

Seems like it was not completed for some reason. 

The thing that is happening in all tests here is the same. As explained [here](https://github.com/strimzi/strimzi-kafka-bridge/pull/836#issuecomment-1869509917)

ppatierno: Tbh I am not following you. All tests are green in the main branch. Not sure you broke something on yours.

ppatierno: > In the main branch, the tests for the binary embedded format are ignored, if you manually run them, it will probably fail.

Ok, but it's totally unrelated to your PR and what you are trying to achieve. I don't follow why you are mentioning this more times.

antonio-pedro99: > > In the main branch, the tests for the binary embedded format are ignored, if you manually run them, it will probably fail.
> 
> Ok, but it's totally unrelated to your PR and what you are trying to achieve. I don't follow why you are mentioning this more times.

Yes, I know! I was just wondering why. 
Anyway, I had to open a fresh PR with all changes and tests now passing.
Please take a look :)

PR #837, opened 2023-09-28
Add container signing and SBOMs

scholzj:
This PR adds support for container signing and SBOM support to Kafka Bridge It is based on the same logic as similar changes done to the Operators and Drain Cleaner repos. It does also slight changes to the CI pipeline - to use Azure steps instead of scripts - in order to bring it closer to the other pipelines and make it easier to reuse the code.

_This should have no impact on regular developers / users._


.azure/templates/jobs/build_java.yaml
```diff
@@ -25,14 +25,14 @@ jobs:
     steps:
       # Get cached Maven repository
       - template: "../steps/maven_cache.yaml"
-      - template: '../steps/setup_java.yaml'
+      - template: '../steps/prerequisites/install_docker.yaml'
```
ppatierno, line 28: should this be `install_java.yaml`?

.azure/templates/steps/prerequisites/install_java.yaml
```diff
@@ -1,12 +1,11 @@
-# Step to setup JAVA on the agent
+# Step to configure JAVA on the agent
 parameters:
   - name: JDK_VERSION
-    default: '11'
-
+    default: '17'
```
ppatierno, line 4: aren't we still building bridge with Java 11 even if it can run on JVM 17?

.azure/templates/jobs/build_java.yaml
scholzj: Good catch. Fixed.

.azure/templates/steps/prerequisites/install_java.yaml
scholzj: It is basically just the default version that is not used anywhere. But I changed it to 11.

ppatierno (approved):
LGTM. Thanks!

PR #855, opened 2023-12-13
Trying to add retries on s390x tests failing

ppatierno:
Travis CI s390x pipeline only seems to have a lot of flakyness with tests failing.
This PR adds some retries to make it more automatic (instead of me re-run jobs more and more times).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ jobs:
     os: linux
     arch: s390x
     jdk: openjdk11
-    script: mvn install -Dmaven.javadoc.skip=true
+    script: mvn install -Dmaven.javadoc.skip=true -Dsurefire.rerunFailingTestsCount=10
     addons:
       apt:
         packages:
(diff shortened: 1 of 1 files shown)


Frawless (approved):
Maybe it would be great to put a note there and log an issue?

ppatierno:
> Maybe it would be great to put a note there and log an issue?

I added a comment. Not sure a GitHub issue is really needed. The PR describes what is it for. We don't need to track it imho.

ppatierno:
@Frawless s390x failed as usual but I don't see any retry. Any idea?

ppatierno:
hey @nirmannarang because you were the first one who contributed to s390x in Travis CI, are you able to investigate why tests always keep failing only on this platform and not the others? There is flakyness only on s390x and not on all the other platforms. Thanks!

Frawless:
@ppatierno exactly. Maybe 10 retries is a little bit more than we need, but as long as the tests takes several seconds it's fine from I guess.

 I agree with you, from my POV there is some race in the code that is only visible on s390x.

Frawless (approved):

PR #856, opened 2023-12-19
Added missing messaging semantic attributes on the consumer spans

ppatierno:
The current bridge implementation uses the Kafka producer interceptor way for injecting span on producing (by using the OpenTelemetry implementation) but it doesn't the same on the consumer side (because producer and consumers span are not linked the right way, as the OTel community is working to fix for the future).
For this reason we generate the Kafka consumer span manually but right now we are missing some "messaging" related smantic attributes which are added by the OTel interceptor on producer automatically.
This PR fixes this problem.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
+++ b/src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
@@ -92,6 +92,9 @@ class OpenTelemetryHandle implements TracingHandle {
     public <K, V> void handleRecordSpan(ConsumerRecord<K, V> record) {
         String operationName = record.topic() + " " + MessageOperation.RECEIVE;
         SpanBuilder spanBuilder = get().spanBuilder(operationName);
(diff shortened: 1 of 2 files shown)


src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -90,8 +91,11 @@ private SpanBuilder getSpanBuilder(RoutingContext routingContext, String operati
 
     @Override
     public <K, V> void handleRecordSpan(ConsumerRecord<K, V> record) {
-        String operationName = record.topic() + " " + MessageOperation.RECEIVE;
+        String operationName = record.topic() + " " + MessageOperation.RECEIVE.name().toLowerCase(Locale.ROOT);
```
scholzj, line 94: What is the motivation behind using `Locale.ROOT`? I do not think we use it in many other places.

ppatierno: It's the way the official interceptor applies the lowecase, I was just applying the same.

scholzj (approved):

PR #858, opened 2023-12-28
feat: added a new text embedded format

### LINKED ISSUE #739: Add a new `string` embedded format

antonio-pedro99:
This fixes #739 

Note: I closed #836 and started a fresh PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Dependency updates (Kafka 3.7.0, Kubernetes configuration provider 1.1.2, Vert.x 4.5.4, Netty 4.1.107.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.3)
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
-* Introduced a new text embedded format to send and receive plain for record key and value.
+* Introduced a new text embedded format to send and receive plain strings for record key and value.
 
 ## 0.27.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Dependency updates (Kafka 3.7.0, Kubernetes configuration provider 1.1.2, Vert.x 4.5.4, Netty 4.1.107.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.3)
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
+* Introduced a new text embedded format to send and receive plain for record key and value.
 
 ## 0.27.0
 
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -629,7 +629,7 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
             case TEXT:
                 return BridgeContentType.KAFKA_JSON_TEXT;
         }
-        return null;
+        throw new IllegalArgumentException();
     }
 
     private boolean checkAcceptedBody(String accept) {
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -306,9 +306,8 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                             BridgeContentType.KAFKA_JSON, JsonUtils.jsonToBytes(error.toJson()));
                 } else {
                     responseStatus = HttpResponseStatus.OK;
-                    HttpUtils.sendResponse(routingContext, responseStatus.code(),
-                            this.format == EmbeddedFormat.BINARY ? BridgeContentType.KAFKA_JSON_BINARY : BridgeContentType.KAFKA_JSON_JSON,
-                            buffer);
+
+                    HttpUtils.sendResponse(routingContext, responseStatus.code(), getContentType(), buffer);
                 }
             } catch (JsonDecodeException e) {
                 LOGGER.error("Error decoding records as JSON", e);
@@ -617,6 +616,18 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpBinaryMessageConverter();
             case TEXT:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpTextMessageConverter();
+        }
+        return null;
+    }
+
+    private String getContentType() {
+        switch (this.format) {
+            case JSON:
+                return BridgeContentType.KAFKA_JSON_JSON;
+            case BINARY:
+                return BridgeContentType.KAFKA_JSON_BINARY;
+            case TEXT:
+                return BridgeContentType.KAFKA_JSON_TEXT;
         }
         return null;
     }
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
@@ -20,6 +20,7 @@ import io.strimzi.kafka.bridge.config.BridgeConfig;
 import io.strimzi.kafka.bridge.converter.MessageConverter;
 import io.strimzi.kafka.bridge.http.converter.HttpBinaryMessageConverter;
 import io.strimzi.kafka.bridge.http.converter.HttpJsonMessageConverter;
+import io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter;
 import io.strimzi.kafka.bridge.http.converter.JsonDecodeException;
 import io.strimzi.kafka.bridge.http.converter.JsonUtils;
 import io.strimzi.kafka.bridge.http.model.HttpBridgeError;
@@ -614,6 +615,8 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpJsonMessageConverter();
             case BINARY:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpBinaryMessageConverter();
+            case TEXT:
+                return (MessageConverter<K, V, byte[], byte[]>) new HttpTextMessageConverter();
         }
         return null;
     }
@@ -624,6 +627,8 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                 return format == EmbeddedFormat.JSON;
             case BridgeContentType.KAFKA_JSON_BINARY:
                 return format == EmbeddedFormat.BINARY;
+            case BridgeContentType.KAFKA_JSON_TEXT:
+                return format == EmbeddedFormat.TEXT;
         }
         return false;
     }
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -229,7 +229,7 @@ public class HttpSourceBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpJsonMessageConverter();
             case BridgeContentType.KAFKA_JSON_BINARY:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpBinaryMessageConverter();
-            case BridgeContentType.KAFKA_JSON_TEXT:
+            case TEXT:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpTextMessageConverter();
         }
         return null;
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -229,7 +229,7 @@ public class HttpSourceBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpJsonMessageConverter();
             case BridgeContentType.KAFKA_JSON_BINARY:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpBinaryMessageConverter();
-            case TEXT:
+            case BridgeContentType.KAFKA_JSON_TEXT:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpTextMessageConverter();
         }
         return null;
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -16,6 +16,7 @@ import io.strimzi.kafka.bridge.config.BridgeConfig;
 import io.strimzi.kafka.bridge.converter.MessageConverter;
 import io.strimzi.kafka.bridge.http.converter.HttpBinaryMessageConverter;
 import io.strimzi.kafka.bridge.http.converter.HttpJsonMessageConverter;
+import io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter;
 import io.strimzi.kafka.bridge.http.converter.JsonUtils;
 import io.strimzi.kafka.bridge.http.model.HttpBridgeError;
 import io.strimzi.kafka.bridge.http.model.HttpBridgeResult;
@@ -228,6 +229,8 @@ public class HttpSourceBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpJsonMessageConverter();
             case BridgeContentType.KAFKA_JSON_BINARY:
                 return (MessageConverter<K, V, byte[], byte[]>) new HttpBinaryMessageConverter();
+            case TEXT:
+                return (MessageConverter<K, V, byte[], byte[]>) new HttpTextMessageConverter();
         }
         return null;
     }
--- a/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
@@ -509,6 +509,38 @@ public class ProducerIT extends HttpBridgeITAbstract {
     }
 
     @Test
+    void sendTextMessageWithWrongValue(VertxTestContext context) throws InterruptedException, ExecutionException {
+        KafkaFuture<Void> future = adminClientFacade.createTopic(topic);
+
+        JsonObject value = new JsonObject().put("message", "Hi, this is a Kafka Bridge");
+
+        JsonArray records = new JsonArray();
+        JsonObject json = new JsonObject();
+        json.put("value", value);
+        records.add(json);
+
+        JsonObject root = new JsonObject();
+        root.put("records", records);
+
+        future.get();
+
+        // produce and check the status code
+        producerService()
+                .sendRecordsRequest(topic, root, BridgeContentType.KAFKA_JSON_TEXT)
+                .sendJsonObject(root, ar -> {
+                    context.verify(() -> {
+                        assertThat(ar.succeeded(), is(true));
+                        HttpResponse<JsonObject> response = ar.result();
+                        HttpBridgeError error = HttpBridgeError.fromJson(response.body());
+                        assertThat(response.statusCode(), is(HttpResponseStatus.UNPROCESSABLE_ENTITY.code()));
+                        assertThat(error.getCode(), is(HttpResponseStatus.UNPROCESSABLE_ENTITY.code()));
+                        assertThat(error.getMessage(), is("Because the embedded format is 'text', the value must be a string"));
+                    });
+                    context.completeNow();
+                });
+    }
+
+    @Test
     void sendMessageWithNullValueTest(VertxTestContext context) throws InterruptedException, ExecutionException {
         String topic = "sendMessageWithNullValueTest";
         KafkaFuture<Void> future = adminClientFacade.createTopic(topic);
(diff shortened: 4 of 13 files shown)


antonio-pedro99:
@ppatierno all tests are now passing.
I wonder why the `Docs build` is failing as I did not change anything related to docs!

ppatierno:
@antonio-pedro99 build is failing on doc with following error ...

```shell
M	documentation/book/api/paths.adoc
  make docu_api \
```

antonio-pedro99:
> @antonio-pedro99 build is failing on doc with following error ...
> 
> ```shell
> ERROR: Uncommitted changes in documentation:
> M	documentation/book/api/paths.adoc
> Run the following to add up-to-date resources:
>   make docu_api \
>     && git add documentation/book/ \
>     && git commit -s -m 'Update generated documentation'
> ```

`[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  4.118 s
[INFO] Finished at: 2023-12-29T20:49:05+05:30
[INFO] ------------------------------------------------------------------------
[WARNING] 
[WARNING] Plugin validation issues were detected in 1 plugin(s)
[WARNING] 
[WARNING]  * io.github.swagger2markup:swagger2markup-maven-plugin:1.3.7
[WARNING] 
[WARNING] For more or less details, use 'maven.plugin.validation' property with one of the values (case insensitive): [BRIEF, DEFAULT, VERBOSE]
[WARNING] 
[ERROR] Failed to execute goal io.github.swagger2markup:swagger2markup-maven-plugin:1.3.7:convertSwagger2markup (generate-apidoc) on project kafka-bridge: Failed to execute goal 'convertSwagger2markup': Error creating extended parser class: Could not determine whether class 'org.pegdown.Parser$$parboiled' has already been loaded: Unable to make protected final java.lang.Class java.lang.ClassLoader.findLoadedClass(java.lang.String) accessible: module java.base does not "opens java.lang" to unnamed module @358ab600 -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
make: *** [Makefile:63: docu_api] Error 1
`

ppatierno:
Just ran on your branch locally the command from Azure pipeline ...

```shell
mvn -e -V -B -P apidoc io.github.swagger2markup:swagger2markup-maven-plugin:convertSwagger2markup@generate-apidoc
```

and it works like a charm. It produces a `documentation/book/api/paths.adoc` file which you should push.

antonio-pedro99:
> Just ran on your branch locally the command from Azure pipeline ...
> 
> ```shell
> mvn -e -V -B -P apidoc io.github.swagger2markup:swagger2markup-maven-plugin:convertSwagger2markup@generate-apidoc
> ```
> 
> and it works like a charm. It produces a `documentation/book/api/paths.adoc` file which you should push.

I need to double-check, it still does not work on my side

ppatierno:
@antonio-pedro99 I have just pushed on your branch with the missing file. Please update your local branch checking out the changes.

antonio-pedro99:
> @antonio-pedro99 I have just pushed on your branch with the missing file. Please update your local branch checking out the changes.

work done!
Waiting for more reviews

scholzj (commented):
@antonio-pedro99 Thanks for the PR. I wonder how the conversion / error handling works here. I guess it either does not work that well, or it should be documented how does it (not) work.

For example, when I sent a following produce request, where the format is text but the value is JSON, it seems to pass fine:
```
curl -v -X POST \
    192.168.1.220:8080/topics/kafka-test-apps \
    -H 'content-type: application/vnd.kafka.text.v2+json' \
    -d '{
      "records": [
          {
              "key": "tricky-payload",
              "value": {"Ahoj": "Svete"}
          }
      ]
  }'
```

But when I try to consume it, I get empty value:
```
curl -X GET http://192.168.1.220:8080/consumers/my-bridge-group/instances/my-http-client/records \
  -H 'accept: application/vnd.kafka.text.v2+json'
[
  {
    "topic": "kafka-test-apps",
    "key": "tricky-payload",
    "value": "",
    "partition": 0,
    "offset": 69
  }
]
```

This seems to be the issue with the producer here where the bridge decodes the JSON as empty string. When I produce it as JSON and consume as text, I get the more expected `"value": "{\"Ahoj\":\"Svete\"}"`. Not sure this is a bug or expected. But if ti is expected, I thik it is worth documenting how to produce the messages.

The response seems to have also a wrong content type `application/vnd.kafka.json.v2+json`. That looks like something you have to fix in `HttpSink BridgeEndpoint.pollHandler`.

antonio-pedro99:
> @antonio-pedro99 Thanks for the PR. I wonder how the conversion / error handling works here. I guess it either does not work that well, or it should be documented how does it (not) work.
> 
> For example, when I sent a following produce request, where the format is text but the value is JSON, it seems to pass fine:
> 
> ```
> curl -v -X POST \
>     192.168.1.220:8080/topics/kafka-test-apps \
>     -H 'content-type: application/vnd.kafka.text.v2+json' \
>     -d '{
>       "records": [
>           {
>               "key": "tricky-payload",
>               "value": {"Ahoj": "Svete"}
>           }
>       ]
>   }'
> ```
> 
> But when I try to consume it, I get empty value:
> 
> ```
> curl -X GET http://192.168.1.220:8080/consumers/my-bridge-group/instances/my-http-client/records \
>   -H 'accept: application/vnd.kafka.text.v2+json'
> [
>   {
>     "topic": "kafka-test-apps",
>     "key": "tricky-payload",
>     "value": "",
>     "partition": 0,
>     "offset": 69
>   }
> ]
> ```
>
> This seems to be the issue with the producer here where the bridge decodes the JSON as empty string. When I produce it as JSON and consume as text, I get the more expected `"value": "{\"Ahoj\":\"Svete\"}"`. Not sure this is a bug or expected. But if ti is expected, I thik it is worth documenting how to produce the messages.

That is a good catch, tbh I am not sure how it is supposed to be handled, I implemented the converter following our discussions on #739. I think it's worth to think about how to handle this case. Allow me some time to think a bit more.
 
> The response seems to have also a wrong content type `application/vnd.kafka.json.v2+json`. That looks like something you have to fix in `HttpSink BridgeEndpoint.pollHandler`.

This is fixed in the recent commit.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -615,6 +614,18 @@ private MessageConverter<K, V, byte[], byte[]> buildMessageConverter() {
...
+    private String getContentType() {
+        switch (this.format) {
+            case JSON:
+                return BridgeContentType.KAFKA_JSON_JSON;
+            case BINARY:
+                return BridgeContentType.KAFKA_JSON_BINARY;
+            case TEXT:
+                return BridgeContentType.KAFKA_JSON_TEXT;
+        }
+        return null;
```
ppatierno, line 626: This should be something like ... `throw new IllegalArgumentException();`. It should never happen but returning `null` doesn't even make any sense.

ppatierno (changes requested):

ppatierno:
I think what Jakub found is a bug and the way it should work is that you could produce using an embedded format but receiving with a different one by accepting that there could be some "differences" in what you get.
I mean, you could produce using JSON format and value being a JSON but then you should be able to consume using TEXT format and getting the value as a text not a json (in the end the json value is a .... text). Of course, taking into account the escape of quoting fields in the JSON value.

scholzj:
> I mean, you could produce using JSON format and value being a JSON but then you should be able to consume using TEXT format and getting the value as a text not a json (in the end the json value is a .... text). Of course, taking into account the escape of quoting fields in the JSON value.

I think that works fine in the consumer IIRC what I was trying. It is more what happens the producer with `text` format sends a JSON where we can maybe throw an error or something.

ppatierno:
> It is more what happens the producer with text format sends a JSON where we can maybe throw an error or something.

Regarding this, because the producer is sending with "text" format and it's a JSON, why and how the bridge should know to validate it's a valid JSON. I mean, it's a plain text. Of course, it can contain a JSON but does it really matter for the bridge? It's just bringing text.

scholzj:
I guess you can check if the text is empty/null if there are any other
nodes? Anyway, as I said, maybe also just clarifying it in the docs would
be fine.

antonio-pedro99:
Hey @ppatierno @scholzj, just got busy with some work here. I will get back to you guys over the weekend.

ppatierno:
@antonio-pedro99 are you willing to address this soon?

antonio-pedro99:
> @antonio-pedro99 are you willing to address this soon?

Yes , I will!

antonio-pedro99:
Are you planning any release soon?

ppatierno:
> Are you planning any release soon?

No specific plan for now, I was just wondering if you were still interested in it because of 2 weeks after your last comment :-)

antonio-pedro99:
> > Are you planning any release soon?
> 
> No specific plan for now, I was just wondering if you were still interested in it because of 2 weeks after your last comment :-)

Yeah, I try to complete this on this weekend.

ppatierno:
@antonio-pedro99 any news?

antonio-pedro99:
Not yet :( 😞

It's been hard handling the time. Once there is an update,  I'll let you
know.

antonio-pedro99:
> > I mean, you could produce using JSON format and value being a JSON but then you should be able to consume using TEXT format and getting the value as a text not a json (in the end the json value is a .... text). Of course, taking into account the escape of quoting fields in the JSON value.
> 
> I think that works fine in the consumer IIRC what I was trying. It is more what happens the producer with `text` format sends a JSON where we can maybe throw an error or something.

what is consumer IIRC?

antonio-pedro99:
> > It is more what happens the producer with text format sends a JSON where we can maybe throw an error or something.
> 
> Regarding this, because the producer is sending with "text" format and it's a JSON, why and how the bridge should know to validate it's a valid JSON. I mean, it's a plain text. Of course, it can contain a JSON but does it really matter for the bridge? It's just bringing text.

So you suggest validating the JSON and make the possible conversion?

If the format is text, it makes more sense to expect `text` only.

scholzj:
Looks like the build failed on checkstyle:

```
[ERROR] /home/vsts/work/1/s/src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java:27:5: NPath Complexity is 217 (max allowed is 200). [NPathComplexity]
```

Maybe you can have a look if the code can be fixed - if not, you should be able to use annotation to override the warning.

ppatierno (changes requested):
Overall it looks goot to me but as we have `sendTextMessage` I think we should also have a test about what we discussed during the last community call. It's about testing that if the producer is sending a JSON, so not a proper text, in the "value", the client will get an HTTP response with an error.

antonio-pedro99:
> Overall it looks goot to me but as we have `sendTextMessage` I think we should also have a test about what we discussed during the last community call. It's about testing that if the producer is sending a JSON, so not a proper text, in the "value", the client will get an HTTP response with an error.

Yes, I was just thinking about that :)

scholzj:
> Yes, I was checking it. It's related to NPath, looks like there is a method that is exceeding the max allowed.
A way to fix it might be splitting the code into multiple private methods inside the new converter, but I don't like it as it will be very inconsistent with others converter's implementation.

Right, so as I said, you could use the annotation to just suppress the checkstyle for given method / class.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -777,6 +809,19 @@ Handler<AsyncResult<HttpResponse<JsonObject>>> verifyBadRequest(VertxTestContext
             });
     }
 
+    Handler<AsyncResult<HttpResponse<JsonObject>>> verifyUnprocessableEntity(VertxTestContext context, String message) {
```
ppatierno, line 812: I don't see this method being used, or? Can you remove it?

ppatierno:
@antonio-pedro99 can you check the build please, because it's failing due to a used deprecated method and we have checks about getting errors for this kind of things. Thanks!

antonio-pedro99:
> @antonio-pedro99 can you check the build please, because it's failing due to a used deprecated method and we have checks about getting errors for this kind of things. Thanks!

Hi Paolo, it seems that object Node has been deprecated, but we are still using it in other's converter implementation.

antonio-pedro99:
> @antonio-pedro99 It is the `put` method with JsonObject parameter that is deprecated I think. It says `Deprecated Since 2.4 use either set(String, JsonNode) or replace(String, JsonNode),`. So I think you need to fix only the once occurence of that. Not the other put calls that are not deprecated.

Yes, I changed accordingly.

antonio-pedro99:
> @antonio-pedro99 it's still failing but on a different line this time, same deprecation problem. Could you fix it please? Also, I would double check for other same deprecations before pushing otherwise this back and forth will continue :-) You have to fix the DCO as well. Thanks!

where can I see the line with the warning, please?

ppatierno:
@antonio-pedro99 just follow the "Details" link on the failed build named "kafka-bridge", you will get to this page https://github.com/strimzi/strimzi-kafka-bridge/pull/858/checks?check_run_id=22126447065 with 2 errors/2 warnings link where you can reach the failed builds and clicking on them you get the log.

ppatierno:
@antonio-pedro99 sorry but I would say that you should investigate more into it because the error is coming from the new test you added, so the issue is in the test itself or in your implementation. Run the test locally and check/debug. If it's ok maybe there is some flakyness due to race conditions somewhere? 

```shell
ProducerIT.lambda$sendTextMessageWithWrongValue$30:531->lambda$sendTextMessageWithWrongValue$29:534 » NullPointer
```

antonio-pedro99:
> @antonio-pedro99 sorry but I would say that you should investigate more into it because the error is coming from the new test you added, so the issue is in the test itself or in your implementation. Run the test locally and check/debug. If it's ok maybe there is some flakyness due to race conditions somewhere?
> 
> ```shell
> ProducerIT.lambda$sendTextMessageWithWrongValue$30:531->lambda$sendTextMessageWithWrongValue$29:534 » NullPointer
> ```

Running the test individually, it works, so I am assuming this might be related to race condition. Plus, it shows that the error is coming from 
```bash
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "io.vertx.core.json.JsonObject.getInteger(String)" is null

	at io.strimzi.kafka.bridge.http.model.HttpBridgeError.fromJson(HttpBridgeError.java:61)
	at io.strimzi.kafka.bridge.http.ProducerIT.lambda$sendTextMessageWithWrongValue$29(ProducerIT.java:534)
	at io.vertx.junit5.VertxTestContext.verify(VertxTestContext.java:376)
	at io.strimzi.kafka.bridge.http.ProducerIT.lambda$sendTextMessageWithWrongValue$30(ProducerIT.java:531)
```
This happening while invoking the `HttpBridgeError.fromJson(response.body());`

ppatierno:
I found the problem and it's an interesting bug we have in the bridge since ages I think. The problem was raised by your additional test because we are moving from a JSON producer to a TEXT producer.
The HTTP producer used for tests keeps the connection alive so the message converter (based on the format in the HTTP request) is created only when the first test start (which is JSON producer) and it's the same for all tests.
So when it comes to TEXT test, it's still using the JSON message converter and the exception is not raised (because the TEXT converter is not used at all). When running the test alone, of course the right message converter is used.
I will open an issue, and think about the right fix.

antonio-pedro99:
> I found the problem and it's an interesting subtle bug we have in the bridge since ages I think. The problem was raised by your additional test because we are moving from a JSON producer to a TEXT producer (when running more tests). Until know we had only JSON producer-based tests running. The HTTP producer used for tests keeps the connection alive so the message converter (based on the format in the HTTP request) is created only when the first test start (which is JSON producer) and it's the same for all tests. So when it comes to TEXT test, it's still using the JSON message converter and the exception is not raised (because the TEXT converter is not used at all). When running the test alone, of course the right message converter is used. I will open an issue, and think about the right fix.

I was trying this as well, I tried to change the order of the tests, but it still didn't work at all. 

Should we skip this test for now?

ppatierno:
> Should we skip this test for now?

We have two options. We can disable this test for now by using the `@Disabled` annotation if we want this feature merged for 0.28.0. Or we can wait for me to fix it, which could happen soon or not, and delaying this PR to be merged. @scholzj what's your opinion on it?

antonio-pedro99:
Because the test passes when ran individually,  I think it makes sense to just disable for now

ppatierno:
I could also have the bug fixed by tomorrow. I am already on it.

antonio-pedro99:
> I could also have the bug fixed by tomorrow. I am already on it.

The PR is almost 2 months old, waiting for few more days does not hurt 🤣🤣

ppatierno:
Thinking more ... I am in doubt it's an actual bug ... I mean.
When a consumer is created, for example, it specifies the format in the consumer creation API call and it stays the same for its entire life.
Shouldn't it be the same for a producer? Until the producer disconnect we could expect that it's going to send the same data format and not mixing them across different HTTP requests (on the same HTTP connection, with keep-alive).
So I was wondering if the problem is in the tests, where the producer should disconnect if it wants to send a different data format. @scholzj what's your take on this? wdyt?

ppatierno:
I decided to start a discussion not opening an issue https://github.com/strimzi/strimzi-kafka-bridge/discussions/873

ppatierno:
@antonio-pedro99 you need to rebate against main because I fixed the issue with the multiple HTTP requests with different Content-Type.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -244,6 +244,53 @@ void sendBinaryMessageWithKey(VertxTestContext context) throws InterruptedExcept
...
+
+        producerService()
+                .sendRecordsRequest(topic, root, BridgeContentType.KAFKA_JSON_TEXT)
+                .sendJsonObject(root, verifyOK(context));
+
+        Properties consumerProperties = Consumer.fillDefaultProperties();
+        consumerProperties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaUri);
+
+        KafkaConsumer<String, String> consumer = KafkaConsumer.create(vertx, consumerProperties,
+                new StringDeserializer(), new KafkaJsonDeserializer<>(String.class));
```
ppatierno, line 271:
```suggestion
                new StringDeserializer(), new StringDeserializer());
```

ppatierno:
@antonio-pedro99 I think you had problem with rebase. There are changes which are already in main, see all the "endpoint" classes.

src/main/java/io/strimzi/kafka/bridge/http/HttpAdminBridgeEndpoint.java
```diff
@@ -52,7 +52,7 @@ public class HttpAdminBridgeEndpoint extends HttpBridgeEndpoint {
      * @param context the HTTP bridge context
      */
     public HttpAdminBridgeEndpoint(BridgeConfig bridgeConfig, HttpBridgeContext context) {
-        super(bridgeConfig, null);
+        super(bridgeConfig);
```
ppatierno, line 55: this is already in main. This class is not involved at all in the text embedded format anyway.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridgeEndpoint.java
```diff
@@ -15,19 +14,16 @@
...
     private Handler<HttpBridgeEndpoint> closeHandler;
 
     /**
      * Constructor
      *
      * @param bridgeConfig the bridge configuration
-     * @param format the embedded format for consumed messages
      */
-    public HttpBridgeEndpoint(BridgeConfig bridgeConfig, EmbeddedFormat format) {
+    public HttpBridgeEndpoint(BridgeConfig bridgeConfig) {
```
ppatierno, line 25: this is already in main.

src/main/java/io/strimzi/kafka/bridge/http/HttpSinkBridgeEndpoint.java
```diff
@@ -85,9 +87,10 @@ public class HttpSinkBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
      */
     public HttpSinkBridgeEndpoint(BridgeConfig bridgeConfig, HttpBridgeContext<K, V> context, EmbeddedFormat format,
                                   Deserializer<K> keyDeserializer, Deserializer<V> valueDeserializer) {
-        super(bridgeConfig, format);
+        super(bridgeConfig);
         this.httpBridgeContext = context;
         this.kafkaBridgeConsumer = new KafkaBridgeConsumer<>(bridgeConfig.getKafkaConfig(), keyDeserializer, valueDeserializer);
+        this.format = format;
```
ppatierno, line 93: all changes around format are already in main.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -48,18 +49,18 @@ public class HttpSourceBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
...
-        super(bridgeConfig, format);
+        super(bridgeConfig);
         this.kafkaBridgeProducer = new KafkaBridgeProducer<>(bridgeConfig.getKafkaConfig(), keySerializer, valueSerializer);
     }
 
     @Override
     public void open() {
         this.name = this.bridgeConfig.getBridgeID() == null ? "kafka-bridge-producer-" + UUID.randomUUID() : this.bridgeConfig.getBridgeID() + "-" + UUID.randomUUID();
         this.closing = false;
-        this.messageConverter = this.buildMessageConverter();
```
ppatierno, line 62: All these changes are already in main.

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -106,6 +107,14 @@ public void handle(RoutingContext routingContext, Handler<HttpBridgeEndpoint> ha
         SpanHandle<K, V> span = tracing.span(routingContext, operationName);
 
         try {
+            String requestContentType = routingContext.request().getHeader("Content-Type") != null ?
+                    routingContext.request().getHeader("Content-Type") : BridgeContentType.KAFKA_JSON_BINARY;
+            // create a new message converter only if it's needed because the Content-Type is different from the previous request
+            if (!requestContentType.equals(contentType)) {
+                contentType = requestContentType;
+                messageConverter = this.buildMessageConverter(contentType);
+            }
```
ppatierno, line 116: These are already in main.

ppatierno (changes requested):

antonio-pedro99:
> @antonio-pedro99 I think you had problem with rebase. There are changes which are already in main, see all the "endpoint" classes.

Sorry, might be fixed now!

ppatierno (approved):
LGTM. Thanks for the PR!
@scholzj do you have anything to add?

scholzj (approved):
LGTM. But I guess we should have CHANGELOG record for this as well?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Dependency updates (Kafka 3.7.0, Kubernetes configuration provider 1.1.2, Vert.x 4.5.4, Netty 4.1.107.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.3)
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
+* Introduced a new text embedded format to send and receive plain for record key and value.
```
ppatierno, line 7: "plain" what? I guess "plain string"?

antonio-pedro99: plain string yes!

antonio-pedro99, line 7:
```suggestion
* Introduced a new text embedded format to send and receive plain strings for record key and value.
```

ppatierno:
@scholzj any further comments or can I merge this?

PR #864, opened 2024-01-31
Use Log4j2 instead of SLF4J for our own loggers

scholzj:
Right now, the bridge seems to be using SLF4J for logging. It seems to have also a mix of ways how the loggers are defined. 

This PR moves it to use Log4j2 and unifies the way the loggers are created (as a `static final` variables called `LOGGER`). The only exception is the logging of the HTTP request that seems to not use logger based on the class but rather based on the endpoint - this behavior is kept. These changes bring the Bridge more in-line with Strimzi operators and should make it easier to manage and test the logging dependencies. It also allows it to remove SLF4J as a direct dependency (but it remains a transitive dependency for things such as Kafka clients). This does change only the API used for logging. The actual logging was already before done with Log4j2.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,7 +100,6 @@
 		<maven.compiler.source>11</maven.compiler.source>
 		<maven.compiler.target>11</maven.compiler.target>
 		<log4j.version>2.17.2</log4j.version>
-		<slf4j.version>1.7.36</slf4j.version>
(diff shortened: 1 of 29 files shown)


ppatierno (approved):
LGTM. Assuming it's still logging :-)

PR #867, opened 2024-02-01
Update open telemetry dependencies and switch to using the built in JDK http client for reporting traces

SamBarker:
This PR removes the dependency on OkHttp and thus Kotlin in a fully backwards compatible fashion. It does this by replacing the `opentelemetry-exporter-sender-okhttp` (which is the OpenTelemetry default) with two separate senders. One for each transport. 

- `opentelemetry-exporter-sender-jdk` enables sending traces using `http/protobuf` which is the lowest common denominator option, as its the only method required by the OpenTelemetry specifications. 
- `opentelemetry-exporter-sender-grpc-managed-channel` enables support for sending traces using `grpc` and thus feature parity with the OkHttp sender.

In doing so it has brought in a newer version of the SemanticConventions library which has deprecated several of the used conventions. I have added the new fields where appropriate and suppressed the deprecation warnings as we need to propagate the changes to users.


pom.xml
```diff
@@ -126,8 +126,8 @@
 		<spotbugs.version>4.7.3</spotbugs.version>
 		<maven.spotbugs.version>4.7.3.0</maven.spotbugs.version>
 		<strimzi-oauth.version>0.14.0</strimzi-oauth.version>
-		<opentelemetry.alpha-version>1.19.0-alpha</opentelemetry.alpha-version>
-		<opentelemetry.version>1.19.0</opentelemetry.version>
+		<opentelemetry.version>1.34.1</opentelemetry.version>
+		<opentelemetry.instrumentation.version>1.32.0-alpha</opentelemetry.instrumentation.version>
```
ppatierno, line 130: why using two different versions and not aligning them?

pom.xml
```diff
@@ -138,8 +138,28 @@
...
+			</dependency>
+			<dependency>
+				<groupId>io.opentelemetry</groupId>
+				<artifactId>opentelemetry-bom-alpha</artifactId>
+				<version>${opentelemetry.version}-alpha</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
```
ppatierno, line 161: I am not for using bom. Overwriting dependencies for many reasons (need for a specific version, CVE fixes and so on) would be a nightmare by using a bom. A bom could override everything itself. To override you should include other boms (where the override could be done) and so on. Tbh I would just use the same way we have been using without boms. @scholzj what's your take on this?

pom.xml
SamBarker: AFICT there is nothing to be gained from aligning them (happy to be proven wrong) so went with latest and greatest for each.

pom.xml
SamBarker: I'm not sure I understand the objection, a dependency with a version tag in it will override what ever is imported from the BOM. 

The BOM represents the open telemetry projects tested set of dependencies, thus reducing the  reducing the potential for inter module dependency issues. 

I'd also note that looking at the set of dependencies managed by the Otel bom are not all at the same version (the majority are at the version specified by the bom import but not all)

SamBarker: Its also worth mentioning that otel themselves say we [should use](https://arc.net/l/quote/uuvuaxoz) the bom.

ppatierno: What if I have to override a transient OTel dependency here in the bridge? And maybe this dependency is coming from a bom imported by the OTel bom itself? I will end to import this other bom here to override the dependency. I still think we should avoid using bom. It has been working fine until today without using boms and with any issues. We have the freedom to change/override dependencies when needed, without any issues.

scholzj: BOM screw up other (transitive) dependencies in a weird ways and make it hard to override them. I would definitely avoid it as well.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -65,6 +65,7 @@ public String serviceName(BridgeConfig config) {
     @Override
     public void initialize() {
         System.setProperty("otel.metrics.exporter", "none"); // disable metrics
+        System.setProperty("otel.exporter.otlp.protocol", "http/protobuf"); // force the use of http/protobuf rather than GRPC this will (silently) override the OTEL_EXPORTER_OTLP_PROTOCOL environment variable
```
scholzj, line 68: What is the impact on existing users?

ppatierno: So currently users have to use a tracing backend exposing the OTLP related port with gRPC support (i.e. Jaeger backend has port 4317). Staying this way would not work, so it's a kind of braking change. The user has to expose the OTLP related port with HTTP support (i.e. Jaeger backend has port 4318).

Other question I have here is why forcing protobuf if the user would like to use JSON for example? The HTTP port also supports `http/json` as OTLP protocol, or? So if user doesn't set it, it would default to `grpc` which would be not supported by us, maybe we should "force" the user to set it, raising a warning/exception if not set? Or we could default to `http/protobuf` if not set.

scholzj: So I guess we should not do it this way? We should certainly not hardcode it into the code. But we probably first deprecate what we have and give users some options to migrate? It seems to be like this would essentially break at the time of upgrade.

ppatierno: To avoid having a Kotlin dependency through okhttp we need to replace it with the JDK http client which doesn't support gRPC so I don't see a way for deprecation in this scenario. The OpenTelemetry library can use okhttp or the JDK http client but not both at same time to allow users to easily migrate.
The other way would be taking okhttp for one more release but saying to move using http/protobuf as protocol instead grpc. But this solution would not allow us to get rid of Kotlin dependency.

ppatierno: @scholzj just to make my point clearer ... I see your point and I agree that a smooth migration with deprecation is better. To go this way we should first verify that okhttp really works when setting the OTLP protocol as `http/protobuf` instead of the default `grpc`. So if we go this way, we should hold on with this change and taking it for a future release (not the next one).

scholzj: Yeah, I guess if OkHttp supports both, we should stick with it for a few more releases and just instruct users to switch to the HTTP protocol as GRPC will not be supported in the future.

scholzj: >  To go this way we should first verify that okhttp really works when setting the OTLP protocol as http/protobuf instead of the default grpc.

I think you need to have the users control the switch as they also need to reconfigure ports, possibly their tracing backend etc.

ppatierno: > I think you need to have the users control the switch as they also need to reconfigure ports, possibly their tracing backend etc.

The users should be already able to do the switch by setting the `OTEL_EXPORTER_OTLP_PROTOCOL` env var (to `http/protobuf`) when starting the bridge. This is valid on bare metal / vm.  With the operator in place, the user could use the `ContainerTemplate` instead to set the env var during the migration phase. When grpc support is dropped we could "hard code" the env var set by the operator itself directly when configuring the `KafkaBridge` instance.

scholzj: > When grpc support is dropped we could "hard code" the env var set by the operator itself directly when configuring the KafkaBridge instance.

I guess defaulting would be nicer than hardcoding it. But yes, it seems like we can quite easily give them time to migrate to HTTP first and then we can drop OkHttp.

pom.xml
SamBarker: True, but it had backwards incompatible changes I didn't fancy reasoning about. I will align them when I remove the bom's

pom.xml
SamBarker: I've never noticed/realised I was having issues caused by using bom's. I'll remove them.

SamBarker:
With fresh eyes I realised there was a way to restore GRPC support (however that does require OpenTelemetry 1.34.0+). By adding dependencies on `opentelemetry-exporter-sender-grpc-managed-channel` and `grpc-netty-shaded` we can maintain GRPC support and remove the Kotlin dependency. 

The build is now failing due to deprecations in the `SemanticAttributes` some of them are simple renames however others involve changing the value e.g. `SemanticAttributes.HTTP_URL` becomes `SemanticAttributes.URL_FULL` which means its not a backwards compatible change. 

Obviously I'm happy to push up the simple renames however I'm looking for input from the maintainers for what to do about the non backwards compatible changes. The build fails due to using the deprecated properties but changing them breaks compatibility.

ppatierno:
> With fresh eyes I realised there was a way to restore GRPC support (however that does require OpenTelemetry 1.34.0+). By adding dependencies on opentelemetry-exporter-sender-grpc-managed-channel and grpc-netty-shaded we can maintain GRPC support and remove the Kotlin dependency.

Is it a documented way or an hack? Also I think that the deprecation issues makes things more complicated.

SamBarker:
Oh and as I forgot to mention it earlier the need for 1.34.0 rather than 1.28.0 where the sender was introduced was [this bug fix](https://github.com/open-telemetry/opentelemetry-java/pull/6110)

pom.xml
SamBarker: bom's removed in https://github.com/strimzi/strimzi-kafka-bridge/pull/867/commits/25df0af1df7f7ee1ff530e3a91e376f612c895cc

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -52,7 +52,9 @@ static void setCommonAttributes(SpanBuilder builder, RoutingContext routingConte
         builder.setAttribute(SemanticAttributes.HTTP_METHOD, routingContext.request().method().name()); // TODO remove in release after 0.28
         builder.setAttribute(SemanticAttributes.HTTP_REQUEST_METHOD, routingContext.request().method().name());
         builder.setAttribute(SemanticAttributes.HTTP_URL, routingContext.request().uri()); // TODO remove in release after 0.28
-        builder.setAttribute(SemanticAttributes.URL_FULL, routingContext.request().uri());
```
SamBarker: `url.full` is a client specified attribute the [server side specification](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions) breaks it up into component parts. However `url.full` is the documented replacement for `http.url`. So I'm taking the opportunity of deprecating `http.url` to improve the spec compliance.

pom.xml
ppatierno: About using 2.0 we have to pay more attention.
It seems there are a lot of breaking changes https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.0.0
The main ones impacting us would be:

* defaulting the OTLP protocol from `grpc` to `http/protobuf`. To keep backward compatibility on our side, it means we should "force" grpc in the bridge `otel.exporter.otlp.protocol=grpc` if the user doesn't set `OTEL_EXPORTER_OTLP_PROTOCOL` env var explicitely.
* we should explicitely disable the logs exporter `otel.logs.exporter=none` as we do for the metrics exporter already.

pom.xml
```diff
@@ -138,6 +139,8 @@
 		<!-- property to skip surefire tests during failsafe execution -->
 		<!--suppress UnresolvedMavenProperty -->
 		<skip.surefire.tests>${skipTests}</skip.surefire.tests>
+		<opentelemetry-semconv.version>1.21.0-alpha</opentelemetry-semconv.version>
```
ppatierno, line 142: can we move this definition together with all the others opentelemetry related above? Also it seems latest release is [1.23.1-alpha](https://github.com/open-telemetry/semantic-conventions-java/releases/tag/v1.23.1)

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -95,6 +95,21 @@ OTEL_EXPORTER_ZIPKIN_ENDPOINT=http://localhost:9411/api/v2/spans <2>
 <1> The name of the tracing system. In this example, Zipkin is specified.
 <2> The endpoint of the specific selected exporter that listens for spans. In this example, a Zipkin endpoint is specified.
 
+== Supported properties
```
ppatierno, line 98: I haven't had a section like this in the past. I am not sure we want it now otherwise we should list all span attributes (i.e. otel.status_code, span.kind, ....). I would take them out but of course taking what you have in the CHANGELOG mentioning new attributes and the ones deprecated and planned to be removed.

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
+* Removed the dependency on OkHttp and thus Kotlin.
+* Dependency update (OpenTelemetry Semconv from 1.19.0-alpha to 1.21.0-alpha)
+  * The following span attributes are being replaced replacements:
+    1. `http.method` is being replaced with `http.request.method`
+    2. `http.url` is being replaced with `url.scheme`, `url.path` & `url.query`
+    3. `messaging.destination` is being replaced with `messaging.destination.name`
+    4. `http.status_code` is being replaced with `http.response.status_code`
+  * The following span attributes are slated for removal:
+    1. `messaging.destination.kind` & `messaging.destination.kind=topic`
```
ppatierno, line 15: why it's mentioned twice? I mean the attribute name and then the same with the value which in our case will always be `topic`.

pom.xml
```diff
@@ -238,24 +241,52 @@
...
 			<artifactId>opentelemetry-exporter-otlp</artifactId>
 			<version>${opentelemetry.version}</version>
+			<exclusions>
+				<exclusion>
+					<groupId>io.opentelemetry</groupId>
+					<artifactId>opentelemetry-exporter-sender-okhttp</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>io.opentelemetry</groupId>
+					<artifactId>opentelemetry-extension-incubator</artifactId>
```
ppatierno, line 269: out of curiosity, why this one needs to be excluded?

pom.xml
```diff
@@ -238,24 +241,52 @@
...
+		</dependency>
+		<dependency>
+			<groupId>io.opentelemetry</groupId>
+			<artifactId>opentelemetry-exporter-sender-grpc-managed-channel</artifactId>
+			<version>${opentelemetry.version}</version>
+			<scope>runtime</scope>
+		</dependency>
+		<dependency>
+			<groupId>io.grpc</groupId>
+			<artifactId>grpc-netty-shaded</artifactId>
```
ppatierno, line 287: can you elaborate why this dependency is needed please?

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
SamBarker: I added them as it felt a bit random to just document them in the changelog with out mentioning them anywhere else. As much for discoverability as anything else. 

I only documented the bridge specific tags, which seems like a reasonable choice. We could add a sentence along the lines of `In addition to the open telemetry standard tags we include the following`.

CHANGELOG.md
SamBarker: I phrased it clumsily I was trying to capture that both the key and value types were being deprecated.

pom.xml
SamBarker: This is the dependency which provides the grpc transport that the GRPC sender uses instead of OkHttp.

pom.xml
SamBarker: Hmm good question, I copied from the infinispan change I'll dig a bit deeper.

pom.xml
SamBarker: I can bump to 2.0 if you would prefer.

documentation/modules/proc-configuring-kafka-bridge-tracing.adoc
```diff
@@ -95,6 +95,21 @@ OTEL_EXPORTER_ZIPKIN_ENDPOINT=http://localhost:9411/api/v2/spans <2>
 <1> The name of the tracing system. In this example, Zipkin is specified.
 <2> The endpoint of the specific selected exporter that listens for spans. In this example, a Zipkin endpoint is specified.
 
+== Supported Span attributes
+The Kafka Bridge adds, in addition to the standard Open Telemetry attributes, the following attributes from the https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions[Open Telemetry standard conventions for HTTP] to its spans.
```
ppatierno, line 99:
```suggestion

The Kafka Bridge adds, in addition to the standard OpenTelemetry attributes, the following attributes from the https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions[OpenTelemetry standard conventions for HTTP] to its spans.
```

ppatierno (commented):
@SamBarker I had another pass but I see some comments were not addressed.

pom.xml
SamBarker: I've removed the exclusion and it doesn't make a difference to the `dependency:tree` so I guess it doesn't apply in our context.

pom.xml
SamBarker: Do you want to move to 2.0 now or not? I've addressed the logs exporter point as that applied without moving to 2.0.0.

ppatierno: I would avoid moving to a 2.x version for now. It would be anyway only for the instrumented Kafka client but I would leave using a 1.3x version for now.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 * Dependency updates (Kafka 3.6.1, Kubernetes configuration provider 1.1.2, Vert.x 4.5.2, Netty 4.1.106.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.2)
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
+* Removed the dependency on OkHttp and thus Kotlin.
+* Dependency update (OpenTelemetry Semconv from 1.19.0-alpha to 1.21.0-alpha)
```
ppatierno, line 8: I guess we are missing all the other bumps you made in the pom.
Also it's enough just naming component - version so for example ...

```shell
Dependency update (OpenTelemetry Semconv 1.21.0-alpha, ....)
```

scholzj: This probably should also not have its own line in the changelog but be added to the existing dependency updates?

ppatierno: Yeah right.

SamBarker: I've added Semconv to the global dependency list, but thought it was worth calling out this particular change as it leads to the changes of the tags included on spans. 

I can re-phrase if that is desired but I think mentioning the changes is merited.

scholzj: If it is for whatever reason more important, you should probably rephrase it then:
* It doesn't make sense to have to separate entries for `Dependency update (...)`
* If this is more important for some reason, explain it there. If it changes some tags, make it clear there what and how.

You seem to have the list of changes already, so you can just say something like 
```
* OpenTelemetry Semconv update causes following incompatible changes:
    * ...
```

CHANGELOG.md
```diff
@@ -2,8 +2,16 @@
 
 ## 0.28.0
 
-* Dependency updates (Kafka 3.7.0, Kubernetes configuration provider 1.1.2, Vert.x 4.5.4, Netty 4.1.107.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.3)
+* Dependency updates (Kafka 3.7.0, Kubernetes configuration provider 1.1.2, Vert.x 4.5.4, Netty 4.1.107.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.3, OpenTelemetry 1.34.1, OpenTelemetry Semconv 1.21.0-alpha, OpenTelemetry instrumentation 1.32.0-alpha)
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
+* Removed the dependency on OkHttp and thus Kotlin.
+* Dependency update (OpenTelemetry Semconv from 1.19.0-alpha to 1.21.0-alpha)
+  * The following span attributes are being replaced replacements:
```
scholzj, line 9: `are being replaced replacements` seems like a typo.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -46,10 +46,15 @@ class OpenTelemetryHandle implements TracingHandle {
...
     private Tracer tracer;
 
+    @SuppressWarnings("deprecation")
     static void setCommonAttributes(SpanBuilder builder, RoutingContext routingContext) {
         builder.setAttribute(SemanticAttributes.PEER_SERVICE, KAFKA_SERVICE);
-        builder.setAttribute(SemanticAttributes.HTTP_METHOD, routingContext.request().method().name());
-        builder.setAttribute(SemanticAttributes.HTTP_URL, routingContext.request().uri());
+        builder.setAttribute(SemanticAttributes.HTTP_METHOD, routingContext.request().method().name()); // TODO remove in release after 0.28
+        builder.setAttribute(SemanticAttributes.HTTP_REQUEST_METHOD, routingContext.request().method().name());
+        builder.setAttribute(SemanticAttributes.HTTP_URL, routingContext.request().uri()); // TODO remove in release after 0.28
```
scholzj, line 54: The CHANGELOG suggests these are removed. Here you say remove after release 0.28. Why? If we keep them for backwards compatibility, that is great. But might make sense to give it few more releases. It should be also explained in the changelog properly what is being done.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -65,6 +70,7 @@ public String serviceName(BridgeConfig config) {
     @Override
     public void initialize() {
         System.setProperty("otel.metrics.exporter", "none"); // disable metrics
+        System.setProperty("otel.logs.exporter", "none"); // disable logs
```
scholzj, line 73: MAybe the comment can add some more explanation what logs and why are we disabling them?

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -89,12 +95,14 @@ private SpanBuilder getSpanBuilder(RoutingContext routingContext, String operati
         return spanBuilder;
     }
 
+    @SuppressWarnings("deprecation")
```
scholzj, line 98: A comment what is deprecated might be useful. Also, if something should be removed later, an issue would be nice for tracking and link it in the comments.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
SamBarker: I've updated the changelog to better reflect the change and reworded this comment to more accurately reflect the intent.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
SamBarker: I've added a comment explaining the suppressions and raised https://github.com/strimzi/strimzi-kafka-bridge/issues/875 to remove the deprecated attributes.

CHANGELOG.md
```diff
@@ -5,13 +5,12 @@
...
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
 * Removed the dependency on OkHttp and thus Kotlin.
-* Dependency update (OpenTelemetry Semconv from 1.19.0-alpha to 1.21.0-alpha)
-  * The following span attributes are being replaced replacements:
-    1. `http.method` is being replaced with `http.request.method`
-    2. `http.url` is being replaced with `url.scheme`, `url.path` & `url.query`
-    3. `messaging.destination` is being replaced with `messaging.destination.name`
-    4. `http.status_code` is being replaced with `http.response.status_code`
-  * The span attribute `messaging.destination.kind=topic` is slated for removal.
+* This release deprecates several attributes (inline with changes from OpenTelemetry Semconv) which it attaches to spans. Both the depricated attribute and its replacement will be added to spans in the current release. The depreciated attributes will be removed in a future version.
```
ppatierno, line 8:
```suggestion
* This release deprecates several attributes (inline with changes from OpenTelemetry Semconv) which it attaches to spans. Both the deprecated attribute and its replacement will be added to spans in the current release. The deprecated attributes will be removed in a future version.
```

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -70,7 +70,7 @@ public String serviceName(BridgeConfig config) {
     @Override
     public void initialize() {
         System.setProperty("otel.metrics.exporter", "none"); // disable metrics
-        System.setProperty("otel.logs.exporter", "none"); // disable logs
+        System.setProperty("otel.logs.exporter", "none"); // disable otel logs exporter. This has no effect on logging in the bridge.
```
ppatierno, line 73: I know it's not related to this PR but you expand the comment about "disable metrics" the same way for consistency please? Something like you have for logs, i.e. `// disable otel metrics exporter`. It sounds useless opening a different PR just for that. Thanks.

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
SamBarker: Resolved with https://github.com/strimzi/strimzi-kafka-bridge/pull/867/commits/8474fbb6d2b0cdbca3bee4237ca340a5702ee357

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
SamBarker: resolved with https://github.com/strimzi/strimzi-kafka-bridge/pull/867/commits/14e150d7a0104c235e2094a57afec185d535c19d

SamBarker:
@ppatierno @scholzj I believe everything has been addressed and the PR checks are happy modulo a PR approval  😉

ppatierno (approved):
LGTM. Thanks for the PR!

ppatierno:
I tested (again) this PR and it seems to work fine, even switching between using `grpc` (default) to `http/protobuf`.
@SamBarker could you fix the conflicts please? We want to validate that with the same versions (as you set on https://github.com/strimzi/strimzi-kafka-operator/pull/9758), OpenTelemetry works in the operator as well (KafkaConnect and KafkaMirrorMaker). After this check, I think we can move forward and merge both.

SamBarker:
> We want to validate that with the same versions (as you set on [strimzi/strimzi-kafka-operator#9758](https://github.com/strimzi/strimzi-kafka-operator/pull/9758)), OpenTelemetry works in the operator as well (KafkaConnect and KafkaMirrorMaker). After this check, I think we can move forward and merge both.

I'm not clear what you want me to do, if anything, for that. The versions are the same across both this and the operator PR and the operator regression suite passed. Though that ran with the version of the bridge with out these changes.

scholzj (approved):

ppatierno:
> I'm not clear what you want me to do, if anything, for that. The versions are the same across both this and the operator PR and the operator regression suite passed. Though that ran with the version of the bridge with out these changes.

Maybe it was a misleading comment but I didn't want nothing to do from you :-D
I meant to wait for the PR on the operator to work fine before merging this one, in order to validate that same OTel versions were working fine on both (operator and bridge).

PR #869, opened 2024-02-02
Mark sending Open Telemetry traces via GRPC as deprecated

SamBarker:
Following up on from the discussion from #867 

I have confirmed that the OkHTTP Exporter speaks `http/protobuf` however changing the port to match the the change in transport is important, failing to change the port results in

```
Feb 02, 2024 2:59:09 AM io.opentelemetry.sdk.internal.ThrottlingLogger doLog
SEVERE: Failed to export spans. The request could not be executed. Full error message: Connection reset
```

It doesn't affect the operations of the bridge just the telemetry.


src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -65,11 +68,26 @@ public String serviceName(BridgeConfig config) {
     @Override
     public void initialize() {
         System.setProperty("otel.metrics.exporter", "none"); // disable metrics
+        final String configuredOtlpProtocol = resolveConfiguredOtelProtocol();
+        if (configuredOtlpProtocol == null || configuredOtlpProtocol.equalsIgnoreCase("grpc")) {
+            LOGGER.warn("The HTTP-Kafka bridge is deprecating OTLP GRPC support in favour of plain HTTP transports. Please switch to http/protobuf and update collector addresses to match");
+        }
```
SamBarker, line 74: I was wondering if we should apply a default of `http/protobuf` here if nothing is set, but given we have to be configured with the appropriate port that sounds like a dubious choice.

SamBarker:
I've realised their is a better way of determining the existing protocol but have run out of time to push the update.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Dependency updates (Kafka 3.6.1, Kubernetes configuration provider 1.1.2, Vert.x 4.5.2, Netty 4.1.106.Final, Jackson FasterXML 2.16.1, Micrometer 1.12.2)
 * Fixed missing messaging semantic attributes to the Kafka consumer spans
+* Deprecating the use of GRPC in the Open Telemetry integration. Please switch to using `http/protobuf` by specifying `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`
```
ppatierno, line 7: I think this should be stated in the documentation as well. @PaulRMellor can we coordinate on this, when this PR will land?

scholzj, line 7:
```suggestion
* **The support of GRPC in the Open Telemetry integration is deprecated and will be removed in the future.**
  Please switch to using `http/protobuf` by specifying `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` and update the collector address.
```

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -65,11 +68,26 @@ public String serviceName(BridgeConfig config) {
     @Override
     public void initialize() {
         System.setProperty("otel.metrics.exporter", "none"); // disable metrics
+        final String configuredOtlpProtocol = resolveConfiguredOtelProtocol();
+        if (configuredOtlpProtocol == null || configuredOtlpProtocol.equalsIgnoreCase("grpc")) {
+            LOGGER.warn("The HTTP-Kafka bridge is deprecating OTLP GRPC support in favour of plain HTTP transports. Please switch to http/protobuf and update collector addresses to match");
```
scholzj, line 73:
```suggestion
            LOGGER.warn("The OTLP GRPC support in HTTP Bridge is deprecated in favour of plain HTTP transports. Please switch to http/protobuf and update collector addresses to match.");
```

src/main/java/io/strimzi/kafka/bridge/tracing/OpenTelemetryHandle.java
```diff
@@ -65,11 +68,26 @@ public String serviceName(BridgeConfig config) {
...
+    private static String resolveConfiguredOtelProtocol() {
+        final Map<String, String> envVars = System.getenv();
+        final String tracesProtocol = System.getProperty("otel.exporter.otlp.traces.protocol", envVars.get("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"));
+        //The traces protocol is more specific than the general protocol, so it should take prescience
+        if (tracesProtocol != null) {
+            return tracesProtocol;
+        } else {
+            return System.getProperty("otel.exporter.otlp.protocol", envVars.get("OTEL_EXPORTER_OTLP_PROTOCOL"));
+        }
+    }
```
scholzj, line 89: Please update the docs to explain the difference between those two and their expected and default values.

CHANGELOG.md
ppatierno: It should also be "OpenTelemetry" (no space).

ppatierno:
@SamBarker I was wondering if this PR makes sense alone or we could include these changes within the #867

SamBarker:
> @SamBarker I was wondering if this PR makes sense alone or we could include these changes within the https://github.com/strimzi/strimzi-kafka-bridge/pull/867

This PR only makes sense if we don't merge 867. If we merge 867 there is no deprecation.

ppatierno:
> This PR only makes sense if we don't merge 867. If we merge 867 there is no deprecation.

@SamBarker good point. You are right. So I think we should aim to have #867 merged soon.
@scholzj wdyt?

scholzj:
@ppatierno Honestly, I'm not entirely sure what the other PR does. The description is a mess - is the strikethrough text a formatting issue? Or what is its meaning? I think using the JDK client is preferable when possible, yes. But not sure what else does the PR do. We also need to make sure the same approach works everywhere - including the other operands. We do not want to deprecate GRPC and push users to move to HTTP for MM2 or Connect and stick with GRPC here if what is what the other PR does.

ppatierno:
@scholzj not sure what's not clear there but let me try to explain.
That PR is aiming to remove the okhttp dependency used by the OTLP exporter, because okhttp depends on Kotlin and we prefer not having such a dependency. The replacement for the okhttp library is using the JDK HTTP client which doesn't support gRPC out of box, but only HTTP. So, to avoid removing and deprecating gRPC (what this PR does), the OpenTelemetry community provides a way to plugin the "sender" you want to use for sending traces so ... we are adding a gRPC sender to be used together with the JDK HTTP client.
The result is we still support gRPC (together with HTTP) for sending traces so no issues for users.
The PR also updates some tracing tags related to new semantic convention from OpenTelemetry. It comes from updating the OTel dependencies and we need to update such dependencies because of the "gRPC sender" I mentioned above (which was added from version OTel 1.34.0).

scholzj:
@ppatierno Do not explain it here. It should be clear after reading the description of that PR. But at least to me it is not really clear what from that description applies, what is just some weird formatting error etc.

SamBarker:
We appear to be approaching [consensus](https://github.com/strimzi/strimzi-kafka-bridge/pull/867#issuecomment-1971035067 ) that #867 is a viable path forward and thus we don't need to deprecate GRPC and thus we do not need this PR. 

I will re-open and co-ordinate with @PaulRMellor if deprecating GRPC comes back to being the preferred choice.

PR #880, opened 2024-03-06
Fixed handling of wrong key when text embedded format is used by the producer

ppatierno:
The embedded format in the bridge is representative for both key and value.
With the new added `text` embedded format, there is check on the value only while it should be the same for the key.
Currently, if the producer uses `text` embedded format but sends a JSON key, the message is accepted but the consumer gets an empty key on the other side. This PR fixes the issue by rejecting the message having a wrongly formatted key (in JSON).


ppatierno:
This PR should be cherry-picked to the `release-0.28.x` branch as well.

PR #886, opened 2024-03-26
Add a test case for multiple content type for specific kafkatopic

see-quick:
This PR adds a test case where it uses binary and JSON content types within specific KafkaTopic.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
@@ -1203,9 +1203,8 @@ public class ProducerIT extends HttpBridgeITAbstract {
                 record.put("value", Base64.getEncoder().encodeToString(value.encode().getBytes()));
                 break;
             case BridgeContentType.KAFKA_JSON_TEXT:
-                // Convert JSON objects to string representations
-                record.put("key", key.encode());
-                record.put("value", value.encode());
+                record.put("key", "key"); // Use pure plain text for the key
+                record.put("value", "value"); // Use pure plain text for the value
                 break;
             default:
                 throw new RuntimeException("Un-supported content type:" + contentType);
--- a/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
@@ -39,6 +39,7 @@ import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.Properties;
+import java.util.Random;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -1142,10 +1143,7 @@ public class ProducerIT extends HttpBridgeITAbstract {
     @ParameterizedTest
     @MethodSource("contentTypeCombinations")
     void dynamicContentTypeHandling(String firstContentType, String secondContentType, VertxTestContext context) throws Exception {
-        final JsonObject key = new JsonObject().put("id", 123);
-        final JsonObject value = new JsonObject().put("id", 10).put("price", 150).put("description", "Hello world");
-
-        JsonObject record = createDynamicRecord(key, value, firstContentType);
+        JsonObject record = createDynamicRecord(firstContentType);
         JsonArray records = new JsonArray().add(record);
         JsonObject root = new JsonObject().put("records", records);
 
@@ -1166,7 +1164,7 @@ public class ProducerIT extends HttpBridgeITAbstract {
                 }
             });
 
-        record = createDynamicRecord(key, value, secondContentType);
+        record = createDynamicRecord(secondContentType);
         records = new JsonArray().add(record);
         root = new JsonObject().put("records", records);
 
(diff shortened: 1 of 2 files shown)


src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -1117,4 +1118,77 @@ void sendSimpleMessageWithWrongContentType(VertxTestContext context) throws Inte
...
+
+    /**
+     * Handles the response from sending a record to Kafka, logging the outcome and verifying the HTTP status code.
+     * This method is used as a callback for the asynchronous HTTP request to send records to Kafka.
+     *
+     * @param context       The VertxTestContext used for assertions and to mark the test as completed.
+     * @param ar            The result of the asynchronous HTTP request, containing either the response or an error.
+     * @param contentType   The content type of the message that was sent, used for logging purposes.
+     */
+    private void handleResponse(final VertxTestContext context, final AsyncResult<HttpResponse<JsonObject>> ar,
```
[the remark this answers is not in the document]
see-quick: Well, I don't want to refactor the whole test suite. I didn't want to make redundant code. Agree with changing the signature of the method (i.e., adding status code as additional parameter). I could create an issue to refactor it with such auxiliary method.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -1117,4 +1118,77 @@ void sendSimpleMessageWithWrongContentType(VertxTestContext context) throws Inte
...
+    /**
+     * Creates a JSON object representing records to send. Depending on the content type, the method adjusts the key and value
+     * by encoding them in base64 format for binary content.
+     *
+     * @param key           The key of the message
+     * @param value         The value of the message
+     * @param contentType   The content type of the message, determining whether the message is treated as JSON or binary.
+     * @return A JsonObject representing the records to send, formatted according to the specified content type.
+     */
+    private JsonObject createRecordsToSend(String key, String value, final String contentType) {
```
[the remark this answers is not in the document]
see-quick: > you are using String for key and value but takes into account that when content type is JSON they can be ... JSON. Of course a string including quotes is going to be a valid JSON on the other side. I was just wondering if we can really have just one method which would not work if I want to put a JSON object as key and/or value.

I think we can have such method. Do you think it makes sense to test it:
1. with `value` as simple string
2. with `value` as json object
?
and make this as `@Parametrized test`. Same test path basically.

> Also what about the newly added text embedded format?

Should I address it in this PR or we can do it in another one? :)

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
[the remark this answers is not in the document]
ppatierno: Tbh if you are planning to refactor all tests right after this PR I am ok, otherwise I am for removing this method and just duplicate the code as we have all across the other tests. Otherwise it will end to have this method here for months without refactoring all the other tests, there will be no value in the method in the end.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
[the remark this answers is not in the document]
ppatierno: well with `binary`, `json` and `text` it's not just about the value but also the key which can be ... binary, json or text. Doing with parametrized test makes sense but maybe parameterizing both?

> Should I address it in this PR or we can do it in another one? :)

Well. This PR is the one where we are trying to address the different embedded formats so I would do here.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -198,7 +203,7 @@ void sendSimpleMessageWithKey(VertxTestContext context) throws InterruptedExcept
     @Disabled("Will be check in the next PR, this is just external tests for Bridge")
     @DisabledIfEnvironmentVariable(named = "EXTERNAL_BRIDGE", matches = "((?i)TRUE(?-i))")
     @Test
-    void sendBinaryMessageWithKey(VertxTestContext context) throws InterruptedException, ExecutionException {
+    void sendBinaryMessageWithKey(VertxTestContext context) {
```
ppatierno, line 206: you removed the throws but the test is disabled. Any reasons? I mean, is this change related to the PR? Are you going to check and enable this test as well?

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -1117,4 +1122,98 @@ void sendSimpleMessageWithWrongContentType(VertxTestContext context) throws Inte
...
+        switch (contentType) {
+            case BridgeContentType.KAFKA_JSON_JSON:
+                // Both key and value are wrapped in a JsonObject
+                record.put("key", new JsonObject().put("actualKey", key));
+                record.put("value", new JsonObject().put("message", value));
+                break;
+            case BridgeContentType.KAFKA_JSON_BINARY:
+                // Both key and value are encoded in Base64
+                String base64Key = Base64.getEncoder().encodeToString(key.getBytes());
+                String base64Value = Base64.getEncoder().encodeToString(value.getBytes());
```
ppatierno, line 1204: do we need these two additional local vars?

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -1117,4 +1122,98 @@ void sendSimpleMessageWithWrongContentType(VertxTestContext context) throws Inte
...
+    }
+
+    private JsonObject createDynamicRecord(String key, String value, String contentType) {
+        JsonObject record = new JsonObject();
+
+        switch (contentType) {
+            case BridgeContentType.KAFKA_JSON_JSON:
+                // Both key and value are wrapped in a JsonObject
+                record.put("key", new JsonObject().put("actualKey", key));
+                record.put("value", new JsonObject().put("message", value));
```
ppatierno, line 1199: as I said in a previous comment you are passing JSON string not a full JSON object.
I mean, taking `value` as an example, you are passing something like `Hello, world!` which will be encoded as JSON string so including quotes like `"Hello, world!"`. But what if we want a test with `value` being something like this:

```json
{ "id": 10, "price": 150, "description": "Hello world" }
```

The above is a valid JSON value when embedded format `json` is used. The same is valid for key.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
see-quick: They were not used...so I have just removed them. I can rollback this change... )) because it's not related to this PR

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
see-quick: I think not will remove it

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
see-quick: Okay, I have updated to use `JSON` objects for all types and for `KAFKA_JSON_TEXT` just call `encode/decode` for string representation.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -1117,4 +1122,95 @@ void sendSimpleMessageWithWrongContentType(VertxTestContext context) throws Inte
...
+                record.put("value", value);
+                break;
+            case BridgeContentType.KAFKA_JSON_BINARY:
+                record.put("key", Base64.getEncoder().encodeToString(key.encode().getBytes()));
+                record.put("value", Base64.getEncoder().encodeToString(value.encode().getBytes()));
+                break;
+            case BridgeContentType.KAFKA_JSON_TEXT:
+                // Convert JSON objects to string representations
+                record.put("key", key.encode());
+                record.put("value", value.encode());
```
ppatierno, line 1208: Have you verified what this `encode` return when the `JsonObject` is just a string? I mean ... a JSON string includes quotes so for example the String "foo" will be `"foo"`, while a plain text will be just `foo`. I am worried that the `encode` is returning a JSON string which is not what we want, but a plain one.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -1117,4 +1122,94 @@ void sendSimpleMessageWithWrongContentType(VertxTestContext context) throws Inte
...
+                record.put("key", key);
+                record.put("value", value);
+                break;
+            case BridgeContentType.KAFKA_JSON_BINARY:
+                record.put("key", Base64.getEncoder().encodeToString(key.encode().getBytes()));
+                record.put("value", Base64.getEncoder().encodeToString(value.encode().getBytes()));
+                break;
+            case BridgeContentType.KAFKA_JSON_TEXT:
+                record.put("key", "key"); // Use pure plain text for the key
+                record.put("value", "value"); // Use pure plain text for the value
```
ppatierno, line 1207: but then it's hard coded this way? If you pass a string through a JsonObject, can't you get the raw value without the quotes? Asking ... needs to be verified.

see-quick: I can use 
```java
key.getString("id");
value.getString("description");
```
but that way it would be test-specific based on the JsonObject you provide. 

It would work in test case where we provide:
```java
final JsonObject key = new JsonObject().put("id", 123);
final JsonObject value = new JsonObject().put("id", 10).put("price", 150).put("description", "Hello world");
```
If you want this that way I am fine. But I didn't want to not rely on JSON objects inside the test...

ppatierno: I see that, but hard coded would not be better. In the end I see the same problem for the `binary` type.
You are encoding in Base64 the JSON objects anyway .... but in reality they can be anything ... they are just binary not specifically JSON (in the end you are getting bytes from JSON). I am starting to think we cannot have one generic method but we need several ones for handling: JSON object for `json`, just bytes as binary for `binary` and plain string for `text`.

ppatierno (approved):
LGTM. Thanks!

PR #887, opened 2024-03-26
Better look of logs within application or tests

see-quick:
This PR updates the logging of the `log4j2.properties` file. It is just a nuanced design change. 
Before:
image

After:
image


ppatierno:
@see-quick what about having the same for the bridge itself and not only for tests?

see-quick:
> @see-quick what about having the same for the bridge itself and not only for tests?

Sure, done :).

im-konge (approved):

ppatierno (approved):

PR #888, opened 2024-03-26
Removed embedded format leftover from source endpoint

ppatierno:
Trivial PR to remove an unused parameter from the source endpoint.


see-quick (approved):
LGTM, I think also we can remove:
```java
String contentType = httpServerRequest.getHeader("Content-Type") != null ?
                httpServerRequest.getHeader("Content-Type") : BridgeContentType.KAFKA_JSON_BINARY;
```
in HttpBridge.java cause such local variable will not be used anymore

im-konge (approved):

PR #895, opened 2024-04-17
fixed no schema matches error

### LINKED ISSUE #892: `No schema matches` error raised when key or value in a `json` record is a JSON array
When using the `json` embedded format and sending a message like the following ...

```json
{
  "records": [
    {
      "value": "my-value",
      "key": ["element1",{"field": "element2"}]
    }
  ]
}
```

The bridge returns ...

```json
{
    "error_code": 400,
    "message": "Validation error on: /records/0/key - No schema matches"
}
```

This error is raised by the internal Vert.x JSON schema component which is doing the validation on the `key` field based on the definition we have in the OpenAPI ...

```json
"key": {
    "oneOf": [
        {
            "type": "object"
        },
        {
            "type": "string"
        }
    ]
}
```

It should work because it's defined as a JSON object, and the array in the example is a valid JSON object.
So it seems that the array is making troubles to the validator.
The same happens if using the same to the `value` field of course.

antonio-pedro99:
This should fix #892.

I have added tests for both cases when the value or key is an array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.29.0
 
 * Dependency updates (Vert.x 4.5.7, Netty 4.1.108.Final)
-* Added support for records key and value to be JSON array when using the JSON embedded format.
+* Added support for records key and value to be JSON array when using the `json` embedded format.
 
 ## 0.28.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.29.0
 
 * Dependency updates (Vert.x 4.5.7, Netty 4.1.108.Final)
-* Fixed No schema matches error when using the `json` embedded format and the record key or value is an array.
+* Added support for records key and value to be JSON array when using the JSON embedded format.
 
 ## 0.28.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.29.0
 
 * Dependency updates (Vert.x 4.5.7, Netty 4.1.108.Final)
+* Fixed No schema matches error when using the `json` embedded format and the record key or value is an array.
 
 ## 0.28.0
 
--- a/src/main/resources/openapi.json
+++ b/src/main/resources/openapi.json
@@ -1495,6 +1495,9 @@
                     "key": {
                         "oneOf": [
                             {
+                                "type": "array"
+                            },
+                            {
                                 "type": "object"
                             },
                             {
@@ -1810,6 +1813,9 @@
                     "value": {
                         "oneOf": [
                             {
+                                "type": "array"
+                            },
+                            {
                                 "type": "object",
                                 "nullable": true
                             },
@@ -1821,6 +1827,9 @@
                     },
                     "key": {
                         "oneOf": [
+                            {
+                                "type": "array"
+                            },
                             {
                                 "type": "object"
                             },
@@ -1888,6 +1897,9 @@
                     "value": {
                         "oneOf": [
                             {
+                                "type": "array"
+                            },
+                            {
                                 "type": "object"
                             },
                             {
@@ -1897,6 +1909,9 @@
                     },
                     "key": {
                         "oneOf": [
+                            {
+                                "type": "array"
+                            },
                             {
                                 "type": "object"
                             },
--- a/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
@@ -201,6 +201,109 @@ public class ProducerIT extends HttpBridgeITAbstract {
         });
     }
 
+    @Test
+    void sendSimpleMessangeWithArrayKey(VertxTestContext context) throws InterruptedException, ExecutionException {
+        KafkaFuture<Void> future = adminClientFacade.createTopic(topic, 2, 1);
+
+        String value = "message-value";
+        JsonArray array = new JsonArray();
+        array.add("some-element").add(new JsonObject().put("some-field", "element-2"));
+
+        JsonArray records = new JsonArray();
+        JsonObject json = new JsonObject();
+        json.put("value", value);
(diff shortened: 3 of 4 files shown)


src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -201,6 +201,109 @@ void sendSimpleMessageWithKey(VertxTestContext context) throws InterruptedExcept
...
+        consumer.handler(record -> {
+            context.verify(() -> {
+                assertThat(record.value(), is(value));
+                assertThat(record.topic(), is(topic));
+                assertThat(record.partition(), notNullValue());
+                assertThat(record.offset(), is(0L));
+                record.headers().forEach(header -> {
+                    assertThat(header.key(), is("key"));
+                    assertThat(header.value().toString(), is(array.encode()));
+                });
```
ppatierno, line 241: I can't see any assert around the array you sent through the key field.

src/main/resources/openapi.json
```diff
@@ -1897,6 +1909,9 @@
                     },
                     "key": {
                         "oneOf": [
+                            {
+                                "type": "array"
+                            },
```
ppatierno, line 1914: All the above updates should be done in the openapiv2.json file as well.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -201,6 +201,109 @@ void sendSimpleMessageWithKey(VertxTestContext context) throws InterruptedExcept
...
+        consumer.handler(record -> {
+            context.verify(() -> {
+                record.headers().forEach(header -> {
+                    assertThat(header.key(), is("value"));
+                    assertThat(header.value().toString(), is(value.encode()));
+                });
+                assertThat(record.topic(), is(topic));
+                assertThat(record.partition(), notNullValue());
+                assertThat(record.offset(), is(0L));
+                assertThat(record.key(), nullValue());
```
ppatierno, line 292: as the previous one I don't see any check that value is the array you expected.

ppatierno (changes requested):

antonio-pedro99:
Hi @ppatierno, I will get back to you very soon with the updates

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -201,6 +201,103 @@ void sendSimpleMessageWithKey(VertxTestContext context) throws InterruptedExcept
...
+
+        KafkaConsumer<byte[], String> consumer = KafkaConsumer.create(vertx, consumerProperties,
+                new ByteArrayDeserializer(), new KafkaJsonDeserializer<>(String.class));
+        consumer.handler(record -> {
+            context.verify(() -> {
+                assertThat(record.value(), is(value));
+                assertThat(record.topic(), is(topic));
+                assertThat(record.partition(), notNullValue());
+                assertThat(record.offset(), is(0L));
+                assertThat(record.key(), is(key.toBuffer().getBytes()));
```
antonio-pedro99, line 238: @ppatierno I am not sure if this is the best way to assert the key(which is in bytes). 
Any suggestions?

ppatierno: The mistake here is totally different. Why did you delete the headers check? Kafka headers (with key-value) and key in the message are two totally different things. You should revert back the headers check part wherever you removed it.

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
antonio-pedro99: > The mistake here is totally different. Why did you delete the headers check? Kafka headers (with key-value) and key in the message are two totally different things. You should revert back the headers check part wherever you removed it.

Hmmm, I thought it would be better to follow other tests conventions. 
For example,  the `sendSimpleMessageWithKey(VertxTestContext context)` does not do any `header` check.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.29.0
 
 * Dependency updates (Vert.x 4.5.7, Netty 4.1.108.Final)
+* Fixed No schema matches error when using the `json` embedded format and the record key or value is an array.
```
ppatierno, line 6: I would reword and changing in terms of added support for key and value to be JSON array and not fixed bug imho.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
ppatierno: Yeah ok makes sense, even because we are not sending headers here as well.

ppatierno (changes requested):

CHANGELOG.md
antonio-pedro99: Maybe something like `Added support for key and value to be JSON array. Works with `json` embedded format.` would better?

ppatierno: Maybe a mix of that like "Added support for records key and value to be JSON array when using the `json` embedded format".

antonio-pedro99: Nice one!

antonio-pedro99, line 6:
```suggestion
* Added support for records key and value to be JSON array when using the json embedded format.
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.29.0
 
 * Dependency updates (Vert.x 4.5.7, Netty 4.1.108.Final)
+* Added support for records key and value to be JSON array when using the JSON embedded format.
```
ppatierno, line 6:
```suggestion
* Added support for records key and value to be JSON array when using the `json` embedded format.
```

ppatierno (changes requested):

scholzj (approved):

ppatierno (approved):

PR #896, opened 2024-04-19
Bumped Jackson version

ppatierno:
Trivial PR to bump Jackson version aligning with operator (via https://github.com/strimzi/strimzi-kafka-operator/pull/9999)


scholzj (approved):
I'm fine if you wanna update it. But I updated the Jakson version because of Fabric8 that is not used here. So I wonder if it is better to stick with 2.16.1 used by VErt.x in the bridge.

ppatierno:
> I'm fine if you wanna update it. But I updated the Jakson version because of Fabric8 that is not used here. So I wonder if it is better to stick with 2.16.1 used by VErt.x in the bridge.

No I missed that. It's better to stick with what Vert.x is bringing. I will update when Vert.x updates it as well. Thanks.

PR #899, opened 2024-05-01
docs(bridge): cleans up naming of kafka bridge

PaulRMellor:
**Documentation**

In a few places in the docs we say "Strimzi Kafka Bridge" without needing to make this distinction. These instances have been updated to say just "Kafka Bridge"

### DIFF
--- a/documentation/book/api/overview.adoc
+++ b/documentation/book/api/overview.adoc
@@ -1,9 +1,9 @@
-= Strimzi Kafka Bridge API Reference
+= Kafka Bridge API Reference
 
 
 [[_overview]]
 == Overview
-The Strimzi Kafka Bridge provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol.
+The Kafka Bridge provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol.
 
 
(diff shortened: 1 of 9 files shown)


documentation/book/api/overview.adoc
```diff
@@ -1,9 +1,9 @@
-= Strimzi Kafka Bridge API Reference
+= Kafka Bridge API Reference
 
 
 [[_overview]]
 == Overview
-The Strimzi Kafka Bridge provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol.
+The Kafka Bridge provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol.
```
scholzj, line 6: I guess this needs to be also changed in the OpenAPI JSONs here: https://github.com/strimzi/strimzi-kafka-bridge/tree/main/src/main/resources?

scholzj (approved):
I think you need to update the Open API JSONs to get the build passing. LGTM otherwise.

documentation/book/api/overview.adoc
ppatierno: Yeah, I would change both `title` and `description` on both the openapi JSONs.

PaulRMellor: Right. Thanks

ppatierno (approved):
LGTM

PR #909, opened 2024-05-31
Update the Prometheus JMX Collector and Client

scholzj:
This PR updates the Prometheus JMX Collector to 1.0.1. This new version has many changes to how it treats the metrics with some keywords such as `_total`. This required an update to the metrics configuration YAMl to maintain the compatibility of the produced metrics (essentially, the counters using `_total` suffix need to be mapped to counters in the rules).

In addition, the Prometheus Simple Client is now archived and not anymore supported by the Prometheus JMX Exporter. So it also moves the bridge to use the current Prometheus Client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 0.30.0
+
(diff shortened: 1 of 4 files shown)


ppatierno (approved):
LGTM. I assume you checked that metrics are coming as expected.

Issue #921, opened 2024-08-04
`path` label in Bridge metrics can contain very different values and that makes it hard to work with the metrics


scholzj:
The `path` label in the Bridge metrics seems to contain very different values. That breaks dashbords and makes it hard to work with the metrics. For example, for the poll metrics, you can get the metrics like this:

```
$ curl -s localhost:8080/metrics | grep strimzi_bridge_http_server_requests_total | grep GET | grep consumers
strimzi_bridge_http_server_requests_total{code="200",method="GET",path="/consumers/my-group5/instances/my-consumer5/records?timeout=100",} 432.0
strimzi_bridge_http_server_requests_total{code="200",method="GET",path="http://my-bridge-bridge-service:8080/consumers/my-group4/instances/my-consumer4/records?timeout=100",} 1035.0
```

As you can notice:
* The first metric contains the path parameters (`?timeout=100`)
* The second metric containers the full address (`http://my-bridge-bridge-service:8080`) as well as the path parameters (`?timeout=100`)

Both of these breaks the Strimzi Bridge Grafana Dashabord that expects the path to be always only `/consumers/<NAME>/instances/<NAME>/records`. So the dashboard does not show this metric. I noticed it at the [Poll chart](https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/examples/metrics/grafana-dashboards/strimzi-kafka-bridge.json#L2531) but this likely impacts also other metrics / charts.

scholzj:
Note: The two metrics above were produced by our own examples -> the first one is from the plain Java example and the other from the Vert.x example. So those can be used to reproduce this.

ppatierno:
I had a little bit of investigation and it seems that this metrics part is not working as expected since ... 0.20.0 release (using Vert.x 4.1.0) :-o.
The last working for me providing the right `path` tag is with 0.19.0 (using Vert.x 3.9.3).
I went through the Vert.x micrometer implementation and more specifically the part which provides the HTTP server request ones (because these are not metrics generated by us) in the `VertxHttpServerMetrics` class and currently they are using the `request.uri()` for the `path` tag:

https://github.com/vert-x3/vertx-micrometer-metrics/blob/4.5.9/src/main/java/io/vertx/micrometer/impl/VertxHttpServerMetrics.java#L79

Up to the latest 3.9.x (so 3.9.16) they were using the `request.path()` instead:

[...]

PR #922, opened 2024-08-05
Fixed wrong metrics "path" tag by using a custom tags provider

ppatierno:
This PR is a workaround for #921 to revert back having the HTTP resource path in the `path` metrics tag instead of the complete URI (even together with the host and query parameters) as used by latest Vert.x versions.
It's also related to the issue https://github.com/vert-x3/vertx-micrometer-metrics/issues/225 I opened on the Vert.x micrometer repo to get why they changed this way.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -6,6 +6,7 @@
 package io.strimzi.kafka.bridge;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
 import io.strimzi.kafka.bridge.config.BridgeConfig;
 import io.strimzi.kafka.bridge.config.ConfigRetriever;
 import io.strimzi.kafka.bridge.http.HttpBridge;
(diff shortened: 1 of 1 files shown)


scholzj (approved):
The code looks good to me. But ... given we have the bug there since 0.20 -> should we wait for the Vert.x response and if they did it intentionally and not just as a but should we instead adjust our dashboards to Vert.x solution?

ppatierno:
Yeah, I raised the PR just to make people aware of what was going to be a potential workaround.
We can wait for them of course but tbh if they are not going to fix that I am not sure the `path` tag provided by Vert.x is right. The HTTP resource path has its own meaning and it should not contain host and query parameters. So let's see how the discussion goes with them. My feeling is that they went this way because the commit change also moved from `HttpServerRequest` to use `HttpRequest` and this one doesn't expose `path()` but just `uri()`. Maybe they just didn't notice the implication of that. I think we should not change our dashboards but fixing (with them or by ourselves).

ppatierno:
As per discussion on #921 I am going to close this PR. We prefer to have it fixed on dashboards or properly by Vert.x folks (still waiting for their answer).

Issue #434, opened 2020-06-20
Ability to create topics using REST API calls 


metaversed:
We need a POST method that allows creation of new topics while using Kafka Bridge

opensourceonly:
This works while using  /topics/{topicname} to create a new topic assuming auto.create.topics.enable is set to true for kafka brokers

quickbooks2018:
https://www.youtube.com/watch?v=1Bv5SQW0BjI&t=336s&ab_channel=cloudgeeksinc

ilkerkocatepe:
Hello @ppatierno, I can work on this if still relevant

ppatierno:
@ilkerkocatepe sure, great! I just assigned it to you ;-)

antonio-pedro99:
> Hello @ppatierno, I can work on this if still relevant

Hi @ilkerkocatepe, is there any update on this? 
I am willing to take this one, please let me know if I should take it or not.

ilkerkocatepe:
Hello @antonio-pedro99, 
I'm currently working on this, I finished the code, I just need to test it :)

mabulgu:
Since this is resolved via https://github.com/strimzi/strimzi-kafka-bridge/pull/929 should we close this issue? 

 @ilkerkocatepe just a quick tip: Next time you can use `Closes: #ISSUE_ID` in the PR so it will automatically close the issue when the PR gets merged;)

PR #927, opened 2024-09-21
[#434] feat: added create topic endpoint

ilkerkocatepe:
This PR resolves issue https://github.com/strimzi/strimzi-kafka-bridge/issues/434

First of all, I am very sorry for the delay. From now on, I will make sure to take a task when I can complete it quickly.


ilkerkocatepe:
Hello @ppatierno, I have an issue with build docs, do you have an idea regarding fix it?

ppatierno:
> Hello @ppatierno, I have an issue with build docs, do you have an idea regarding fix it?

@ilkerkocatepe we have issues with swagger2markup. Some of its dependencies are disappeared from JCenter (and I also read that JCenter was dismissed by JFrog and all calls are redirected to Maven Central). I am investigating about a replacement for swagger2markup (maype using openapi-generator) but it's not there yet, so the build problem with doc is expected right now. I am sorry.

ilkerkocatepe:
> > Hello @ppatierno, I have an issue with build docs, do you have an idea regarding fix it?
> 
> @ilkerkocatepe we have issues with swagger2markup. Some of its dependencies are disappeared from JCenter (and I also read that JCenter was dismissed by JFrog and all calls are redirected to Maven Central). I am investigating about a replacement for swagger2markup (maype using openapi-generator) but it's not there yet, so the build problem with doc is expected right now. I am sorry.

thank you @ppatierno could you review the pr?

src/main/java/io/strimzi/kafka/bridge/KafkaBridgeAdmin.java
```diff
@@ -84,6 +86,29 @@ public CompletionStage<Set<String>> listTopics() {
...
+     * Creates a topic with given name
+     *
+     * @param topicName topic name to create
+     * @return a CompletionStage Void
+     */
+    public CompletionStage<Void> createTopic(String topicName) {
+        LOGGER.trace("Create topic thread {}", Thread.currentThread());
+        LOGGER.info("Create topic {}", topicName);
+        CompletableFuture<Void> promise = new CompletableFuture<>();
+        this.adminClient.createTopics(Collections.singletonList(new NewTopic(topicName, 2, (short) 1)))
```
ppatierno, line 99: why are partitions and replication factor hard coded? They should be parameters passed through the API imho.

src/main/resources/openapi.json
```diff
@@ -753,6 +753,56 @@
                 }
             ]
         },
+        "/create-topic/{topicname}": {
```
ppatierno, line 756: I don't think this API design is going to work. What are you going to have if we want the topic deletion, a new `/delete-topic` endpoint? It's not how REST API design works and scales.
I think we would need a new more general `/admin` endpoint? So a corresponding `/admin/topics` on which we can use the usual POST to create, DELETE to delete and so on? I think you should think more broader than just adding the topic creation, in terms of the API evolution for additional operations.

ppatierno (changes requested):

ppatierno:
@ilkerkocatepe other than taking a look at the comments I left, you could rebase against the main. We have fixed the build documentation related issue (now using a new maven plugin).

src/main/java/io/strimzi/kafka/bridge/KafkaBridgeAdmin.java
ilkerkocatepe: Sorry I missed this, will be fixed 👍

src/main/resources/openapi.json
ilkerkocatepe: You're right, I thought about it but couldn't decide, I will do as you said.

ppatierno:
@ilkerkocatepe I see you rebased but building the doc failed with this ...

```shell
M	documentation/book/api/.openapi-generator/openapi.json-generate-apidoc.sha256
M	documentation/book/api/index.adoc
  make docu_api \
```

Of course this is because you are changing the OpenAPI spec so the new openapi-generator is generatic a new index.adoc (with a different SHA) which you didn't push.
There is no need to push them now even because, as agreed, you are going to change the OpenAPI spec again so remember just to push the new above files when the new implementation is done.

ilkerkocatepe:
@ppatierno 
I finished the implementation that we have agreed. Could you check again if it's okay?
I pushed new doc files but it's conflicted

ppatierno:
@ilkerkocatepe did you rebase against main before pushing your changes?

ppatierno:
@ilkerkocatepe I see the PR in a little mess ... I don't think you changed 58 files :-D It contains all the other commits coming from my fix on the bridge doc and they should not be there. You should fix this first (I think the conflict will go away) and then I can properly review. Thanks!

ilkerkocatepe:
> @ilkerkocatepe did you rebase against main before pushing your changes?

of course but something went wrong :(
I'm gonna open a clear pr :)

PR #928, opened 2024-09-24
Replaced swagger2markup maven plugin with openapi-generator

ppatierno:
The current build of the bridge is broken, more specifically the part related to the documentation generation.
This is because the Maven plugin we use, swagger2markup, has been abandoned (since years) and lately some of its dependencies were removed from JCenter (which seems to be also going to be dismissed by JFrog).
Several times we tried to replace it and now it's the right time.
This PR is an attempt to use the openapi-generator project to do so. This project provides several generators for clients and documentation. It's able to generate code and content starting from the OpenAPI specification.
The generator used in this PR is the `asciidoc` one in order to generate the documentation for our OpenAPI specification in asciidoc format without making any change in the documentation flow generation we have today (through the Makefile).

The output I have got for the official website is pretty similar to the current one with some caviets:

* it seems openapi-generator doesn't have support for generating the examples in asciidoc starting from the examples we have in the OpenAPI specification. The only workaround they provide is by using "snippets". With snippets we can put adoc examples in specific folders and they are included in the final asciidoc file. The drawback of this approach is duplication: every time we change or add a new example to the OpenAPI spec, we need to copy the corresponding snippet into the adoc file.
* today our OpenAPI spec v3 defines our "key" and "value" for some components (i.e. ConsumerRecord, ProducerRecord) by using "oneOf" because they can be array, JSON (object) or string. From a generator perspective this is something that can't be handled by the generator itself because on code and documentation side, the generator can't create meaningful objects for a language. The "oneOf" approach is instead used by the vertx-web-openapi to validate the content of the body from the HTTP request. The end result of this is that right now the field defined as "oneOf" are rendered with a custom "empty" type by the openapi-generator and they are not rendered at all by the current swagger2markup (yes! we didn't notice that but for example, if you look for ProducerRecord in the current doc, the fields key and value are completely missing).
* the generated adoc file has some duplicated sections (I opened an issue on the openapi generator project to get insights https://github.com/OpenAPITools/openapi-generator/issues/19659) but I also noticed that the final rendering in HTML (through the asciidoctor) contains just one, so it's not a big issue.

In general, we should continue to investigate how to improve and fix the above "issues" but right now this is the only way I see to unlock the bridge build.


ppatierno:
I also noticed that this happened to Keycloack project in the past and they fixed by pushing the missing artifact in a personal namespace dropping JCenter. To be honest I am not so keed to this approach https://github.com/keycloak/keycloak/pull/5070

documentation/book/api/snippet/GET/http-response.adoc
```diff
@@ -0,0 +1,9 @@
+==== Example HTTP response
+
+===== Response 200
+[source,json]
+----
+{
+  "bridge_version" : "0.16.0"
```
scholzj, line 7: Why 0.16.0?

Makefile
```diff
@@ -50,17 +50,19 @@ release_package: java_package
 docu_html: docu_htmlclean docu_check
 	mkdir -p documentation/html
 	$(CP) -vrL documentation/book/images documentation/html/images
-	asciidoctor -v --failure-level WARN -t -dbook -a ProductVersion=$(RELEASE_VERSION) documentation/book/bridge.adoc -o documentation/html/bridge.html
+	# override snippetDir for asciidoctor to get right path to the snippets
+	asciidoctor -v --failure-level WARN -t -dbook -a ProductVersion=$(RELEASE_VERSION) -a snippetDir=${PWD}/documentation/book/api/snippet/ documentation/book/bridge.adoc -o documentation/html/bridge.html
```
scholzj, line 54: Why is this `snippetDir` needed? The path should be relative.

documentation/book/api/index.adoc
```diff
@@ -0,0 +1,3785 @@
...
+:keywords: openapi, rest, Kafka Bridge API Reference
+:specDir: 
+:snippetDir: ./documentation/book/api/snippet/
+:generator-template: v1 2019-12-20
+:info-url: https://openapi-generator.tech
+:app-name: Kafka Bridge API Reference
+
+= Kafka Bridge API Reference
+team@openapitools.org
+0.1.0
```
scholzj, line 14: This looks wierd?

documentation/book/api/index.adoc
```diff
@@ -0,0 +1,3785 @@
...
+
+== Introduction
+The Kafka Bridge provides a REST API for integrating HTTP based client applications with a Kafka cluster. You can use the API to create and manage consumers and send and receive records over HTTP rather than the native Kafka protocol. 
+
+
+// markup not found, no include::{specDir}intro.adoc[opts=optional]
```
scholzj, line 20: What does this mean?

documentation/book/api/snippet/GET/http-response.adoc
ppatierno: It's a very old example :-D I will update to the last one even if I think we can't get it automatically so it will stay at 0.30.0 for years again :-P

scholzj: I guess it should be removed, or have some placeholder or something if it cannot be kept up-to-date automatcially.

documentation/book/api/index.adoc
ppatierno: You can have some additional adoc to be included. We don't have this intro.adoc so the openapi-generator tool is just putting this comment as placeholder.

Makefile
ppatierno: It was a long story for having this working.
Yes this path is relative, as you can see in the pom.xml the `snippetDir` is defined as `./documentation/book/api/snippet/`. This path works for the openapi-generator which uses the pom.xml folder as basedir. This path doesn't work for asciidoctor (so it can't resolve the includes) because its base dir is the folder of the processed source file. Overriding `snippetDir` on the asciidoctor command line fixes the issue.

documentation/book/api/snippet/GET/http-response.adoc
ppatierno: I made it automatic, I was able to use the `{ProductVersion}` variable we pass on the command line already (not sure why because it wasn't used elsewhere, but was useful for this case :-)). 
Now getting something like this when building the snapshot version:

image

scholzj (commented):
Ok, I guess I do not have a strong opinion on this.

documentation/modules/con-overview-open-api-spec-kafka-bridge.adoc
```diff
@@ -13,7 +13,7 @@ The Kafka Bridge OpenAPI specification is in JSON format.
 You can find the OpenAPI JSON files in the `src/main/resources/` folder of the Kafka Bridge source download files.
 The download files are available from the {ReleaseDownload}.
 
-You can also use the xref:_openapi[`GET /openapi` method] to retrieve the OpenAPI v2 specification in JSON format.
+You can also use the xref:openapi[`GET /openapi` method] to retrieve the OpenAPI v2 specification in JSON format.
```
PaulRMellor, line 16: Do we want to talk about the V2 and v3 endpoints here and guide users to v3?

PaulRMellor: I guess these xrefs work okay from IDs in HTML output?  In the index.adoc, the refs have a period `[.openapi]`. Same with the `commit` xrefs etc

PaulRMellor (approved):
Good to find a solution to this ongoing issue. The content maps to what we have currently. If we want to maintain a single bridge doc including the API content, considering the API is stable so we don't need to change it much, this works fine.  

Otherwise, we could consider spinning off the API content and hosting HTML output that we link to from the main bridge docs. Or it is possible to include HTML content in an asciidoc book (but that might introduce other issues).

documentation/modules/con-overview-open-api-spec-kafka-bridge.adoc
ppatierno: > Do we want to talk about the V2 and v3 endpoints here and guide users to v3?

Take a note please and let's leave this change for a different PR. For this one, let's focus on having bridge doc back again.

> I guess these xrefs work okay from IDs in HTML output? In the index.adoc, the refs have a period [.openapi]. Same with the commit xrefs etc

The new generator generates the section headers with the period instead of the underscore but I also noticed that if I put the period here ... the link doesn't work. I had to remove it. Is it some asciidoc stuff you can confirm?

ppatierno: As agreed offline with Paul everything seems to be fine.

PR #929, opened 2024-10-05
[#434] feat: added create topic endpoint

ilkerkocatepe:
This PR resolves issue https://github.com/strimzi/strimzi-kafka-bridge/issues/434


src/main/java/io/strimzi/kafka/bridge/KafkaBridgeAdmin.java
```diff
@@ -84,6 +86,31 @@ public CompletionStage<Set<String>> listTopics() {
...
 
+    /**
+     * Creates a topic with given name
+     *
+     * @param topicName topic name to create
+     * @param partitions number of partitions
+     * @param replicationFactor replication factor
+     * @return a CompletionStage Void
+     */
+    public CompletionStage<Void> createTopic(String topicName, int partitions, short replicationFactor) {
```
ppatierno, line 97: why are you enforcing the user to provide partitions and replicationFactor? The `NewTopic` class provides also a way to have them as optional, I think the user should be able to ask for creating a topic by using the defaults as well.

src/main/resources/openapi.json
```diff
@@ -753,6 +753,74 @@
...
+                        "content": {
+                            "application/vnd.kafka.v2+json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/TopicMetadata"
+                                }
+                            }
+                        }
+                    },
+                    "404": {
+                        "description": "The specified topic was not found.",
```
ppatierno, line 775: what does this mean when you are creating a topic? You are not looking for a topic which could be not found, or?

src/main/resources/openapi.json
```diff
@@ -753,6 +753,74 @@
...
+                            }
+                        }
+                    }
+                }
+            },
+            "parameters": [
+                {
+                    "name": "topicname",
+                    "in": "path",
+                    "description": "Name of the topic to send records to or retrieve metadata from.",
```
ppatierno, line 798: wrong description.

src/main/resources/openapi.json
```diff
@@ -753,6 +753,74 @@
...
+                ],
+                "description": "Creates a topic with given topic name.",
+                "operationId": "createTopic",
+                "responses": {
+                    "200": {
+                        "description": "Topic metadata",
+                        "content": {
+                            "application/vnd.kafka.v2+json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/TopicMetadata"
```
ppatierno, line 769: Is topic creation really returning anything? I see the API just having void as well as your implementation.

src/main/resources/openapiv2.json
```diff
@@ -686,6 +686,66 @@
         }
       ]
     },
+    "/admin/topics/{topicname}": {
```
ppatierno, line 689: all the comments for the OpenAPI v3 apply to this V2 definition as well.

ppatierno (changes requested):

ilkerkocatepe:
Hi @ppatierno,
thank you for the review, I updated the files which you mentioned.

src/main/java/io/strimzi/kafka/bridge/KafkaBridgeAdmin.java
```diff
@@ -84,6 +87,31 @@ public CompletionStage<Set<String>> listTopics() {
         return promise;
     }
 
+    /**
+     * Creates a topic with given name
```
ppatierno, line 91: It would be helpful to complete the description mentioning also optional partitions and replication factor.

src/main/resources/openapi.json
```diff
@@ -615,7 +615,7 @@
                         }
                     },
                     "404": {
-                        "description": "The specified topic was not found.",
+                        "description": "The path was not found.",
```
ppatierno, line 618: I guess this overall 404 error was copied/pasted from the corresponding GET operation we already have but it's not still clear to me when this error should happen (there is also no test for it). I mean, you run a POST against an endpoint like `/topics/{topicname}` where the parameter on the path is a "input" parameter for the topic creation so it's clear it can't exist already (if you are creating it). When a NOT FOUND 404 should happen then?

ppatierno (commented):
@ilkerkocatepe thanks for addressing my comments. I left two more and also I think you should rebase again because you are bringing some changes in the pom.xml which are not related to your PR but commits already merged into main.

src/main/resources/openapi.json
ilkerkocatepe: You're right, I copied that from another API. I removed it, if it is ok :)

src/main/resources/openapi.json
```diff
@@ -613,24 +613,6 @@
                                 }
                             }
                         }
-                    },
-                    "404": {
```
ppatierno, line 617: I asked to remove 404 from the admin endpoint when creating a topic (and you did that) but why removing the 404 when you GET a topic? it could not exist. Anyway, it's totally unrelated to this PR.

ppatierno (changes requested):

src/main/resources/openapi.json
ilkerkocatepe: I'm so sorry, it's totally my fault

ppatierno:
@ilkerkocatepe thanks for addressing my comments but thinking more on your API proposal I have a more "destructive" one :-)
This proposal is about using the `/topics/{topicName}` URL where you specify the topic name as part of the path and the passing replication factor and partitions via query parameters. While it would work, I think it's not that much extensible for future additions like allowing to specify topic configuration when creating a topic.
I think we should go in a different direction by using a JSON as body of the HTTP POST request where you can specify parameters. What I have in mind is something like this, let me know what you think ...

The URL would be just `/admin/topics/` and the POST request could have a body like the following:

```json
{
    "topic_name": "my-new-topic",
    "partitions_count": 5,
    "replication_factor": 3,
}
```

The above JSON would be extensible in the future, allowing to have a "configs" field for example with a properties map for topic configuration parameters.
Of course, it means changing the OpenAPI spec as well and providing examples of body within the spec as well.
Wdyt?

ilkerkocatepe:
You are absolutely right. Also, it will be proper way according to REST standards. I would like to edit implementation :)

ppatierno:
@ilkerkocatepe great to hear that you are still onboard! My apologies for realising this so late. Thanks for the effort!

ilkerkocatepe:
@ppatierno no worries, I develop similar things at work too, I should think about it :)

src/main/resources/openapi.json
```diff
@@ -2193,6 +2239,32 @@
                     }
                 ],
                 "nullable" : true
+            },
+            "CreateTopic": {
+                "title": "CreateTopic",
```
ppatierno, line 2244: I guess we should add the "required" field as well to have "topic_name" as required. It will help the OpenAPI validator to check it. It should allow to avoid having the double check on topic name in the code as well.

src/main/resources/openapiv2.json
```diff
@@ -2027,6 +2070,30 @@
         "string",
         "null"
       ]
+    },
+    "CreateTopic": {
+      "title": "CreateTopic",
```
ppatierno, line 2075: Ditto as above about the "required" field.

ppatierno (commented):
This is a really big improvement towards the end! I left a few comments.

ilkerkocatepe:
Hi @ppatierno, I updated points that you mentioned. But I have a question, I added required tag for topic_name and now if request body doesn't contain topic_name, endpoint gives 400 code. So I changed test response to BAD_REQUEST. What do you think about validations on HttpAdminBridgeEndpoint.doCreateTopic?

src/main/java/io/strimzi/kafka/bridge/http/HttpAdminBridgeEndpoint.java
```diff
@@ -173,6 +179,57 @@ public void doGetTopic(RoutingContext routingContext) {
...
+            HttpBridgeError error = new HttpBridgeError(
+                    HttpResponseStatus.UNPROCESSABLE_ENTITY.code(),
+                    "Request body must be a JSON object"
+            );
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.UNPROCESSABLE_ENTITY.code(),
+                    BridgeContentType.KAFKA_JSON, JsonUtils.jsonToBytes(error.toJson()));
+            return;
+        }
+
+        if (jsonBody.getString("topic_name").isBlank()) {
```
ppatierno, line 200: with the field being required, this check should not be needed anymore.

src/main/resources/openapi.json
```diff
@@ -753,6 +753,52 @@
...
+                    "required": true
+                },
+                "responses": {
+                    "201": {
+                        "description": "Created",
+                        "content": {
+                            "application/vnd.kafka.v2+json": {}
+                        }
+                    },
+                    "400": {
```
ppatierno, line 781: I would remove this one because we don't have this kind of error described in any of the other operations (even if it can happen on the OpenAPI validation side).

src/main/resources/openapiv2.json
```diff
@@ -686,6 +686,49 @@
...
+              "$ref": "#/definitions/NewTopic"
+            }
+          }
+        ],
+        "responses": {
+          "201": {
+            "description": "Created",
+            "schema": {}
+          },
+          "400": {
```
ppatierno, line 717: Ditto as above

documentation/book/api/index.adoc
```diff
@@ -2289,6 +2289,95 @@ endif::internal-generation[]
...
+| 400
+| Request body is null or it doesn&#39;t contain topic name.
+|  <<Error>>
+
+|===
+
+===== Samples
+
+
+// markup not found, no include::{snippetDir}admin/topics/POST/http-request.adoc[opts=optional]
```
ppatierno, line 2362: because of how the openapi-generator works for building the doc you would need to add an `http-request.adoc` file with an example request in the `/documentation/book/api/snippet/admin/topics/POST` folder (you need to create the subfolders path `/admin/topics/POST`).
For the adoc file you could take the consumer creation as an example https://github.com/strimzi/strimzi-kafka-bridge/blob/main/documentation/book/api/snippet/consumers/%7Bgroupid%7D/POST/http-request.adoc
We just need an example JSON body.

ilkerkocatepe:
Hi @ppatierno thank you for review, and sorry for delay. the pr is ready for review again :)

src/main/resources/openapi.json
```diff
@@ -753,6 +753,34 @@
...
+                        }
+                    },
+                    "required": true
+                },
+                "responses": {
+                    "201": {
+                        "description": "Created",
+                        "content": {
+                            "application/vnd.kafka.v2+json": {}
+                        }
```
ppatierno, line 779: I think we can just leave the description and removing the content.

src/main/resources/openapiv2.json
```diff
@@ -686,6 +686,37 @@
...
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/NewTopic"
+            }
+          }
+        ],
+        "responses": {
+          "201": {
+            "description": "Created",
+            "schema": {}
```
ppatierno, line 715: ditto as above, I think we can remove the schema field.

ppatierno (approved):
I think we are there, I left one nit but LGTM overall!
If you could make that change it would be great. Thanks for your work!

ppatierno:
@ilkerkocatepe other than my last comments, we would need a line in the CHANGELOG about this addition. Thanks!

src/main/resources/openapi.json
```diff
@@ -753,6 +753,34 @@
                 }
             ]
         },
+        "/admin/topics": {
+            "post": {
+                "tags": [
+                    "Topics"
+                ],
+                "description": "Creates a topic with given name, partitions count and replication factor.",
```
PaulRMellor, line 761:
```suggestion
                "description": "Creates a topic with given name, partitions count, and replication factor.",
```

src/main/resources/openapi.json
```diff
@@ -753,6 +753,34 @@
...
+                "operationId": "createTopic",
+                "requestBody": {
+                    "description": "Creates a topic with given name, partitions count and replication factor.",
```
PaulRMellor, line 764:
```suggestion
                    "description": "Creates a topic with given name, partitions count, and replication factor.",
```

src/main/resources/openapiv2.json
```diff
@@ -686,6 +686,37 @@
         }
       ]
     },
+    "/admin/topics": {
+      "post": {
+        "tags": [
+          "Topics"
+        ],
+        "description": "Creates a topic with given name, partitions count and replication factor.",
```
PaulRMellor, line 694:
```suggestion
        "description": "Creates a topic with given name, partitions count, and replication factor.",
```

PaulRMellor (approved):
Looks good to me.
Just a couple of nits for doc consistency

scholzj (approved):
Is it intentional that it contains the dpeendency changes etc.?

ppatierno:
@ilkerkocatepe as Jakub said you have something which belongs to an already existing commit from here https://github.com/strimzi/strimzi-kafka-bridge/commit/ea03491fa72e3471345a1e7cf8eedbcacbce794c
Can you fix it please?

ppatierno:
@ilkerkocatepe we are planning to start releasing bridge 0.31.0 RC1 by the end of the week and it would be great to have this in. Do you think you are able clean this PR (removing the unrelated commit) without opening a new one and copying over the code there? Otherwise I will just merge this .

ilkerkocatepe:
I think I messed up everything and need help :/

PR #945, opened 2024-11-25
Fix metadata order in response when producing records

### LINKED ISSUE #944: Produced Records result offset order not deterministic
While running through the quickstart for the 0.31.0 RC I noticed that in the [produce example](https://strimzi.io/docs/bridge/latest/#proc-producing-messages-from-bridge-topics-partitions-bridge) the second record in the array is forced to partition 2, but in the result array the second item is not always partition 2, eg:

```
curl -X POST \
  http://localhost:8080/topics/bridge-quickstart-topic \
  -H 'content-type: application/vnd.kafka.json.v2+json' \
  -d '{
    "records": [
        {
            "key": "my-key",
            "value": "sales-lead-0001"
        },
        {
            "value": "sales-lead-0002",
            "partition": 2
        },
        {
            "value": "sales-lead-0003"
        }
    ]
}'
{"offsets":[{"partition":0,"offset":28},{"partition":0,"offset":29},{"partition":2,"offset":36}]}
```
and checking where they went directly:

```
./kafka-console-consumer.sh   --topic bridge-quickstart-topic   --bootstrap-server :9092   --property print.key=true   --property print.partition=true
...
Partition:0	"my-key"	"sales-lead-0001"
Partition:0	null	"sales-lead-0003"
Partition:2	null	"sales-lead-0002"
```
the record did go to partition 2, so it looks like some non-determinism in the `offsets` response.

This is testing against a single kafka 3.9.0 broker deployed per the Apache Kafka [quickstart](https://kafka.apache.org/quickstart) in KRaft mode. I have observed this behaviour with bridge 0.30.0 and 0.31.0-rc1

Maybe there's a misconception about send callback execution order on the Kafka Producer:
```
Callbacks for records being sent to the same partition are guaranteed to execute in order. That is, in the following example callback1 is guaranteed to execute before callback2:
```
The ordering is only guaranteed per partition.

robobario:
fix: preserve record send order in response metadata
    
This uses the send CompletionStages to keep the original ordering of records.

kafka-clients [KafkaProducer#send](https://kafka.apache.org/39/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#send(org.apache.kafka.clients.producer.ProducerRecord,org.apache.kafka.clients.producer.Callback)) only guarantees that callbacks are executed in order per partition, the ordering across partitions is not guaranteed.

> Callbacks for records being sent to the same partition are guaranteed to execute in order. That is, in the following example callback1 is guaranteed to execute before callback2:
>
> producer.send(new ProducerRecord<byte[],byte[]>(topic, partition, key1, value1), callback1);
> producer. send(new ProducerRecord<byte[],byte[]>(topic, partition, key2, value2), callback2);

Closes #944

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Dependency updates (Kafka 3.9.0, Vert.x 4.5.11, Netty 4.1.115.Final)
 * Added support for creating a new topic via endpoint.
+* Fixed metadata order on the HTTP "offsets" response when producing records.
 
 ## 0.30.0
 
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -35,6 +35,7 @@ import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
 
 /**
  * Represents an HTTP bridge source endpoint for the Kafka producer operations
@@ -137,7 +138,6 @@ public class HttpSourceBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
 
             return;
         }
-        List<HttpBridgeResult<?>> results = new ArrayList<>(records.size());
 
         // fulfilling the request of sending (multiple) record(s) sequentially but in a separate thread
         // this will free the Vert.x event loop still in place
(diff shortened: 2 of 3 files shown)


robobario:
the ppc64le build failed with:
```
SeekIT.seekToOffsetMultipleTopicsWithNotSuscribedTopic(VertxTestContext) java.lang.AssertionError: 
Expected: is "No current assignment for partition seekToOffsetNotSubscribedTopic-0"
     but: was "No current assignment for partition seekToOffseSubscribedTopic-0"
```
doesn't sound related to produce changes

ppatierno (approved):
LGTM! Good catch @robobario! Thanks!

scholzj (approved):
LGTM if it works.

ppatierno:
I am going to cherry-pick this one for the `release-0.31.x` branch for having it in the 0.31.0 release.

PR #951, opened 2024-12-04
Use the correct content-type for metrics

scholzj:
This PR contributes to https://github.com/strimzi/strimzi-kafka-operator/issues/10902 and sets the correct content-type for the metrics HTTP endpoint.


ppatierno (approved):
LGTM.

PR #954, opened 2024-12-17
Add support for Strimzi Metrics Reporter

fvaleri:
This patch adds support for metrics types configuration, integrating the [Strimzi Metrics Reporter](https://github.com/strimzi/metrics-reporter), and custom Prometheus JMX Exporter configuration. The custom configuration feature is needed for Cluster Operator integration, but it can also be leveraged by standalone users. See [proposal 43](https://github.com/strimzi/proposals/pull/143) for more details.


fvaleri:
@OwenCorrigan76 fyi

config/application.properties
```diff
@@ -1,5 +1,15 @@
 #Bridge related settings
 bridge.id=my-bridge
+
+# uncomment the following line to enable JMX Exporter metrics, check the documentation for more details
+#bridge.metrics=jmxPrometheusExporter
+
+# uncomment the following lines to enable Strimzi Reporter metrics, check the documentation for more details
+#bridge.metrics=strimziMetricsReporter
```
mimaison, line 8: From a configuration standpoint I would expect that users would just need to set `bridge.metrics=strimziMetricsReporter` and maybe the allowlist and not set the other individual configurations `kafka.prometheus.metrics.reporter.listener.enable`, `kafka.metric.reporters`.

documentation/modules/proc-configuring-kafka-bridge-jmx-metrics.adoc
```diff
@@ -1,22 +1,22 @@
-[id='proc-configuring-kafka-bridge-metrics-{context}']
-= Configuring metrics
+[id='proc-configuring-kafka-bridge-jmx-metrics-{context}']
```
mimaison, line 1: I think users should follow one method _OR_ the other, otherwise you probably end up with duplicated metrics.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,35 @@
+[id='proc-configuring-kafka-bridge-smr-metrics-{context}']
```
mimaison, line 1: I don't know if the bridge has metrics dashboards but if so they probably need updating when using the metrics reporter.

mimaison (commented):
I've not looked at the code yet. I left comments trying to understand how this will be used.

config/application.properties
fvaleri: This is how it will work when using the Cluster Operator. Talking with @ppatierno, we agreed that, for the standalone Bridge, the user should be free to set all configurations through the properties file.

documentation/modules/proc-configuring-kafka-bridge-jmx-metrics.adoc
fvaleri: Correct. Thanks.

scholzj (commented):
I wonder if this should go through a dedicated proposal first. The original proposal - https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md - sadly didn't cover the bridge in sufficient depth and there will be some backward compatibility impact here or in the operators repo that needs to be carefully considered. A proposal would be the right place to discuss it and decide on the best approach.

fvaleri:
> there will be some backward compatibility impact here or in the operators repo that needs to be carefully considered

I agree. Probably the backwards incompatible changes are mostly on the operator code, where we have to introduce `metricsConfig` at the CRD level. @ppatierno wdyt?

config/application.properties
ppatierno: Yeah. When bridge is running on bare metal / VM, it's the user to provide the application.properties file for the bridge with its configuration. The bridge doesn't have to change the properties file in any way enforcing other parameters which the user didn't set.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 ## 0.32.0
 
 * Dependency updates (JMX exporter 1.1.0)
+* Added support for the [Strimzi Metrics Reporter](https://github.com/strimzi/metrics-reporter) metrics.
+  This is a Kafka plugin that directly exports metrics in Prometheus format, and can be enabled by setting `bridge.metrics=strimziMetricsReporter`.
+
+### Changes, deprecations and removals
+
+* `KAFKA_BRIDGE_METRICS_ENABLED` has been deprecated.
+  The equivalent configuration is `bridge.metrics=jmxPrometheusExporter`.
```
ppatierno, line 12: "The equivalent configuration ..." is not that clear. I would state clearly that to keep the current behaviour and use JMX exporter, the property has to be set to `jmxPrometheusExporter`

pom.xml
```diff
@@ -108,8 +108,9 @@
 		<kafka-env-var-config-provider.version>1.1.0</kafka-env-var-config-provider.version>
 		<maven.checkstyle.version>3.3.0</maven.checkstyle.version>
 		<checkstyle.version>10.12.2</checkstyle.version>
-		<hamcrest.version>2.2</hamcrest.version>
 		<junit.version>5.8.2</junit.version>
+		<hamcrest.version>2.2</hamcrest.version>
```
ppatierno, line 112: why moving this line?

pom.xml
```diff
@@ -126,14 +127,15 @@
...
+		<strimzi-metrics-reporter.version>0.1.0</strimzi-metrics-reporter.version>
 		<opentelemetry.version>1.34.1</opentelemetry.version>
 		<opentelemetry-alpha.version>1.34.1-alpha</opentelemetry-alpha.version>
 		<opentelemetry.instrumentation.version>1.32.0-alpha</opentelemetry.instrumentation.version>
 		<opentelemetry-semconv.version>1.21.0-alpha</opentelemetry-semconv.version>
 		<grpc-netty-shaded.version>1.61.0</grpc-netty-shaded.version>
 		<micrometer.version>1.12.3</micrometer.version>
-		<jmx-prometheus-collector.version>1.1.0</jmx-prometheus-collector.version>
 		<prometheus-client.version>1.3.4</prometheus-client.version>
+		<jmx-prometheus-collector.version>1.1.0</jmx-prometheus-collector.version>
```
ppatierno, line 138: why moving this line?

ppatierno (commented):
I left just some minor comments but I will review it again when the proposal is accepted (to avoid reviewing now something that will be changed later by the final proposal)

CHANGELOG.md
fvaleri: Ok, let me rephrase.

pom.xml
fvaleri: No reason, will revert.

pom.xml
fvaleri: No reason, will revert.

fvaleri:
@strimzi/maintainers proposal 92 has been accepted, and this PR is aligned, so open for review. Thanks.

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -132,25 +104,73 @@ private static Future<HttpBridge> deployHttpBridge(Vertx vertx, BridgeConfig bri
...
+    /**
+     * Return a JmxCollectorRegistry instance with the YAML configuration filters.
+     * This is loaded from a custom config file if present, or from the default configuration file.
+     *
+     * @return JmxCollectorRegistry instance
+     * @throws MalformedObjectNameException
+     * @throws IOException
+     */
+    private static JmxCollectorRegistry getJmxCollectorRegistry(BridgeConfig bridgeConfig) throws MalformedObjectNameException, IOException {
+        if (bridgeConfig.getJmxExporterConfigPath() != null && Files.exists(bridgeConfig.getJmxExporterConfigPath())) {
```
scholzj, line 156: If the file does not exist, you should probably exit with an error and not silently ignore it.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -90,6 +123,30 @@ public String getBridgeID() {
         }
     }
 
+    /**
+     * @return the metric system to be used in the bridge
+     */
+    public String getMetrics() {
```
scholzj, line 129: Would this better describe what it does?

```suggestion
    public String getMetricsType() {
```

Also, as you defined the Enum, should you use it instead?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -166,7 +166,9 @@ public void start(Promise<Void> startPromise) {
                 routerBuilder.operation(this.OPENAPI.getOperationId().toString()).handler(this.OPENAPI);
                 routerBuilder.operation(this.OPENAPIV2.getOperationId().toString()).handler(this.OPENAPIV2);
                 routerBuilder.operation(this.OPENAPIV3.getOperationId().toString()).handler(this.OPENAPIV3);
-                routerBuilder.operation(this.METRICS.getOperationId().toString()).handler(this.METRICS);
+                if (metricsReporter != null) {
+                    routerBuilder.operation(this.METRICS.getOperationId().toString()).handler(this.METRICS);
+                }
```
scholzj, line 171: Why exactly is this change needed?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -603,6 +606,8 @@ private void errorHandler(RoutingContext routingContext) {
             }
         } else if (routingContext.statusCode() == HttpResponseStatus.NOT_FOUND.code()) {
             message = HttpResponseStatus.NOT_FOUND.reasonPhrase();
+        } else if (routingContext.statusCode() == HttpResponseStatus.NOT_IMPLEMENTED.code()) {
+            message = HttpResponseStatus.NOT_IMPLEMENTED.reasonPhrase();
```
scholzj, line 610: What are we using the not-implemented code for? Not implemented is not the same as not enabled.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -853,4 +850,17 @@ public void process(RoutingContext routingContext) {
...
+            super(operationId);
+        }
+
+        @Override
+        public void process(RoutingContext routingContext) {
+            HttpBridgeError error = new HttpBridgeError(HttpResponseStatus.GONE.code(), "OpenAPI v2 Swagger not supported");
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.GONE.code(),
+                BridgeContentType.KAFKA_JSON, JsonUtils.jsonToBytes(error.toJson()));
+        }
+    }
```
scholzj, line 865: Is this really related to the metrics?

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java
```diff
@@ -43,14 +44,14 @@ public ProducerRecord<byte[], byte[]> toKafkaRecord(String kafkaTopic, Integer p
                 if (!keyNode.isTextual()) {
                     throw new IllegalStateException("Because the embedded format is 'text', the key must be a string");
                 }
-                key = keyNode.asText().getBytes();
+                key = keyNode.asText().getBytes(Charset.forName("UTF-8"));
```
scholzj, line 47: Same as above, are these changes really related to the metrics? How?

src/main/java/io/strimzi/kafka/bridge/metrics/MetricsReporter.java
scholzj, line 1: Should this be rather some abstract class / interface with two implementations then this kind of _union_ class?

src/main/java/io/strimzi/kafka/bridge/metrics/StrimziCollectorRegistry.java
scholzj, line 1: Should there be some registry interface or something?

CHANGELOG.md
```diff
@@ -9,6 +9,15 @@
   * Both the `/openapi` and `/openapi/v3` endpoints return the OpenAPI v3 definition of the bridge REST API.
 * Removed script to build bridge configuration within the container. 
   It is going to be set up by the Strimzi operator within a ConfigMap and mounted as volume on the bridge pod.
+* Added support for [Strimzi Metrics Reporter](https://github.com/strimzi/metrics-reporter).
+  This is a Kafka plugin that directly exposes metrics in Prometheus format without using JMX, and can be enabled by setting `bridge.metrics=strimziMetricsReporter`.
```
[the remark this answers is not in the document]
fvaleri: Ok, let's remove JMX mention. Users interested in knowing more about implementation details can simply follow the link.

pom.xml
```diff
@@ -305,6 +312,16 @@
...
+		<dependency>
+			<groupId>io.prometheus</groupId>
+			<artifactId>prometheus-metrics-instrumentation-jvm</artifactId>
+			<version>${prometheus-client.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.prometheus</groupId>
+			<artifactId>prometheus-metrics-exporter-httpserver</artifactId>
+			<version>${prometheus-client.version}</version>
+		</dependency>
```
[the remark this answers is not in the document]
fvaleri: Strimzi Metrics Reporter 0.1.0 brings in these deps at 1.3.1, while the Bridge is using 1.3.4. This is why I included them explicitly. Do you think this is wrong or unnecessary?

src/main/java/io/strimzi/kafka/bridge/Application.java
fvaleri: I agree, we should log an error and fail if the configured file is not found. Otherwise we risk the user never realizing that it is not running with the custom config simply due to a small typo, for example. Fixed.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
fvaleri: I agree. Should we do the same with the pre existing `getTracing` method?

pom.xml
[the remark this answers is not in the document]
scholzj: Ok, that sounds like the change is reasonable. But please include a comment with an explanation.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+[source,properties]
+----
+bridge.metrics=strimziMetricsReporter
+----
++
+Optionally, you can set a comma-separated list of regexes used to filter exposed metrics using the `kafka.prometheus.metrics.reporter.allowlist` property.
+If not set, all available metrics will be exposed.
++
+You can add any plugin configuration to the Bridge properties file using the `kafka.` prefix.
```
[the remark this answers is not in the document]
fvaleri: Yes, but we want metrics for all internal clients right?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -55,21 +67,42 @@ public HttpConfig getHttpConfig() {
...
+                throw new IllegalArgumentException(
+                    String.format("Invalid %s configuration, choose one of %s and %s", 
+                        METRICS_TYPE, MetricsType.JMX_EXPORTER, MetricsType.STRIMZI_REPORTER)
+                );
+            }
+            // apply default Strimzi Metrics Reporter configurations if not present
+            if (metricsTypeValue.equals(MetricsType.STRIMZI_REPORTER.toString())) {
+                map.putIfAbsent("kafka.metric.reporters", "io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter");
+                map.putIfAbsent("kafka.prometheus.metrics.reporter.listener.enable", "false");
+                map.putIfAbsent("kafka.prometheus.metrics.reporter.allowlist", ".*");
```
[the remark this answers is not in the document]
fvaleri: > I would expect that this part will be always set by the operator when the bridge is running within Kubernetes. 

Correct.

> If running on bare metal/VM are we sure that we want to set defaults instead of raising an exception for missing configuration?

This is what was agreed in the proposal.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
fvaleri: No. This is due to some validation/spotbug error I got after rebasing.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java
fvaleri: No. This is due to some validation/spotbug error I got after rebasing.

fvaleri:
@ppatierno @scholzj your comments should be addressed, but there are a few open questions. For the new abstraction I used the `MetricsCollector` name, as `MetricsReporter` would create some confusion IMO. Let me know what you think.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
ppatierno: Now I remember why it was named `getTracing`, because the corresponding property is just  `tracing`. So you used `getMetrics`, because of the corresponding property is named just `metrics`.
Not strong opinion, we could leave the consistency with the corresponding property and having `getMetrics` but of course using the enum. But as I said, not strong opinion on this.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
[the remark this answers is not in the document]
ppatierno: Yes, and is it something we want? Don't we want users able to have metrics for some clients but not others?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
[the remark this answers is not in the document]
ppatierno: I think we have a problem here or you should verify how the bridge behaves and fix otherwise.
Let's use `kafka.metric.reporters` as an example and let's assume the user doesn't set it (so this code is going to set it as default) but the user sets `kafka.producer.metric.reporters`. Which option is going to win? I would expect what the user wants for producer and the default one for the others (admin and consumer) but it has to be validated on the bridge because of the way the properties are provided down to the clients' configurations.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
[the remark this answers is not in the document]
fvaleri: This is still possible using the client prefix to configure the `allowlist`. For example, to exclude all producer metrics the user can set `kafka.prometheus.metrics.reporter.allowlist=`.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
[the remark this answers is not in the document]
fvaleri: From [here](https://github.com/strimzi/strimzi-kafka-bridge/blob/0.31.1/src/main/java/io/strimzi/kafka/bridge/KafkaBridgeProducer.java#L88-L89), you can see that the most specific settings always wins. For example, `kafka.producer.metric.reporters` wins over `kafka.metric.reporters`. I think this behavior matches with the user expectation that you describe. Anyway, I'm adding a documentation note to avoid any confusion.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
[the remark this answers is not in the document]
ppatierno: > to exclude all producer metrics the user can set kafka.prometheus.metrics.reporter.allowlist=.

I guess you meant `kafka.producer.prometheus.metrics.reporter.allowlist=.` ? Also should we say it clearly in the doc? Not just mentioning the `kafka.` prefix which covers all clients but also the other prefixes to cover specific clients?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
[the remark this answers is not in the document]
ppatierno: Ok then. I guess it should work.

ppatierno:
@fvaleri can you resolve the conflicts before I have another pass please?

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -88,70 +73,107 @@ public static void main(String[] args) {
...
-     * @return JmxCollectorRegistry instance
-     * @throws MalformedObjectNameException
-     * @throws IOException
+     * @return instance of the MicrometerMetricsOptions on Vert.x
      */
-    private static JmxCollectorRegistry getJmxCollectorRegistry()
+    private static MicrometerMetricsOptions metricsOptions() {
+        Set<String> set = new HashSet<>();
+        set.add(MetricsDomain.NAMED_POOLS.name());
+        set.add(MetricsDomain.VERTICLES.name());
```
ppatierno, line 124: What about using 

```java
Set<String> set = Set.of(MetricsDomain.NAMED_POOLS.name(), MetricsDomain.VERTICLES.name());
```

to replace the above three lines?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -55,21 +67,40 @@ public HttpConfig getHttpConfig() {
...
         HttpConfig httpConfig = HttpConfig.fromMap(map);
-
         return new BridgeConfig(map.entrySet().stream()
                 .filter(e -> e.getKey().startsWith(BridgeConfig.BRIDGE_CONFIG_PREFIX))
                 .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)),
                 kafkaConfig, httpConfig);
     }
 
+    private static void validateAndApplyDefaults(Map<String, Object> map) {
+        String metricsTypeValue = (String) map.get(METRICS_TYPE);
```
ppatierno, line 87: can't you get the enum from the String (as by using `MetricsType.fromString`) method and working with enum values instead of using string "equals" across the method?
You could also use explicit `BridgeConfig.METRICS_TYPE` as done in the `getMetricsType()` method.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -140,68 +140,67 @@ private void startInactiveConsumerDeletionTimer(Long timeout) {
 
     @Override
     public void start(Promise<Void> startPromise) {
-
         RouterBuilder.create(vertx, "openapi.json")
-                .onComplete(ar -> {
```
ppatierno, line 145: please don't change this indentation. It has been working pretty fine with people involved in the bridge so far. It's also not related to the PR.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Which kind of validation/spotpug error? We don't get any right now in the main. This refactoring seems to be useless from a pure code perspective.

src/main/java/io/strimzi/kafka/bridge/http/converter/HttpTextMessageConverter.java
ppatierno: Again, why it's not happening today in the main?

ppatierno (changes requested):

config/application.properties
```diff
@@ -1,5 +1,16 @@
...
+
+# uncomment the following line to enable Prometheus JMX Exporter, check the documentation for more details
+#bridge.metrics=jmxPrometheusExporter
+# optionally, you can set the file path of your custom configuration
+#bridge.metrics.exporter.config.path=/path/to/my-exporter-config.yaml
+
+# uncomment the following line to enable Strimzi Metrics Reporter, check the documentation for more details
+#bridge.metrics=strimziMetricsReporter
+# optionally, you can filter the exposed metrics using a comma separated list of regexes
+#kafka.prometheus.metrics.reporter.allowlist=.*
```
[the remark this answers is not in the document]
fvaleri: Comment updated. On line 9 we already say check doc for more details.

src/main/java/io/strimzi/kafka/bridge/Application.java
fvaleri: Ok, I will also inline.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
fvaleri: Yes, I just forgot to update this method. Getting the enum also does validation.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
fvaleri: That was unintentional, and due to conflicts resolution. Reverted.

fvaleri:
> Which kind of validation/spotpug error? We don't get any right now in the main. This refactoring seems to be useless from a pure code perspective.

```sh
[INFO] --- spotbugs:4.7.3.0:check (default-cli) @ kafka-bridge ---
[INFO] BugInstance size is 3
[INFO] Error size is 0
[INFO] Total bugs: 3
[ERROR] Low: Private method io.strimzi.kafka.bridge.http.HttpBridge.contentTypeToFormat(String) is never called [io.strimzi.kafka.bridge.http.HttpBridge] At HttpBridge.java:[lines 644-652] UPM_UNCALLED_PRIVATE_METHOD
[ERROR] High: Found reliance on default encoding in io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter.toKafkaRecord(String, Integer, byte[]): String.getBytes() [io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter, io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter] At HttpTextMessageConverter.java:[line 46]Another occurrence at HttpTextMessageConverter.java:[line 53] DM_DEFAULT_ENCODING
[ERROR] High: Found reliance on default encoding in io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter.toMessages(ConsumerRecords): new String(byte[]) [io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter, io.strimzi.kafka.bridge.http.converter.HttpTextMessageConverter] At HttpTextMessageConverter.java:[line 104]Another occurrence at HttpTextMessageConverter.java:[line 105] DM_DEFAULT_ENCODING
```

> Again, why it's not happening today in the main?

I see these errors on the main branch too. Just tried with a fresh clone.

ppatierno:
> I see these errors on the main branch too. Just tried with a fresh clone.

Ok so I see another issue here. Our `make all` is not running the `spotbugs` target automatically and it should be fixed.
@fvaleri could you remove these changes from this PR, and opening a new one where you have those fixes and adding spotbugs as target during the make all process so that it also runs in the Azure and Travis CI pipeline please?

fvaleri:
The problem is that we run spotbugs before building in the CI. I'll fix that in the spinoff PR.

fvaleri:
@ppatierno rebased. Let's wait for the CI to complete.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -558,10 +557,14 @@ private void openapi(RoutingContext routingContext) {
...
     private void metrics(RoutingContext routingContext) {
-        routingContext.response()
+        if (metricsCollector != null) {
+            routingContext.response()
                 .putHeader("Content-Type", "text/plain; version=0.0.4; charset=utf-8")
                 .setStatusCode(HttpResponseStatus.OK.code())
-                .end(metricsReporter.scrape());
+                .end(metricsCollector.scrape());
+        } else {
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.NOT_FOUND.code(), null, null);
```
ppatierno, line 566: So ... I like this because as today, when metrics are not enabled, we are returning 200 OK but with an empty body. I think that's wrong so it's good to have 404 NOT_FOUND. It would need a couple of changes:

1. changing the openapi spec. You have to update the `/metrics` endpoint specification by adding 404 in the possible responses with a corresponding example (we do the same for other endpoints you can take inspiration from).
2. related to point one we usually have a content when there are errors by building an `HttpBridgeError` instance and then sending it. You can find several example of its usage across the codebase. It could be something like this:

```java
HttpBridgeError error = new HttpBridgeError(HttpResponseStatus.NOT_FOUND.code(), "put some useful message here to state that metrics are disabled");
HttpUtils.sendResponse(routingContext, HttpResponseStatus.NOT_FOUND.code(),
                        BridgeContentType.KAFKA_JSON, JsonUtils.jsonToBytes(error.toJson()));
```

ppatierno (changes requested):

ppatierno (approved):
LGTM. Thanks Fede and sorry for all my requests! :-)

ppatierno:
@scholzj do you have anything to add here? so we can proceed with merging it?

scholzj:
> @scholzj do you have anything to add here? so we can proceed with merging it?

You do not have the sufficient number of approvals, so you cannot proceed with merging it regardless of what I think.

documentation/book/api/index.adoc
```diff
@@ -1386,6 +1386,7 @@ Retrieves the bridge metrics in Prometheus format.
 ===== Content Type
 
 * text/plain
+* application/json
```
scholzj, line 1389: Do we support JSON based metrics now? How?

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
++
+.Configuration for enabling metrics
+
+[source,properties]
+----
+bridge.metrics=strimziMetricsReporter
+----
++
+Optionally, you can set a comma-separated list of regexes used to filter exposed metrics using the `kafka.prometheus.metrics.reporter.allowlist` property.
+If not set, all available metrics will be exposed.
```
scholzj, line 23: Didn't we discussed in the past that we will have some more reasonable default than `.*`?

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+When needed, it is possible to configure the allowlist per client type.
+For example, set `kafka.admin.prometheus.metrics.reporter.allowlist=` to exclude all admin client metrics.
+
+In case the same property is set with multiple prefixes, the most specific always wins.
+For example, `kafka.producer.prometheus.metrics.reporter.allowlist` wins over `kafka.prometheus.metrics.reporter.allowlist`. 
+
+You can add any plugin configuration to the Bridge properties file using `kafka.`, `kafka.admin.`, `kafka.producer.`, and `kafka.consumer.` prefixes.
+See the https://github.com/strimzi/metrics-reporter[Strimzi Metrics Reporter] documentation for more details.
```
scholzj, line 32: What exactly does this mean, given we have only one set of metrics on the output?

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -88,70 +72,105 @@ public static void main(String[] args) {
...
      * @return                      Future for the bridge startup
      */
-    private static Future<HttpBridge> deployHttpBridge(Vertx vertx, BridgeConfig bridgeConfig, MetricsReporter metricsReporter)  {
+    private static Future<HttpBridge> deployHttpBridge(BridgeConfig bridgeConfig) 
+            throws MalformedObjectNameException, IOException {
         Promise<HttpBridge> httpPromise = Promise.promise();
 
-        HttpBridge httpBridge = new HttpBridge(bridgeConfig, metricsReporter);
+        Vertx vertx = createVertxInstance(bridgeConfig);
+        MetricsCollector metricsCollector = getMetricsCollector(bridgeConfig);
```
scholzj, line 86: What is the reason for creating the `metricsCollector` here and not inside `HttpBridge`? It seems to have the BridgeConfig instance and all what it is used for here seems to be the log message which can be surely embedded there as well. (I guess there might be something, but it does not seem obvious)

src/main/java/io/strimzi/kafka/bridge/Application.java
scholzj: I don't understand why this cannot cause exit. There is a different between the config file path not being set and being set to an non-existing file. I don't think you have to treat it the same way for both cases.

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -55,21 +67,35 @@ public HttpConfig getHttpConfig() {
...
+    private static void validateAndApplyDefaults(Map<String, Object> map) {
+        if (map.get(METRICS_TYPE) != null) {
+            // get and validate metrics type
+            MetricsType metricsType = MetricsType.fromString((String) map.get(METRICS_TYPE));
+            // apply default Strimzi Metrics Reporter configurations if not present
+            if (metricsType == MetricsType.STRIMZI_REPORTER) {
+                map.putIfAbsent("kafka.metric.reporters", "io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter");
+                map.putIfAbsent("kafka.prometheus.metrics.reporter.listener.enable", "false");
+                map.putIfAbsent("kafka.prometheus.metrics.reporter.allowlist", ".*");
+            }
```
scholzj, line 95: Is this really the best place how to handle it? To inject the configuration? As we will not be able to distinguish between the user-provided and defaults in logs etc. If nothing else, an info/warning message should be used here.

ppatierno:
I think you totally mislead what I meant. I just meant if you have anything to add which means review, leaving comments and/or approving. Then we can proceed. I wasn't going to merge without another approval, I know the rules. But because you reviewed this PR already I was just asking if you had anything more.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -558,10 +557,16 @@ private void openapi(RoutingContext routingContext) {
...
+        if (metricsCollector != null) {
+            routingContext.response()
                 .putHeader("Content-Type", "text/plain; version=0.0.4; charset=utf-8")
                 .setStatusCode(HttpResponseStatus.OK.code())
-                .end(metricsReporter.scrape());
+                .end(metricsCollector.scrape());
+        } else {
+            HttpBridgeError error = new HttpBridgeError(HttpResponseStatus.NOT_FOUND.code(), "The metrics endpoint is not enabled.");
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.NOT_FOUND.code(),
+                BridgeContentType.KAFKA_JSON, JsonUtils.jsonToBytes(error.toJson()));
```
scholzj, line 568: What exactly is the JSON we return here? Is it compatibile with the metrics/monitoring tools? Would HTTP 404 without any message be more sensible?

ppatierno:
Adding that even if any other maintainer would have approved it, I was anyway going to ask you if you had anything to add despite I had all the rights to merge.

src/main/java/io/strimzi/kafka/bridge/metrics/MetricsCollector.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+public abstract class MetricsCollector {
+    private final PrometheusMeterRegistry vertxRegistry;
+    
+    MetricsCollector() {
+        this.vertxRegistry = (PrometheusMeterRegistry) BackendRegistries.getDefaultNow();
+        if (vertxRegistry != null) {
+            this.vertxRegistry.config().namingConvention(new MetricsNamingConvention());
+        }
+    }
```
scholzj, line 24: Why exactly does this do? Maybe some Javadoc would be helpful?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: That's a good point. I asked @fvaleri to do the same way as we do for other endpoints but in this case returning a JSON doesn't make sense because expected content for metrics endpoint is just text/plain. So I agree just 404 with empty body is better (as first proposal by Fede).

documentation/book/api/index.adoc
ppatierno: See my reply to your other comment. When JSON will be removed from the openapi spec as well, it will be removed from here.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
ppatierno: The Strimzi metrics reporter plugin configuration is a per-client configuration.
The bridge has 3 clients, admin, producer and consumer and you can apply a configuration to all of them by using the `kafka.` prefix or a specific config for each client using the corresponding prefix (i.e. if you want a specific config on the consumer only, you set it with the `kafka.consumer.` prefix).
For example in the above description, you could have an allowList different for admin, producer and consumer by setting it with the appropriate prefix, for example:

````
kafka.consumer.prometheus.metrics.reporter.allowlist=<some list>"
kafka.producer.prometheus.metrics.reporter.allowlist=<some other list>"
```

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
fvaleri: That was discussed as part of the operator changes, but I think we can also apply that logic here.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
fvaleri: For example, I could only be interested in admin client metrics for troubleshooting purpose, so I would set producer and consumer to empty allow list.

config/application.properties
```diff
@@ -1,5 +1,16 @@
 #Bridge related settings
 bridge.id=my-bridge
+
+# uncomment the following line to enable Prometheus JMX Exporter, check the documentation for more details
+#bridge.metrics=jmxPrometheusExporter
+# optionally, set the file path of your custom configuration
+#bridge.metrics.exporter.config.path=/path/to/my-exporter-config.yaml
+
+# uncomment the following line to enable Strimzi Metrics Reporter, check the documentation for more details
+bridge.metrics=strimziMetricsReporter
```
ppatierno, line 10: I guess it was pushed by mistake? It should be commented.

src/main/java/io/strimzi/kafka/bridge/http/HttpConfig.java
```diff
@@ -83,7 +83,7 @@ public long getConsumerTimeout() {
      * @return if CORS is enabled
      */
     public boolean isCorsEnabled() {
-        return Boolean.parseBoolean(this.config.getOrDefault(HTTP_CORS_ENABLED, false).toString());
+        return Boolean.valueOf(this.config.getOrDefault(HTTP_CORS_ENABLED, false).toString());
```
ppatierno, line 86: why this change?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -558,10 +608,14 @@ private void openapi(RoutingContext routingContext) {
...
     private void metrics(RoutingContext routingContext) {
-        routingContext.response()
+        if (metricsCollector != null) {
+            routingContext.response()
                 .putHeader("Content-Type", "text/plain; version=0.0.4; charset=utf-8")
                 .setStatusCode(HttpResponseStatus.OK.code())
-                .end(metricsReporter.scrape());
+                .end(metricsCollector.scrape());
+        } else {
+            HttpUtils.sendResponse(routingContext, HttpResponseStatus.NOT_FOUND.code(), null, null);
```
ppatierno, line 617: I saw you removed the 404 response from the openapi spec. I think we should keep it but just with a "description" field and no "content" section. Something similar happens for some 204 OK NO CONTENT responses.

ppatierno (changes requested):

ppatierno (changes requested):

ppatierno (approved):
LGTM. I left just one (hopefully last) nit. Thanks!

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -6,23 +6,39 @@
...
+
+    /** JMX Exporter configuration file path */
+    private static final String JMX_EXPORTER_CONFIG_PATH = METRICS_TYPE + ".exporter.config.path";
+    
     /** Tracing system to be used in the bridge */
-    public static final String TRACING_TYPE = BRIDGE_CONFIG_PREFIX + "tracing";
+    private static final String TRACING_TYPE = BRIDGE_CONFIG_PREFIX + "tracing";
+    
+    /** Default Strimzi Metrics Reporter allow list. */
+    /* test */ static final String DEFAULT_METRICS_ALLOW_LIST = "kafka_consumer_consumer_metrics.*, " +
```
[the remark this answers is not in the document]
ppatierno: but when you use the constant and you are not on the declaration, it makes more understandable imho. We use to be more explicit in all constants we have across Strimzi even if it means longer names :-)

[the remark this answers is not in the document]
fvaleri: Then I guess it should be `DEFAULT_STRIMZI_METRICS_REPORTER_ALLOW_LIST` right?

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
scholzj: Hmm, maybe this would deserve some clarification in the docs. Especially the `the most specific always wins` could be confusing as you can also take it to the list of the different expressions etc.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
scholzj: So, we now have some different defaults in the code so this should be updated as well as I guess it is not all available metrics anymore?

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
```diff
@@ -6,23 +6,39 @@
...
+    /** JMX Exporter configuration file path */
+    private static final String JMX_EXPORTER_CONFIG_PATH = METRICS_TYPE + ".exporter.config.path";
+    
     /** Tracing system to be used in the bridge */
-    public static final String TRACING_TYPE = BRIDGE_CONFIG_PREFIX + "tracing";
+    private static final String TRACING_TYPE = BRIDGE_CONFIG_PREFIX + "tracing";
+    
+    /** Default Strimzi Metrics Reporter allow list. */
+    /* test */ static final String DEFAULT_STRIMZI_METRICS_REPORTER_ALLOW_LIST = "kafka_consumer_consumer_metrics.*, " +
+        "kafka_producer_kafka_metrics_count_count, kafka_producer_producer_metrics.*";
```
scholzj, line 41: More out of curiosity than anything else ... what does `kafka_producer_kafka_metrics_count_count` show?

scholzj (commented):
Few more nits and questions -> but nothing major anymore.

ppatierno:
@fvaleri other than resolving the remaining comments could you also rebase please? We have the fix for the Travis CI pipeline in main now. Thanks!

src/main/java/io/strimzi/kafka/bridge/config/BridgeConfig.java
fvaleri: It's Kafka producer instances, used by the operator's Bridge example dashboard.

scholzj (approved):
Thanks. LGTM.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+----
++
+Optionally, you can configure a comma-separated list of regexes to filter exposed metrics using the `kafka.prometheus.metrics.reporter.allowlist` property.
+If not configured, a default set of metrics will be exposed.
+
+When needed, it is possible to configure the allowlist per client type.
+For example, set `kafka.admin.prometheus.metrics.reporter.allowlist=` to exclude all admin client metrics.
+
+In case the same property is set with multiple prefixes, the most specific prefix wins.
+For example, `kafka.producer.prometheus.metrics.reporter.allowlist` wins over `kafka.prometheus.metrics.reporter.allowlist`.
```
PaulRMellor, line 29:
```suggestion
+
In the event that the same property is set with multiple prefixes, the most specific prefix takes precedence.
For example, `kafka.producer.prometheus.metrics.reporter.allowlist` takes precedence over `kafka.prometheus.metrics.reporter.allowlist`. 
```

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+Optionally, you can configure a comma-separated list of regexes to filter exposed metrics using the `kafka.prometheus.metrics.reporter.allowlist` property.
+If not configured, a default set of metrics will be exposed.
+
+When needed, it is possible to configure the allowlist per client type.
+For example, set `kafka.admin.prometheus.metrics.reporter.allowlist=` to exclude all admin client metrics.
+
+In case the same property is set with multiple prefixes, the most specific prefix wins.
+For example, `kafka.producer.prometheus.metrics.reporter.allowlist` wins over `kafka.prometheus.metrics.reporter.allowlist`. 
+
+You can add any plugin configuration to the Bridge properties file using `kafka.`, `kafka.admin.`, `kafka.producer.`, and `kafka.consumer.` prefixes.
```
PaulRMellor, line 31: should we move this above "If required, you can configure an `allowlist` for a specific client type."  as we use `kafka.admin' in the example there?

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+[source,properties]
+----
+bridge.metrics=strimziMetricsReporter
+----
++
+Optionally, you can configure a comma-separated list of regexes to filter exposed metrics using the `kafka.prometheus.metrics.reporter.allowlist` property.
+If not configured, a default set of metrics will be exposed.
+
+When needed, it is possible to configure the allowlist per client type.
```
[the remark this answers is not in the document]
PaulRMellor: We need some `+` adds to bring this content inline with step 1.

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
PaulRMellor, line 31:
```suggestion
+
You can add any plugin configuration to the Bridge properties file using `kafka.`, `kafka.admin.`, `kafka.producer.`, and `kafka.consumer.` prefixes.
```

PaulRMellor (commented):
Thanks Fede.
I left a few comments on the docs.

documentation/assemblies/assembly-kafka-bridge-config.adoc
```diff
@@ -8,11 +8,13 @@
 [role="_abstract"]
 Configure a deployment of the Kafka Bridge using configuration properties.
 Configure Kafka and specify the HTTP connection details needed to be able to interact with Kafka.
+It is possible to enable metrics in Prometheus format with Prometheus JMX Exporter or Strimzi Metrics Reporter.
```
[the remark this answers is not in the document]
fvaleri: There are, I'm adding a links so the user can find them easily.

fvaleri:
@PaulRMellor thanks for the great suggestions. Let me know if there is still some doc improvement or issue.

documentation/modules/proc-configuring-kafka-bridge-jmx-metrics.adoc
```diff
@@ -0,0 +1,33 @@
...
+. Set the `bridge.metrics` configuration to `jmxPrometheusExporter`.
++
+.Configuration for enabling metrics
+
+[source,properties]
+----
+bridge.metrics=jmxPrometheusExporter
+----
++
+Optionally, you can configure a custom Prometheus JMX Exporter configuration using the `bridge.metrics.exporter.config.path` property.
```
PaulRMellor, line 22:
```suggestion
Optionally, you can add a custom Prometheus JMX Exporter configuration using the `bridge.metrics.exporter.config.path` property.
```

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+[source,properties]
+----
+bridge.metrics=strimziMetricsReporter
+----
++
+Optionally, you can configure a comma-separated list of regular expressions to filter exposed metrics using the `kafka.prometheus.metrics.reporter.allowlist` property.
+If not configured, a default set of metrics is exposed.
+
++
+When needed, it is possible to configure the allowlist per client type.
```
PaulRMellor, line 26:
```suggestion
When needed, it is possible to configure the `allowlist` per client type.
```

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+For example, using the `kafka.admin` prefix, setting `kafka.admin.prometheus.metrics.reporter.allowlist=` excludes all admin client metrics.
```
PaulRMellor, line 27:
```suggestion
For example, by using the `kafka.admin` prefix and setting `kafka.admin.prometheus.metrics.reporter.allowlist=`, all admin client metrics are excluded.
```

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+
++
+You can add any plugin configuration to the Bridge properties file using `kafka.`, `kafka.admin.`, `kafka.producer.`, and `kafka.consumer.` prefixes.
```
PaulRMellor, line 30:
```suggestion
You can add any plugin configuration to the Kafka Bridge properties file using `kafka.`, `kafka.admin.`, `kafka.producer.`, and `kafka.consumer.` prefixes.
```

documentation/modules/proc-configuring-kafka-bridge-smr-metrics.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+. Run the Kafka Bridge run script.
++
+.Running the Kafka Bridge
+[source,shell]
+----
+./bin/kafka_bridge_run.sh --config-file=<path>/application.properties
+----
++
+With metrics enabled, you can scrape metrics in the Prometheus format from the `/metrics` endpoint of the Kafka Bridge.
```
PaulRMellor, line 42:
```suggestion
With metrics enabled, you can scrape metrics in Prometheus format from the `/metrics` endpoint of the Kafka Bridge.
```

PaulRMellor (approved):
Thanks for dressing the comments, Fede.
Just a few nits as I re-read.

ppatierno:
@fvaleri thanks again for your work!

PR #955, opened 2024-12-19
Using JDK 17

ppatierno:
This PR moves the project to use Maven compiler 17 (from 11) and to use JDK 17 (removing JDK) 11 from pipelines.

### DIFF
--- a/.azure/templates/jobs/build_docs.yaml
+++ b/.azure/templates/jobs/build_docs.yaml
@@ -3,9 +3,9 @@ jobs:
     displayName: 'Build Docs'
     strategy:
       matrix:
-        'java-11':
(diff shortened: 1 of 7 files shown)


scholzj (approved):
The changes seem good. But this is certanly worth some CHANGELOG entry and possibly some docs updare if the docs cover the Java support?

see-quick (approved):

im-konge (approved):

ppatierno:
> The changes seem good. But this is certanly worth some CHANGELOG entry and possibly some docs updare if the docs cover the Java support?

@scholzj I updated the CHANGELOG but didn't find any reference to supported Java version within the documentation.

PR #963, opened 2025-01-14
Removed the script to generate the configuration

ppatierno:
This PR is part of the work for the operator issue https://github.com/strimzi/strimzi-kafka-operator/issues/10959 and remove the script to prepare the bridge configuration when running on container because it's done by the operator by making a ConfigMap then mounted on the bridge pod.


ppatierno:
This PR comes together with https://github.com/strimzi/strimzi-kafka-operator/pull/11032

scholzj (approved):
LGTM. Thanks. Should we merge it so that the other PR can use `:latest` container images?

Or maybe ... have you considered having for some time both mechanisms for better backwards compatibility?

ppatierno:
> LGTM. Thanks. Should we merge it so that the other PR can use :latest container images?

Yeah. I am addressing feedback on that PR first and back to this to merge.

> Or maybe ... have you considered having for some time both mechanisms for better backwards compatibility?

Tbh I think it's much effort to take both mechanisms mostly from the operator perspective to configure in both way depending on the version. I don't think it's worth it. As commented on the other PR we can add a line in the "Major changes, deprecations and removals" section for the operator CHANGELOG to highlight the bridge version to be used.

scholzj:
> > Or maybe ... have you considered having for some time both mechanisms for better backwards compatibility?
>
> Tbh I think it's much effort to take both mechanisms mostly from the operator perspective to configure in both way depending on the version. I don't think it's worth it. As commented on the other PR we can add a line in the "Major changes, deprecations and removals" section for the operator CHANGELOG to highlight the bridge version to be used.

I meant only the Bridge here -> to be able to use new bridge with old operator. I thought it might be useful for the 0.45 patch releases But I guess you are right that it is quite a lot of effort for very little gain. So I'm completely fine to have this merged without it.

ppatierno:
> LGTM. Thanks. Should we merge it so that the other PR can use :latest container images?

I was chatting with @im-konge about this ...
If we merge this PR and we have a new bridge "latest" image, it won't be used when running the STs (i.e. regression, ...) of new PRs in the operator repo because the Azure pipeline uses `BRIDGE_IMAGE: "latest-released"` configuration which means still using latest 0.31.1 bridge release.
So one way to merge the corresponding PR https://github.com/strimzi/strimzi-kafka-operator/pull/11032 could be including (in the same PR) a change with `BRIDGE_IMAGE: "latest"` so it will use the "latest" (not released yet) bridge.
But this change should be brought back to "latest-released" when our 0.32.0 bridge will be out.
Wdyt?

scholzj:
@ppatierno I would expect the STs use by default whatever Bridge version is in the installation files. At the end that is how it knows to use the 0.31.1 version right now. So hen you change it to `:latest` in the installation files, why would the STs not use it?

ppatierno:
Which installation files are you referring to? From my understanding it seems that the bridge STs configuration sticks with "latest-released" maybe taking the version from the bridge.version file. So I had to tweak that YAML replacing "latest-released" with a specific image or I should set "latest" there. Is my understanding right @im-konge ?

scholzj:
> Which installation files are you referring to?

Those in packaging. Just set it there to `latest` and the STs should use that IMHO.

scholzj:
TBH, 

> This could work if changing the 060 Deployment YAML for all STs but it won't work for the ones using the Helm Charts if you see my latest comment on the operator repo PR.

The Helm Charts are the source for the 060-... file. So you need to change it there in the first place. Not sure wh would the STs not respect the Helm Chart. IMHO that is where they get the 0.31.1 version as well by just using whatever is there. I mean, I have not tried any of that - so a lot depends on what the STs do on top of the installation files of course.

ppatierno:
> IMHO that is where they get the 0.31.1 version as well by just using whatever is there

This is not what I get discussing with @im-konge
Lukas, any thoughts?

im-konge:
As I mentioned during our offline conversation, the Helm installation is working the same way for the operator (meaning when you change the values.yaml, it will be respected for the operator), but for the Bridge the whole config is "hard-coded" in the STs:

```
        // image registry config
        values.put("defaultImageRegistry", Environment.getIfNotEmptyOrDefault(Environment.STRIMZI_REGISTRY, Environment.STRIMZI_REGISTRY_DEFAULT));
        values.put("kafkaBridge.image.registry", Environment.STRIMZI_REGISTRY_DEFAULT);

        // image repository config
        values.put("defaultImageRepository", Environment.getIfNotEmptyOrDefault(Environment.STRIMZI_ORG, Environment.STRIMZI_ORG_DEFAULT));
        values.put("kafkaBridge.image.repository", Environment.STRIMZI_ORG_DEFAULT);

        // image tags config
        values.put("defaultImageTag", Environment.getIfNotEmptyOrDefault(Environment.STRIMZI_TAG, Environment.STRIMZI_TAG_DEFAULT));
        values.put("kafkaBridge.image.tag", BridgeUtils.getBridgeVersion());
```

The `BridgeUtils.getBridgeVersion()` takes the version from the `bridge.version` file. So in case that the Bridge PR is merged, I think that you can change the version in `bridge.version` file to `latest` and STs will take that as a version for the HelmChartST (as well as for other tests).

im-konge:
It's done this way, because nobody really needed to change the image for Bridge in the past -> but you already created the issue for having a possibility to configure the Bridge image even for the Helm (and I will look into it).

ppatierno:
After an offline chat with Lukas I am going to merge this one so we have the official Strimzi bridge "latest" image.
To have tests working in the operator PR the thing we can do is changing the `bridge.version` file (as Lukas suggested in order to have the Helm Charts STs working), run the `make bridge_version` and push changes. This way we'll also have the values.yaml and the deployment YAML being changed to use the "latest" bridge and STs should run ok.

PR #966, opened 2025-02-10
Prepare for Vert.x 5

ppatierno:
This PR prepares the code base for Vert.x 5. This should allow us to simplify testing of new Vert.x 5 releases.
It mostly moves from the usage of the callback model (deprecated by Vert.x 5) to the future model.
It also updates Micrometer Metrics to use a version used by the current Vert.x 4 version we use.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -119,15 +119,16 @@ public class Application {
         Promise<HttpBridge> httpPromise = Promise.promise();
 
         HttpBridge httpBridge = new HttpBridge(bridgeConfig, metricsReporter);
-        vertx.deployVerticle(httpBridge, done -> {
-            if (done.succeeded()) {
-                LOGGER.info("HTTP verticle instance deployed [{}]", done.result());
-                httpPromise.complete(httpBridge);
-            } else {
-                LOGGER.error("Failed to deploy HTTP verticle instance", done.cause());
-                httpPromise.fail(done.cause());
-            }
-        });
+        vertx.deployVerticle(httpBridge)
+                .onComplete(done -> {
+                    if (done.succeeded()) {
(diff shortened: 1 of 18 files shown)


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -119,15 +119,16 @@ private static Future<HttpBridge> deployHttpBridge(Vertx vertx, BridgeConfig bri
...
+        vertx.deployVerticle(httpBridge)
+                .onComplete(done -> {
+                    if (done.succeeded()) {
+                        LOGGER.info("HTTP verticle instance deployed [{}]", done.result());
+                        httpPromise.complete(httpBridge);
+                    } else {
+                        LOGGER.error("Failed to deploy HTTP verticle instance", done.cause());
+                        httpPromise.fail(done.cause());
+                    }
+                });
```
katheris, line 131: Here and elsewhere if we are using if (succeeeded) {} else {} why not use `onSuccess` and `onFailure` instead?
```suggestion
        vertx.deployVerticle(httpBridge)
                .onSuccess(result -> {
                        LOGGER.info("HTTP verticle instance deployed [{}]", result);
                        httpPromise.complete(httpBridge);
                }).onFailure(cause -> {
                        LOGGER.error("Failed to deploy HTTP verticle instance", cause);
                        httpPromise.fail(cause);
                });
```

ppatierno: That's a good point. It's just kind of "old" code that we can refactoring the way you suggested.
Are you ok if I get this suggestion for a separate PR? It's not strictly related to the Vert.x 5 preparation.

katheris: Yeah that makes sense

katheris (approved):

ppatierno:
Going to merge this one. The Travis CI failure is not related to this update and some folks for the ppc and s390x platforms are already looking into it.

PR #971, opened 2025-02-21
Renamed Prometheus version property in pom.xml

ppatierno:
Trivial PR to rename the pom property to define the Prometheus version. It was something like prometheus-client because we were including the Prometheus client time ago, but meanwhile Prometheus changed and the client is not brought anymore in the bridge. Renaming to just `prometheus.version` makes more sense to me.


scholzj (approved):

see-quick (approved):

im-konge (approved):

ppatierno:
Going to merge. Travis has queue issues with ppc architecture.

PR #973, opened 2025-02-21
Refactored Future completion with onSuccess and onFailure

ppatierno:
In one of the latest PRs a comment was raised about using the `onSuccess` + `onFailure` methods pattern instead of the `onComplete` method.
This PR applies this refactoring leaving some tests as they are (using `onComplete`) because the failure check is part of the assertion so splitting doesn't make really sense.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -50,12 +50,15 @@ public class Application {
             BridgeConfig bridgeConfig = BridgeConfig.fromMap(config);
             LOGGER.info("Bridge configuration {}", bridgeConfig);
 
-            deployHttpBridge(bridgeConfig).onComplete(done -> {
-                if (done.succeeded()) {
-                    // register tracing - if set, etc
-                    TracingUtil.initialize(bridgeConfig);
-                }
-            });
+            deployHttpBridge(bridgeConfig)
+                    .onSuccess(httpBridge -> {
+                        // register tracing - if set, etc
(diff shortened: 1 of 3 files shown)


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -50,12 +50,15 @@ public static void main(String[] args) {
...
-            });
+            deployHttpBridge(bridgeConfig)
+                    .onSuccess(httpBridge -> {
+                        // register tracing - if set, etc
+                        TracingUtil.initialize(bridgeConfig);
+                    })
+                    .onFailure(t -> {
+                        LOGGER.error("Error starting the bridge", t);
+                        System.exit(1);
+                    });
```
scholzj, line 61: What would happen if you leave out the `onFailure` clause or if you just throw an RuntimeException here? The duplicated code here looks a bit ugly, so I wonder if it is really needed.

ppatierno: Before we had the creation of the "metrics" related stuff which could drive to some exception inside the main so caught by the `catch` you see here.
We moved that creation inside the `deployHttpBridge` (mostly inside the `HttpBridge` creation) and errors would be caught the same way but ... I noticed that the `vertx.deployVerticle` (inside `deployHttpBridge`) could fail the future and this would be caught by the `onFailure` I just added but it was not caught before (the `onComplete` was just ignoring the case of errors) and I guess it was a mistake.

scholzj (approved):

PR #986, opened 2025-05-27
Update to Vert.x 5

ppatierno:
This PR is about bumping the Vert.x version to the latest 5.0.0 (together with Netty).
Due to a big refactoring within the Vert.x OpenAPI component within 5.0.0 release, this PR addresses the two following proposals:

* [Added proposal about additional schema validation errors in the bridge with Vert.x 5.x](https://github.com/strimzi/proposals/pull/156)
* [Added proposal to move HTTP bridge using OpenAPI 3.1.0](https://github.com/strimzi/proposals/pull/157)

The first one is about adding detailed information related to schema validation errors within the JSON object returned by the bridge.
The second one is about bumping the OpenAPI specification to 3.1.0 because of then unsupported "nullable" (which is not a JSON Schema standard construct) within Vert.x 5.0.0 OpenAPI component and replaced by the JSON Schema standard "null" type.
For more details please refers to the proposals.


im-konge (approved):
LGTM, thanks :)

PR #989, opened 2025-06-02
Improve lifecycle of Integration tests

see-quick:
This PR improves our Integration tests lifecycle. It re-implements the approach to **not USE** shared objects during the whole test execution. Therefore, we decided to create new instances of Kafka cluster, Bridge, etc., for each test suite. For edge cases where a different configuration is needed, I have added overridable hooks.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
@@ -34,6 +34,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
```diff
@@ -183,4 +159,102 @@ protected String generateRandomConsumerGroupName() {
...
+     * &#064;Override
+     * protected Map&lt;String, Object&gt; overridableConfig() {
+     *     Map&lt;String, Object&gt; overrides = new HashMap&lt;&gt;();
+     *     overrides.put("bridge.id", "my-custom-bridge");
+     *     return overrides;
+     * }
+     * Return a map with configuration keys to override default values,
+     * or keys mapped to {@code null} to remove them.
+     */
+    protected Map<String, Object> overridableConfig() {
```
ppatierno: why not just `overrideConfig`?

ppatierno (approved):
I tried it locally and works fine (with podman). I left a nit on a method name.

src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
see-quick: Yeah, that's better. I will update it.

PR #994, opened 2025-06-12
[release-0.31.x] Move to the Sonatype Central Portal

ppatierno:
This PR adds the changes needed for migrating to the new Central portal on the `release-0.31.x` branch.
It also adds the release of artifacts within the release pipeline in order to release RCs in Maven Central (due to the missing staging area in the new Sonatype system).

### DIFF
--- a/.azure/templates/jobs/release_artifacts.yaml
+++ b/.azure/templates/jobs/release_artifacts.yaml
@@ -1,0 +1,33 @@
+jobs:
+  - job: 'release_artifacts'
+    displayName: 'Prepare and release artifacts'
+    # Set timeout for jobs
+    timeoutInMinutes: 60
+    # Base system
+    pool:
+      vmImage: 'Ubuntu-22.04'
+    # Pipeline steps
+    steps:
+      # Install Prerequisites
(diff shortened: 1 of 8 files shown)


.azure/templates/jobs/release_artifacts.yaml
```diff
@@ -0,0 +1,33 @@
...
+    # Base system
+    pool:
+      vmImage: 'Ubuntu-22.04'
+    # Pipeline steps
+    steps:
+      # Install Prerequisites
+      - template: '../steps/prerequisites/install_java.yaml'
+
+      # Change the release version
+      - bash: "mvn versions:set -DnewVersion=$(shell echo $(RELEASE_VERSION) | tr a-z A-Z)"
```
im-konge:
```suggestion
      - bash: "mvn versions:set -DnewVersion=$(echo $RELEASE_VERSION | tr a-z A-Z)"
```
actually it should be like this. (I just found it during testing test-containers changes)

ppatierno: Good catch. Fixed!

see-quick (approved):

PR #1000, opened 2025-06-22
Increased retries and delay for assigned partitions in STs

ppatierno:
Fixing on main the retries and delay used when waiting for partitions assignments during STs. It was already done on the `release-0.31.x` branch (for the release).
The current values drive to a too fast waiting 5 x 200 ms = 1000 ms and in just one second the assignment could not happen easily.


see-quick (approved):
Good catch @ppatierno. Thanks 👍 . I wonder if we should try run build pipeline a few times to double-check if this value helps. WDYT?

ppatierno:
They make more sense than the before but as you can see failures can still happen. I think that the ST needs more investigation because it's actually very weird.

Issue #690, opened 2022-11-14
Adding metrics related tests


ppatierno:
It seems that we are missing tests about checking that metrics are exposed correctly on the `/metrics` endpoint.
With missing tests we didn't notice that at some point metrics were broken, more details in #688 
We need to add these tests for checking the `/metrics` endpoint is exposed and metrics are available out of it.

PR #1012, opened 2025-08-19
Added more metrics related ITs

ppatierno:
This PR fixes https://github.com/strimzi/strimzi-kafka-bridge/issues/690 finally adding some ITs related to metrics to check they are exposed correctly with the expected content.


see-quick (approved):
LGTM 👍

PR #1029, opened 2025-10-02
Removed System.out statements from ProducerIT tests to clean up test output.

svsaiabitha:
This PR fixes #1029. Remove System.out statements from ProducerIT tests

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
@@ -132,7 +132,7 @@ public class ProducerIT extends HttpBridgeITAbstract {
                             HttpResponse<JsonObject> response = ar.result();
                             assertThat(response.statusCode(), is(HttpResponseStatus.OK.code()));
                             JsonObject bridgeResponse = response.body();
-                            LOGGER.info("Recieved response:\n{}", bridgeResponse);
+                            LOGGER.info("Received response:\n{}", bridgeResponse);
                             JsonArray offsets = bridgeResponse.getJsonArray("offsets");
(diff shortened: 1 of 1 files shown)


ppatierno:
@svsaiabitha thanks for the PR. A couple of things:

* please fix the DCO: you can find information about how to do that by clicking on "DCO" in the above failed step.
* please always have a description for the PR. In this case it also fixes a specific issue so you can use something like "This PR fixes #<issue number>" because this helps the GitHub automation to close the issue when the PR is merged.

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
```diff
@@ -132,7 +132,7 @@ void sendMessagesToMultiplePartitions(VertxTestContext context) throws Interrupt
                             HttpResponse<JsonObject> response = ar.result();
                             assertThat(response.statusCode(), is(HttpResponseStatus.OK.code()));
                             JsonObject bridgeResponse = response.body();
-                            System.out.println(bridgeResponse);
+                            LOGGER.info("Recieved response:\n{}", bridgeResponse);
```
ppatierno: There is a typo here, it should be "Received"

ppatierno (changes requested):

src/test/java/io/strimzi/kafka/bridge/http/ProducerIT.java
svsaiabitha: Thanks for catching that! I’ve fixed the typo.

ppatierno (approved):
LGTM. Thank you!

see-quick (approved):
Thanks 👍

PR #1040, opened 2025-10-31
Implement support for TLS/SSL on the HTTP interface

### LINKED ISSUE #939: Support TLS/SSL on the HTTP interface
Wanted to ask if there were any plans to support TLS/SSL on the HTTP interface. Looking into the code, this is apparently not implemented, https://github.com/strimzi/strimzi-kafka-bridge/blob/main/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java#L266  

Could it be implemented?

tinaselenge:
Implements the proposal https://github.com/strimzi/proposals/pull/171. 
In the proposal we agreed to default port 443, however, after further discussion, it was decided to change it to 8443.  This change makes more sense as the default port for non SSL server is 8080 and most Linux systems, by default, do not allow non-root users to bind to ports below 1024.

Closes #939

### DIFF
--- a/config/log4j2.properties
+++ b/config/log4j2.properties
@@ -25,3 +25,7 @@ logger.healthy.name = http.openapi.operation.healthy
 logger.healthy.level = WARN
 logger.ready.name = http.openapi.operation.ready
 logger.ready.level = WARN
+
+# Reduce verbosity of RouterBuilderImpl warnings for unimplemented OpenAPI endpoints because bridge servers share the same OpenAPI contract with different endpoints implemented.
+logger.vertx.name = io.vertx.ext.web.openapi.router.impl.RouterBuilderImpl
+logger.vertx.level = ERROR
--- a/documentation/modules/proc-configuring-http-bridge.adoc
+++ b/documentation/modules/proc-configuring-http-bridge.adoc
@@ -65,12 +65,12 @@ For example:
 ----
 bridge.id=my-bridge
 http.host=0.0.0.0
-http.port=443 <5>
+http.port=8443 <5>
 http.ssl.enable=true <6>
 http.ssl.keystore.certificate.location=/etc/ssl/certs/bridge.crt <7>
 http.ssl.keystore.key.location=/etc/ssl/private/bridge.key <8>
 ----
-<5> Configure HTTP Bridge to listen on port 443 for TLS encrypted connections.
+<5> Configure HTTP Bridge to listen on port 8443 for TLS encrypted connections.
 <6> Set `true` to enable TLS encryption between HTTP clients and the HTTP Bridge.
 <7> Location of the keystore file in PEM format.
 <8> Location of the private key in PEM format.
@@ -85,7 +85,7 @@ http.ssl.keystore.key.location=/etc/ssl/private/bridge.key <8>
 |HTTP Bridge host address. The default is `0.0.0.0`.
 
 |http.port
-|HTTP Bridge port. The default is `8080`. If "http.ssl.enable" is set to `true`, then the default is `443`.
+|HTTP Bridge port. The default is `8080`. If "http.ssl.enable" is set to `true`, then the default is `8443`.
 
 |http.management.port
 |HTTP bridge port for management endpoints such as `/healthy`, `/ready` and `/metrics`. The default is `8081`.
--- a/src/main/java/io/strimzi/kafka/bridge/Application.java
+++ b/src/main/java/io/strimzi/kafka/bridge/Application.java
@@ -21,10 +21,8 @@ import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
-import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.core.config.Configurator;
 
 import javax.management.MalformedObjectNameException;
 import java.io.IOException;
@@ -48,9 +46,6 @@ public class Application {
         // disabling OpenMetrics exemplars added within newer Prometheus client (see https://github.com/strimzi/strimzi-kafka-bridge/issues/1023)
         // via https://github.com/prometheus/client_java/blob/main/prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/MetricsProperties.java#L14
         System.setProperty("io.prometheus.metrics.exemplarsEnabled", "false");
-
-        // Reduce verbosity of RouterBuilderImpl warnings for unimplemented OpenAPI endpoints
-        Configurator.setLevel("io.vertx.ext.web.openapi.router.impl.RouterBuilderImpl", Level.ERROR);
 
         LOGGER.info("Strimzi Kafka Bridge {} is starting", Application.class.getPackage().getImplementationVersion());
         try {
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -336,7 +336,7 @@ public class HttpBridge extends AbstractVerticle {
                     .onFailure(t -> LOGGER.info("Error while shutting down HTTP bridge management server"))
             : Future.succeededFuture();
 
-        Future.all(apiServerShutdown, managementServerShutdown)
+        Future.join(apiServerShutdown, managementServerShutdown)
                 .onSuccess(v -> {
                     LOGGER.info("HTTP bridge has been shut down successfully");
                     stopPromise.complete();
@@ -354,25 +354,18 @@ public class HttpBridge extends AbstractVerticle {
 
         if (this.bridgeConfig.getHttpConfig().isSslEnabled()) {
             httpServerOptions.setSsl(true);
-            boolean missingSslConfigsMissing = false;
 
             if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreLocation() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKeyLocation() != null) {
                 httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
                         .setKeyPath(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKeyLocation())
                         .setCertPath(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreLocation()));
-            } else {
-                missingSslConfigsMissing = true;
-            }
-
-            if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey() != null) {
+            } else if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey() != null) {
                 httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
                         .addKeyValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey()))
                         .addCertValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain())));
             } else {
-                if (missingSslConfigsMissing) {
-                    LOGGER.error("Required SSL configurations are missing! Either both of http.ssl.keystore.location and http.ssl.keystore.key.location " +
-                            "or both of http.ssl.keystore.certificate.chain and http.ssl.keystore.key should be configured");
-                }
+                LOGGER.error("Required SSL configurations are missing! Either both of http.ssl.keystore.location and http.ssl.keystore.key.location " +
+                        "or both of http.ssl.keystore.certificate.chain and http.ssl.keystore.key should be configured");
             }
 
             Set<String> sslEnabledProtocols = this.bridgeConfig.getHttpConfig().getHttpServerSslEnabledProtocols();
--- a/src/test/java/io/strimzi/kafka/bridge/config/ConfigTest.java
+++ b/src/test/java/io/strimzi/kafka/bridge/config/ConfigTest.java
@@ -13,11 +13,13 @@ import org.junit.jupiter.api.Test;
 import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 /**
@@ -33,6 +35,7 @@ public class ConfigTest {
         map.put("kafka.consumer.auto.offset.reset", "earliest");
         map.put("http.host", "0.0.0.0");
         map.put("http.port", "8080");
+        map.put("management.port", "8081");
 
         BridgeConfig bridgeConfig = BridgeConfig.fromMap(map);
         assertThat(bridgeConfig.getBridgeID(), is("my-bridge"));
@@ -51,6 +54,7 @@ public class ConfigTest {
         assertThat(bridgeConfig.getHttpConfig().getConfig().size(), is(2));
         assertThat(bridgeConfig.getHttpConfig().getHost(), is("0.0.0.0"));
         assertThat(bridgeConfig.getHttpConfig().getPort(), is(8080));
+        assertThat(bridgeConfig.getHttpConfig().getManagementPort(), is(8081));
     }
 
     @Test
@@ -83,6 +87,41 @@ public class ConfigTest {
         assertThat(bridgeConfig.getHttpConfig().getCorsAllowedMethods(), is("GET,POST,PUT,DELETE,OPTIONS,PATCH"));
         assertThat(bridgeConfig.getHttpConfig().isConsumerEnabled(), is(true));
         assertThat(bridgeConfig.getHttpConfig().isProducerEnabled(), is(true));
+    }
+
+    @Test
+    public void testHttpSslConfig() {
+        Map<String, Object> map = new HashMap<>(Map.of(
+                "http.ssl.enable", "true",
+                "http.ssl.keystore.key.location", "key.key",
+                "http.ssl.keystore.location", "cert.crt",
+                "http.ssl.enabled.protocols", "TLSv1.3",
+                "http.ssl.enabled.cipher.suites", "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
+        ));
+
+        BridgeConfig bridgeConfig = BridgeConfig.fromMap(map);
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKeyLocation(), is("key.key"));
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslKeystoreLocation(), is("cert.crt"));
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslEnabledProtocols(), is(Set.of("TLSv1.3")));
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslCipherSuites(), is(Set.of("TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384")));
+
+    }
+
+    @Test
+    public void testHttpSslDefaults() {
+        Map<String, Object> map = new HashMap<>(Map.of(
+                "http.ssl.enable", "true",
+                "http.ssl.keystore.key", "key.key",
+                "http.ssl.keystore.certificate.chain", "cert.crt"
+        ));
+
+        BridgeConfig bridgeConfig = BridgeConfig.fromMap(map);
+        assertThat(bridgeConfig.getHttpConfig().getPort(), is(8443));
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey(), is("key.key"));
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain(), is("cert.crt"));
+        assertThat(bridgeConfig.getHttpConfig().getHttpServerSslEnabledProtocols(), is(Set.of("TLSv1.2", "TLSv1.3")));
+        assertNull(bridgeConfig.getHttpConfig().getHttpServerSslCipherSuites());
+
     }
 
     @Test


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -236,19 +251,29 @@ public void start(Promise<Void> startPromise) {
                     this.router.errorHandler(HttpResponseStatus.BAD_REQUEST.code(), this::errorHandler);
                     this.router.errorHandler(HttpResponseStatus.NOT_FOUND.code(), this::errorHandler);
 
+                    RouterBuilder managementRouterBuilder = RouterBuilder.create(vertx, contract);
```
tinaselenge, line 252: @ppatierno do you think we should have separate OpenAPI contracts for each RouterBuilder? because in this PR, I've made them both use the same contract, but with different handlers, it results in some warning messages in the log e.g. :
```
2025-11-05 10:34:57 WARN  [vert.x-eventloop-thread-0] RouterBuilderImpl:72 - No handlers found for operation healthy - skipping route creation
```
Since the contract contains all of the endpoints, it expects handlers for all of them for each router that gets created. This does not cause any issue functionally, but just some warning messages, making the log a little busier.

src/test/java/io/strimzi/kafka/bridge/http/MetricsIT.java
```diff
@@ -190,6 +190,7 @@ void metricsAfterConsumerReceiveRecordTest(VertxTestContext context) throws Inte
                         context.completeNow();
                     });
                 });
+        assertThat(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS), is(true));
```
see-quick, line 193: Hmm, there might have been some timing issues without this, I assume. Good catch, I wonder if we have more of these ... in suite (but it's not related to this PR ofc).

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,141 @@
...
+                }));
+    }
+
+    @Override
+    protected Map<String, Object> overrideConfig() {
+        Map<String, Object> configs = new HashMap<>();
+        configs.put(HttpConfig.HTTP_SERVER_SSL_ENABLE, true);
+        configs.put(HttpConfig.HTTP_SERVER_SSL_KEYSTORE_CERTIFICATE_CHAIN, sslCert);
+        configs.put(HttpConfig.HTTP_SERVER_SSL_KEYSTORE_KEY, sslKey);
+        // test machines do not allow non-root to bind low port such as 443 which is the default
```
see-quick:
```suggestion
        // Azure VMs don’t allow non-root users to bind to low ports, such as 443, which is the default
```

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,141 @@
...
+    @Test
+    public void testSslEnabled(VertxTestContext context) {
+        // we have to recreate the client to enable SSL
+        client = WebClient.create(vertx, new WebClientOptions()
+                .setDefaultHost(Urls.BRIDGE_HOST)
+                .setDefaultPort(Urls.BRIDGE_SSL_PORT)
+                .setSsl(true)
+                .setTrustOptions(new PemTrustOptions()
+                        .addCertValue(Buffer.buffer(sslCert)))
+                .setVerifyHost(false));
```
see-quick, line 90: Hmmm, I think you need almost in each test case `WebClient`. I would rather go with the approach of initializing here and not modifying the shared state client. For instance, JUnit doesn’t guarantee test execution order, so even though it currently works, adding a new test that relies on the shared client could break things if another test changes it.

So IMO it should be either the same for all test cases within the test suite or specific instances should just live in the scope of the test case to not influence other test cases.

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,141 @@
...
+            "JFufsVlaoqJ0PARlIqVWDRq7Umyu8zqeKLJiue1jAoGBAI3DYclNMI5r4v0j87uj\n" +
+            "Hj8X7DsZb8c/9Q89w53NUChnbrWKl5Zf78xOqcc/EezKdhsuMPrv0NfEnMtjhO24\n" +
+            "Xnw30cezds8GpHS8+UJr+6EJKxcg3dBDSYiF5Gdj46y/g/SoUKFAGlTdLgFS8gPE\n" +
+            "+6W/aNisaI3LPr2tXvgPuPQy\n" +
+            "-----END PRIVATE KEY-----";
+
+    @Test
+    public void testSslEnabled(VertxTestContext context) {
+        // we have to recreate the client to enable SSL
+        client = WebClient.create(vertx, new WebClientOptions()
```
see-quick:
```suggestion
        WebClient sslClient = WebClient.create(vertx, new WebClientOptions()
```

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,141 @@
...
+                        .addCertValue(Buffer.buffer(sslCert)))
+                .setVerifyHost(false));
+
+        baseService()
+                .getRequest("/topics/my-topic")
+                .send()
+                .onComplete(ar -> context.verify(() -> {
+                    assertThat(ar.result().statusCode(), is(HttpResponseStatus.NOT_FOUND.code()));
+                    context.completeNow();
+                }));
```
see-quick:
```suggestion
        sslClient
                .get("/topics/my-topic")
                .send()
                .onComplete(ar -> context.verify(() -> {
                    assertThat(ar.result().statusCode(), is(HttpResponseStatus.NOT_FOUND.code()));
                    sslClient.close();
                    context.completeNow();
                }));
```

documentation/modules/con-securing-http-interface.adoc
```diff
@@ -6,11 +6,17 @@
 = Securing the HTTP Bridge HTTP interface
 
 [role="_abstract"]
-Authentication and encryption between HTTP clients and the HTTP Bridge is not supported directly by the HTTP Bridge.
-Requests sent from clients to the HTTP Bridge are sent without authentication or encryption.
-Requests must use HTTP rather than HTTPS.
+By default, requests sent from clients to the HTTP Bridge are sent without encryption therefore they must use HTTP.
```
ppatierno: I would not put the sentence this way. I mean, the statement is from a client perspective saying that by default their requests are not encrypted (but how do you know? you are not in control of clients). We should put the sentence from a bridge perspective saying that by default the connection with client is not encrypted but it can be enabled. Hopefully you got what I mean :-)

documentation/modules/con-securing-http-interface.adoc
```diff
@@ -6,11 +6,17 @@
...
+You can configure a TLS-encrypted connection between HTTP clients and the HTTP Bridge through its xref:proc-configuring-kafka-bridge-{context}[properties file].
+If the HTTP Bridge is configured with TLS encryption, requests must use HTTPS.
 
-You can combine the HTTP Bridge with the following tools to secure it:
+You can enable TLS encryption for all HTTP Bridge endpoints except management endpoints such as `/ready`, `/healthy` and `/metrics`.
+These endpoints are typically used only by internal clients and requests to them must use HTTP.
+
+Authentication between HTTP clients and the HTTP Bridge is not supported directly by the HTTP Bridge.
+Requests sent from clients to the HTTP Bridge are sent without authentication.
```
ppatierno: Isn't it a duplication of the previous one?

documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -47,14 +47,77 @@ For example:
...
-http.cors.enabled=true <2>
-http.cors.allowedOrigins=https://strimzi.io <3>
-http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <4>
+http.port=443 <1>
+http.ssl.enable=true <2>
+http.ssl.keystore.certificate.location=/etc/ssl/certs/bridge.crt <3>
+http.ssl.keystore.key.location=/etc/ssl/private/bridge.key <4>
+http.cors.enabled=true <5>
+http.cors.allowedOrigins=https://strimzi.io <6>
+http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <7>
```
ppatierno: but this way you are totally replacing the example with HTTP. Maybe we should have a separate example for HTTPS so showing them both.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -147,16 +151,32 @@ private static JmxMetricsCollector createJmxMetricsCollector(BridgeConfig bridge
...
+                .requestHandler(this.managementRouter)
+                .listen()
+                .onSuccess(httpServer -> {
+                    LOGGER.info("HTTP Bridge server for management endpoints started and listening on port {}", httpServer.actualPort());
+                    this.managementServer = httpServer;
+                })
+                .onFailure(t -> {
+                    LOGGER.error("Error starting HTTP Bridge management server", t);
+                    startPromise.fail(t);
+                });
```
ppatierno: I was wondering if it's right starting up the two HTTP server asynchronously and in parallel but with both completing or failing the same promise `startPromise`. Could it be cause of a race condition? Shouldn't it be better to start them in sequence and having the second one starting up only if the first one was successful. I think that the HTTP bridge has to have them both running for being functional.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -179,9 +199,7 @@ private void startInactiveConsumerDeletionTimer(Long timeout) {
         Long timeoutInMs = timeout * 1000L;
         vertx.setPeriodic(timeoutInMs / 2, ignore -> {
             LOGGER.debug("Looking for stale consumers in {} entries", timestampMap.size());
-            Iterator<Map.Entry<ConsumerInstanceId, Long>> it = timestampMap.entrySet().iterator();
-            while (it.hasNext()) {
-                Map.Entry<ConsumerInstanceId, Long> item = it.next();
+            for (Map.Entry<ConsumerInstanceId, Long> item : timestampMap.entrySet()) {
```
ppatierno: IIRC the code was the way it was because you cannot remove an item from a collection you are going through so the usage of the iterator.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Is this a repeating warning or just a one time warning? I don't expect us to have two separate OpenAPI specifications to be honest. So maybe something to discuss/ask to the Vert.x community.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -303,15 +328,28 @@ public void stop(Promise<Void> stopPromise) {
...
+                        LOGGER.info("HTTP bridge has been shut down successfully");
+                        if (this.managementServer == null) {
+                            stopPromise.complete();
+                        }
+                    })
+                    .onFailure(t -> {
+                        LOGGER.info("Error while shutting down HTTP bridge", t);
+                        stopPromise.fail(t);
+                    });
+        }
```
ppatierno: I think the same concept about starting up the servers in sequence applies to their shutdowns.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -321,6 +359,32 @@ private HttpServerOptions httpServerOptions() {
...
+                httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
+                        .setKeyPath(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKeyLocation())
+                        .setCertPath(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreLocation()));
+            }
+
+            if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey() != null) {
+                httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
+                        .addKeyValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey()))
+                        .addCertValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain())));
+            }
```
ppatierno, line 369: should we, at least, logging some warning if the user enables SSL but doesn't complete the configuration in terms of certificates and keys?

src/test/java/io/strimzi/kafka/bridge/http/MetricsIT.java
tinaselenge: Yeah we do this in several tests to make sure async calls complete first.

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
tinaselenge: That is a good point. In my original change, I had set up a client and server for each test case instead of using the shared client. Maybe it's better to revert to that.

documentation/modules/con-securing-http-interface.adoc
tinaselenge: Yeah I understand. That's what the original sentence said exactly, I only added "by default" part to add that you can enable the encryption after it. I can improve the way it is written.

documentation/modules/con-securing-http-interface.adoc
tinaselenge: I separated the authentication and encryption since we still don't support authentication.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tinaselenge: That's a good point. We definitely want to start the management server only once the api server is started successfully. So that /ready /healthy endpoints serves requests when the main server is up and running.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tinaselenge: Sure I can add that. I had left it because we would immediately get an error saying that cert and key are required.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tinaselenge: Just one time warning at the beginning as we don't go through this part again after the deployment. You can get this warning for each endpoint once, so around 26 lines of warning messages at the beginning.

documentation/modules/con-securing-http-interface.adoc
ppatierno: But the two sentences on line 16 and 17 mention authentication only saying the same thing, or?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: OMG and one time with 26 lines of warning messages? :-o
Users will start thinking they are doing something bad. I will try to hear from my Vert.x friends about it.

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
see-quick: Yeah.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
tinaselenge: If cert or key is missing, we immediately get an error:
```
2025-11-11 12:24:08 ERROR [vert.x-eventloop-thread-0] HttpBridge:194 - Error starting HTTP Bridge server
io.vertx.core.impl.NoStackTraceThrowable: Key/certificate is mandatory for SSL
```
so I don't think adding warning message to say the same thing is necessary here.

ppatierno: Yeah I see, but it's a Vert.x specific error. I would add an error/warning which is more bridge related, explaining what's missing explaining which properties are not configured for example. But if you think the above error is enough I am fine as well.

see-quick (approved):
LGTM from testing POV. Thanks 👍

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Hey @tsegismont I would like to get your insights here if possible :-)
We have one single OpenAPI contract but with this PR we are using it for building two different routers related to two different HTTP servers. We are exposing operations via an HTTP server on port 8080, and management stuff via another HTTP server on port 8081, but still building the two routers from the same OpenAPI.
AFAICS this is the Vert.x line printing such a warning https://github.com/vert-x3/vertx-web/blob/5.0.5/vertx-web-openapi-router/src/main/java/io/vertx/ext/web/openapi/router/impl/RouterBuilderImpl.java#L149C11-L149C21 when some operations (from a contract) are not implemented.
Polluting the log with these warnings is not a great UX. Do you envisage a way to avoid it? Is having two separate OpenAPI contracts the only solution? (so that each router implements exactly all the endpoints in the related contract). Any suggestions or solutions we can't see? Thanks!

tsegismont: This warning and the corresponding error were added on purpose in https://github.com/vert-x3/vertx-web/pull/2384

If you're not interested, how about setting the log level to ERROR for this class?

Btw, I wonder if the error code shouldn't be 501, instead of 503

ppatierno: @tsegismont thanks for looking into this!

Regarding the HTTP error code I agree, so I opened a trivial PR here https://github.com/vert-x3/vertx-web/pull/2828

Regarding the logging, changing the log level was my "last resort". I was mostly wandering if there was any bad practice we are applying here. But at the same time I think that it can be a solution yes.

@tinaselenge maybe we could go with one of these two ways.
Updating the log4j2.properties file with this:

```shell
# Reduce verbosity of RouterBuilderImpl warnings for unimplemented OpenAPI endpoints
logger.routerbuilder.name = io.vertx.ext.web.openapi.router.RouterBuilderImpl
logger.routerbuilder.level = ERROR
```

(which should be also reported in the Strimzi operator repo within the default logging properties file for the bridge when we are going to use the new bridge release)

or we could go the programmatic way with something like this (to be added within the main and without any intervention or change in the operator as well):

```java
// Reduce verbosity of RouterBuilderImpl warnings for unimplemented OpenAPI endpoints
Configurator.setLevel("io.vertx.ext.web.openapi.router.RouterBuilderImpl", Level.ERROR);
```

tsegismont: I don't believe it's a bad practice, it's convenient for you and if it proves to work well, it can be considered a good practice :-)

That said, to me it seems sensible that Vert.x Web OpenAPI logs a warning when it builds router from a spec and some implementations are missing.

tinaselenge: @ppatierno @tsegismont thanks a lot for the suggestions! 

I think doing it programmatically rather changing it for the operator is a bit simpler. I will make the suggested changes.

documentation/modules/con-securing-http-interface.adoc
```diff
@@ -6,11 +6,15 @@
 = Securing the HTTP Bridge HTTP interface
 
 [role="_abstract"]
-Authentication and encryption between HTTP clients and the HTTP Bridge is not supported directly by the HTTP Bridge.
-Requests sent from clients to the HTTP Bridge are sent without authentication or encryption.
-Requests must use HTTP rather than HTTPS.
+By default, connections between HTTP clients and HTTP Bridge are not encrypted,
+but you can configure a TLS-encrypted connection through its xref:proc-configuring-kafka-bridge-{context}[properties file].
```
ppatierno: why this sentence is on a new line, I think we can keep it on the previous one.

documentation/modules/con-securing-http-interface.adoc
```diff
@@ -6,11 +6,15 @@
...
+If the HTTP Bridge is configured with TLS encryption, client requests must use HTTPS instead of HTTP.
```
ppatierno:
```suggestion
If the HTTP Bridge is configured with TLS encryption, client should connect to the bridge by using HTTPS instead of HTTP.
```

documentation/modules/con-securing-http-interface.adoc
```diff
@@ -6,11 +6,15 @@
...
 
-You can combine the HTTP Bridge with the following tools to secure it:
+You can enable TLS encryption for all HTTP Bridge endpoints except management endpoints such as `/ready`, `/healthy` and `/metrics`.
+These endpoints are typically used only by internal clients and requests to them must use HTTP.
+
+Authentication between HTTP clients and the HTTP Bridge is not supported directly by the HTTP Bridge.
+You can combine the HTTP Bridge with the following tools to secure it further:
```
ppatierno, line 16: Of course it means that the following tools become "clients" for the bridge so, if the bridge has TLS enabled, they are the ones to connect to the bridge via TLS and not the HTTP clients anymore (because they go to the proxy instead).
Should we clarify this part? No strong opinion but I would like to get your thoughts.
I would expect that if a proxy is needed between clients and HTTP bridge because of authentication or any other reasons, the user will enable TLS on the proxy and not between proxy and bridge (if they are running on the same environment).

documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -56,5 +56,82 @@ http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <4>
...
+For example:
++
+[source,properties]
+----
+bridge.id=my-bridge
+http.host=0.0.0.0
+http.port=443 <5>
+http.ssl.enable=true <6>
+http.ssl.keystore.certificate.location=/etc/ssl/certs/bridge.crt <7>
+http.ssl.keystore.key.location=/etc/ssl/private/bridge.key <8>
```
ppatierno, line 71: silly question ...
the above two "location"(s) are .ket and .crt files containing the cert/key in PEM format or location of a keystore (as the name let me understand) where the cert and key are stored? I would assume the former, because in case it was the latter, there is no property to specify the password for the keystore. Is my understanding correct? (I know it's more a Vert.x related question)

documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -56,5 +56,82 @@ http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <4>
...
+|The location of the HTTP Bridge server private key file in PEM format.
+PEM is the only format supported for private key.
+
+|http.ssl.keystore.certificate.chain
+|The HTTP Bridge server keystore certificate in PEM format.
+PEM is the only format supported for keystore.
+
+|http.ssl.keystore.key
+|The HTTP Bridge server private key in PEM format.
+PEM is the only format supported for private key.
```
ppatierno, line 112: another stupid question ... maybe related to the previous one
While with "location"(s) we have path to files containing cert/key, here we are putting the cert/key PEM content directly within the properties file, is my understanding correct? I was wondering how much this solution is secure and if we should really support it (again a more Vert.x related question).

documentation/modules/proc-configuring-http-bridge.adoc
tinaselenge: We only support PEM format so it is the former.

documentation/modules/proc-configuring-http-bridge.adoc
tinaselenge: This was discussed in the proposal and decided that we allow these configurations to provide the PEM contents directly rather than just pointing to locations. Of course, not the most secure way but can be useful for some use cases.

documentation/modules/con-securing-http-interface.adoc
tinaselenge: I tried adding a sentence to explain this. Please let me know what you think,  if needs to be more information.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -166,11 +175,18 @@ private void bindHttpServer(Promise<Void> startPromise) {
...
                     this.isReady = true;
-                    this.httpServer = httpServer;
-                    startPromise.complete();
-                })
+                    this.apiServer = apiServer;
+                    return this.vertx.createHttpServer(managementServerOptions)
+                            .connectionHandler(this::processConnection)
+                            .requestHandler(this.managementRouter)
+                            .listen()
+                            .onSuccess(httpServer -> {
```
ppatierno: use `managementServer` instead of `httpServer` as you did with `apiServer`

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -303,24 +326,65 @@ public void stop(Promise<Void> stopPromise) {
         // admin client cleanup
         this.httpBridgeContext.closeHttpAdminClientEndpoint();
 
-        if (this.httpServer != null) {
+        Future<Void> apiServerShutdown = apiServer != null ?
```
ppatierno:
```suggestion
        Future<Void> apiServerShutdown = this.apiServer != null ?
```

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -303,24 +326,65 @@ public void stop(Promise<Void> stopPromise) {
...
-                    .onSuccess(v -> {
-                        LOGGER.info("HTTP-Kafka bridge has been shut down successfully");
-                        stopPromise.complete();
-                    })
-                    .onFailure(t -> {
-                        LOGGER.info("Error while shutting down HTTP-Kafka bridge", t);
-                        stopPromise.fail(t);
-                    });
-        }
+        Future<Void> managementServerShutdown = managementServer != null ?
```
ppatierno:
```suggestion
        Future<Void> managementServerShutdown = this.managementServer != null ?
```

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,135 @@
...
+        WebClient sslClient = WebClient.create(vertx, new WebClientOptions()
+                .setDefaultHost(Urls.BRIDGE_HOST)
+                .setDefaultPort(Urls.BRIDGE_SSL_PORT)
+                .setSsl(true)
+                .setTrustOptions(new PemTrustOptions()
+                        .addCertValue(Buffer.buffer(sslCert)))
+                .setVerifyHost(false));
+
+        sslClient
+                .get("/topics/my-topic")
```
ppatierno: I would just do a GET on the root `/` (see https://github.com/strimzi/strimzi-kafka-bridge/blob/main/src/main/resources/openapi.json#L1490) and check it returns 200 OK instead of assuming you are getting a topic which doesn't exist.

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,135 @@
...
+                    assertThat(ar.result().statusCode(), is(HttpResponseStatus.NOT_FOUND.code()));
+                    context.completeNow();
+                }));
+    }
+
+    @Test
+    public void testUnencryptedConnection(VertxTestContext context) {
+        // Once SSL is enabled, HTTP Bridge server should reject unencrypted connections
+        baseService()
+                .getRequest("/topics")
```
ppatierno: Ditto as above.

src/test/java/io/strimzi/kafka/bridge/http/TlsIT.java
```diff
@@ -0,0 +1,135 @@
...
+                }));
+    }
+
+    @Override
+    protected Map<String, Object> overrideConfig() {
+        Map<String, Object> configs = new HashMap<>();
+        configs.put(HttpConfig.HTTP_SERVER_SSL_ENABLE, true);
+        configs.put(HttpConfig.HTTP_SERVER_SSL_KEYSTORE_CERTIFICATE_CHAIN, sslCert);
+        configs.put(HttpConfig.HTTP_SERVER_SSL_KEYSTORE_KEY, sslKey);
+        // Azure VMs don’t allow non-root users to bind to low ports, such as 443, which is the default
```
ppatierno: This let me think that maybe we should change the default from 443 to 8443.
These days we have been discussing about cloud providers not allowing 204 as a valid HTTP response code for health checks so we should come back to 200 somehow. I would like to avoid a similar situation, even if the workaround is clear, just override the port. But maybe a different default is better?

tinaselenge: I mean that would make it consistent with the operator's default. Strimzi operator does not allow small ports either so the default would have to be set to 8443. Does this mean we need to update the proposal? Or we are ok without changing the proposal as it's perhaps trivial enough?

ppatierno: I think we have several example of proposals that were implemented in a slightly different way. When it comes to implementation you can get differences and other issues. I don't think a proposal is needed here.

tinaselenge:
Thanks so much for the review! @ppatierno @see-quick

documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -56,5 +56,82 @@ http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <4>
...
++
+Or configure HTTP-related properties to enable TLS-encryption between clients and HTTP Bridge.
++
+For example:
++
+[source,properties]
+----
+bridge.id=my-bridge
+http.host=0.0.0.0
+http.port=443 <5>
```
katheris: I guess this is an example, but perhaps we use the default value here anyway
```suggestion
http.port=8443 <5>
```

documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -56,5 +56,82 @@ http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <4>
...
+http.ssl.enable=true <6>
+http.ssl.keystore.certificate.location=/etc/ssl/certs/bridge.crt <7>
+http.ssl.keystore.key.location=/etc/ssl/private/bridge.key <8>
+----
+<5> Configure HTTP Bridge to listen on port 443 for TLS encrypted connections.
```
katheris:
```suggestion
<5> Configure HTTP Bridge to listen on port 8443 for TLS encrypted connections.
```

documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -56,5 +56,82 @@ http.cors.allowedMethods=GET,POST,PUT,DELETE,OPTIONS,PATCH <4>
...
+
+.HTTP related configurations
+|====
+|Name |Descriptions
+
+|http.host
+|HTTP Bridge host address. The default is `0.0.0.0`.
+
+|http.port
+|HTTP Bridge port. The default is `8080`. If "http.ssl.enable" is set to `true`, then the default is `443`.
```
katheris:
```suggestion
|HTTP Bridge port. The default is `8080`. If "http.ssl.enable" is set to `true`, then the default is `8443`.
```

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -47,6 +49,9 @@ public static void main(String[] args) {
         // via https://github.com/prometheus/client_java/blob/main/prometheus-metrics-config/src/main/java/io/prometheus/metrics/config/MetricsProperties.java#L14
         System.setProperty("io.prometheus.metrics.exemplarsEnabled", "false");
 
+        // Reduce verbosity of RouterBuilderImpl warnings for unimplemented OpenAPI endpoints
+        Configurator.setLevel("io.vertx.ext.web.openapi.router.impl.RouterBuilderImpl", Level.ERROR);
```
katheris: Why isn't this being set in the default log4j2 file?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -303,24 +326,65 @@ public void stop(Promise<Void> stopPromise) {
...
-                        LOGGER.info("Error while shutting down HTTP-Kafka bridge", t);
-                        stopPromise.fail(t);
-                    });
-        }
+        Future<Void> managementServerShutdown = this.managementServer != null ?
+            this.managementServer.shutdown()
+                    .onFailure(t -> LOGGER.info("Error while shutting down HTTP bridge management server"))
+            : Future.succeededFuture();
+
+        Future.all(apiServerShutdown, managementServerShutdown)
```
katheris: Should we use `join` here instead so we at least wait for both to finish trying to shutdown before returning?

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -303,24 +326,65 @@ public void stop(Promise<Void> stopPromise) {
...
+            if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey() != null) {
+                httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
+                        .addKeyValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey()))
+                        .addCertValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain())));
+            } else {
+                if (missingSslConfigsMissing) {
+                    LOGGER.error("Required SSL configurations are missing! Either both of http.ssl.keystore.location and http.ssl.keystore.key.location " +
+                            "or both of http.ssl.keystore.certificate.chain and http.ssl.keystore.key should be configured");
+                }
+            }
```
katheris: Could this whole section be simplified to have:

```
if (..KeyStoreLocation() != null && ..KeyLocation() != null) {
    ...
} else if (..CertificateChain != null && ..Key() != null {
    ...
} else {
    LOGGER.error(...);
}
```
```suggestion
            if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreLocation() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKeyLocation() != null) {
                httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
                        .setKeyPath(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKeyLocation())
                        .setCertPath(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreLocation()));
            } else if (bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain() != null && this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey() != null) {
                httpServerOptions.setKeyCertOptions(new PemKeyCertOptions()
                        .addKeyValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreKey()))
                        .addCertValue(Buffer.buffer(this.bridgeConfig.getHttpConfig().getHttpServerSslKeystoreCertificateChain())));
            } else {
                LOGGER.error("Required SSL configurations are missing! Either both of http.ssl.keystore.location and http.ssl.keystore.key.location " +
                            "or both of http.ssl.keystore.certificate.chain and http.ssl.keystore.key should be configured");
            }
```

katheris (commented):
Thanks @tinaselenge, generally looks good but I suggested a few changes

src/main/java/io/strimzi/kafka/bridge/Application.java
tinaselenge: that's a good point!

config/log4j2.properties
```diff
@@ -24,4 +24,8 @@ logger.bridge.additivity = false
 logger.healthy.name = http.openapi.operation.healthy
 logger.healthy.level = WARN
 logger.ready.name = http.openapi.operation.ready
-logger.ready.level = WARN
\ No newline at end of file
+logger.ready.level = WARN
+
+# Reduce verbosity of RouterBuilderImpl warnings for unimplemented OpenAPI endpoints because bridge servers share the same OpenAPI contract with different endpoints implemented.
+logger.vertx.name = io.vertx.ext.web.openapi.router.impl.RouterBuilderImpl
+logger.vertx.level = ERROR
```
ppatierno, line 31: @tinaselenge didn't we decide to go the programmatic way also to avoid copying the same change to the operator default config for the bridge? It was discussed here https://github.com/strimzi/strimzi-kafka-bridge/pull/1040#discussion_r2495577391 FYI @katheris why do you think it's better this way?

katheris: @ppatierno sorry, I didn't see that discussion, hence my question of why we were doing it programmatically. I was thinking from a user perspective, if the user tried to change the level of that class logging wouldn't the programmatic way then override the user defined one? Also I guess it makes it harder for users to understand the default levels if they are separated between programmatic and the lo4j2 file?

tinaselenge: Sorry, I should have linked the discussion in the comment first and explain why I decided to change it. I initially thought it's simpler to set it programatically without having to change anything on the operator side but after the comment from @katheris, I thought about it again and from user's perspective and maintenance point of view, it might be clearer to set it vialog4j2 file (even though it is less likely that users would change logging level for Vertx classes). @ppatierno as you suggested both solutions, which one do you lean towards more?

ppatierno: @tinaselenge @katheris my take on this was that I can't see why a user should change the log level of that specific class. If you look at `RouterBuilderImpl` from Vert.x you can see that it's logging only on a specific like here https://github.com/vert-x3/vertx-web/blob/5.0.5/vertx-web-openapi-router/src/main/java/io/vertx/ext/web/openapi/router/impl/RouterBuilderImpl.java#L149 the warning we were talking about but nothing more. I saw it similar to how we disable Prometheus exemplars and OTEL support for metrics and logs which is done programmatically. But if you think that leaving the user the visibility of the such logging configuration, I can see your point.

tinaselenge: Those are good points. I also thought this made sense because this is a low level Vertx class that users wouldn't normally change logging anyway. But I haven't seen in our projects programatically changing configuring logging level. From maintenance point of view, I thought it might be more consistent to have it in log4j2 file. However, if this is a common practice in general to do it programmatically for it, maybe it makes more sense to revert the change since this class doesn't have many other loggings.

tinaselenge: @katheris do you have any more thoughts on this? I think both ways make sense so I don't mind either one at this point. Doing it via log4j2 file is better from maintainability point of view, as it's more visible than being in the code. Doing it programatically is simpler as users likely won't change the logging for this specific class and this class currently has only the warning logging that we want to silence (but I guess more logging could get added in the future).

katheris: I would probably lean slightly more towards having it in the log4j2 file just to remove any confusion both for users and Strimzi contributors, however I'm happy to go with @ppatierno's preference as the component owner

ppatierno: Let's leave the log4j implementation for now and see how it goes ;-)

ppatierno (approved):
LGTM. Thanks Tina!

PR #1047, opened 2025-11-19
Change success response status of health checks to 200 (OK)

joschi:
Some load balancers cannot handle HTTP response codes other than 200 (OK) in health checks and treat every other response codes (even in the 2xx range) as unhealthy.

The response codes of the `/ready` and `/healthy` HTTP resources has been changed from 204 (No Content) to 200 (OK) to accommodate these load balancer health checks.


ppatierno:
@joschi meanwhile please check that doc build is failing with the following error (in the Azure pipeline):

```shell
M	documentation/book/api/.openapi-generator/openapi.json-generate-apidoc.sha256
M	documentation/book/api/index.adoc
  make docu_api \
make: *** [Makefile:69: docu_check] Error 1
```

you should be able to fix it by just following the above instructions.

see-quick:
> Thanks for the PR! I was wondering if this type of change would need a Strimzi proposal first in order to explain the motivation for the change (which is anyway in the description here) and the compatibility with current users. It would not be that complex and in the past we had similar proposals like https://github.com/strimzi/proposals/blob/main/063-pdb-generation-environment-variable.md and https://github.com/strimzi/proposals/blob/main/028-network-policy-generation-environment-variable.md to use an env var for changing the default behaviour within the Strimzi operator. @strimzi/maintainers wdyt?

SGTM 👍  having the proposal to set motivation and make it more visible (to current users).

joschi:
@ppatierno @see-quick changed the PR according to the latest proposal.

scholzj:
@ppatierno I guess we both missed the CHANGELOG update :-/

Issue #1037, opened 2025-10-23
Run Integration tests with multiple Kafka versions supported by test container


see-quick:
Currently, when we run IT, we test only the latest version, and IMO it would be useful to test it with multiple Kafka versions, which are supported by our strimzi test container (currently 4.0.0, 4.0.1 and 4.1.0).

katheris:
Triaged on 18.12.2025: After discussion (see recording for more details) we're happy to add this as although the system tests in the cluster operator test across different versions the integration tests are more thorough.

This parametrization should be done at the unit test level, rather than through test infrastructure settings.

PR #1053, opened 2025-12-18
Make HttpBridgeITAbstract parametrized to run tests against multiple Kafka versions

Frawless:
This PR bump several dependencies to allow us use `@ParametrizedClass` from junit6 to run our ITs against multiple Kafka versions.

Closes https://github.com/strimzi/strimzi-kafka-bridge/issues/1037

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
@@ -35,13 +35,16 @@ import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.TestInstance;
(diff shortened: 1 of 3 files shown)


src/test/java/io/strimzi/kafka/bridge/http/base/HttpBridgeITAbstract.java
```diff
@@ -82,6 +85,9 @@
 @SuppressWarnings({"checkstyle:JavaNCSS"})
 @Tag(HTTP_BRIDGE)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ParameterizedClass
+// These are Kafka versions supported by Strimzi Test Container. Should be updated together with test-container version bump.
+@ValueSource(strings = {"4.0.0", "4.0.1", "4.1.0", "4.1.1"})
```
see-quick, line 90: Sorry if it was already discussed...I wonder if it would be easier to have a method inside the test container (something like `getSupportedKafkaVersions()`). This way, we would only bump the version of TC and avoid modifying this part as well...

Frawless: Yes, @ppatierno pointed on it on slack, afaiu it is not possible now and it will required new release. I can open issue to track it. It will be great to have it, but I don't think it should block this PR given that new release of TC will be needed.

see-quick: Sure, opening an issue would be great 👍 . Thanks

Frawless: I opened https://github.com/strimzi/test-container/issues/177

PR #1063, opened 2026-01-21
Add GitHub Actions

Frawless:
This PR adds GitHub actions for build, CVE-release, and release. As part of the PR the build jobs from Azure are commented out and GHA will be used for all the builds.

Build: https://github.com/Frawless/strimzi-kafka-bridge/actions/runs/21202247982
Release: https://github.com/Frawless/strimzi-kafka-bridge/actions/runs/21202265364
CVE-release: https://github.com/Frawless/strimzi-kafka-bridge/actions/runs/21202555593

### DIFF
--- a/.github/actions/build/build-binaries/action.yml
+++ b/.github/actions/build/build-binaries/action.yml
@@ -10,6 +10,10 @@ inputs:
     description: "Architecture of GitHub runner"
     required: false
     default: "amd64"
+  mainBuild:
+    description: "Whether this is the main build"
+    required: false
+    default: "false"
 
 runs:
   using: "composite"
@@ -47,15 +51,24 @@ runs:
       run: make java_verify
       env:
         MVN_ARGS: "-e -V -B -Dsurefire.rerunFailingTestsCount=5 -Dfailsafe.rerunFailingTestsCount=2"
+        TESTCONTAINERS_RYUK_DISABLED: "TRUE"
+        TESTCONTAINERS_CHECKS_DISABLE: "TRUE"
 
     - name: Create tarball with target directory
       shell: bash
       run: tar -cvpf bridge-binaries.tar ./target
 
-    - name: Upload binaries artifact
+    - name: Upload binaries artifact (main build)
+      if: ${{ inputs.mainBuild == 'true' }}
       uses: actions/upload-artifact@v5
       with:
         name: bridge-binaries.tar
+        path: bridge-binaries.tar
+
+    - name: Upload binaries artifact with Java version
+      uses: actions/upload-artifact@v5
+      with:
+        name: bridge-binaries-java-${{ inputs.javaVersion }}.tar
         path: bridge-binaries.tar
 
     - name: Clean external SNAPSHOT dependencies from Maven repository
@@ -67,7 +80,7 @@ runs:
         -DreResolve=false || true
 
     - name: Save Maven cache
-      if: github.ref == 'refs/heads/main'
+      if: github.ref == 'refs/heads/main' && inputs.mainBuild == 'true'
       uses: actions/cache/save@v5
       with:
         path: ~/.m2/repository
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,14 +22,18 @@ jobs:
     name: Build Kafka Bridge
     strategy:
       matrix:
-        javaVersion:
-          - "17"
+        # Java 17 is used as default version
+        # In case of adding new version make sure that build-containers and deploy-java has correct static version set
+        java:
+          - version: "17"
+            mainBuild: true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/build/build-binaries
         with:
-          javaVersion: ${{ matrix.javaVersion }}
+          javaVersion: ${{ matrix.java.version }}
+          mainBuild: ${{ matrix.java.mainBuild }}
           runnerArch: "amd64"
 
   build-docs:
@@ -112,6 +116,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/build/deploy-java
         with:
+          # Java 17 is used as default version
           javaVersion: "17"
           runnerArch: "amd64"
           gpgPassphrase: ${{ secrets.GPG_PASSPHRASE }}
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,6 +20,11 @@ on:
   schedule:
     - cron: '23 17 * * 3'
 
+# Declare default permissions
+permissions:
+  contents: read
+  security-events: write
+
 jobs:
   analyze:
     name: Analyze


.github/actions/build/build-containers/action.yml
```diff
@@ -0,0 +1,68 @@
...
+  buildRunId:
+    description: "Build workflow run ID for artifact download"
+    required: false
+    default: ""
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Docker
+      uses: strimzi/strimzi-kafka-operator/.github/actions/dependencies/install-docker@main
```
ppatierno, line 22: isn't this wrong? it's referring to a non existing path, or?
or you are using the path to the Strimzi operator repo to use common stuff. Are we doing the same for other components? I don't think it was a practice before with AZP. Please can you elaborate more, it's me and my lack of knowledge here.

.github/actions/build/build-docs/action.yml
```diff
@@ -0,0 +1,45 @@
+name: "Build Documentation"
+description: "Builds Strimzi documentation in HTML and PDF formats"
```
ppatierno: it's the "Strimzi bridge documentation"

.github/actions/build/deploy-java/action.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Deploy Java artifacts
+      shell: bash
+      run: ./.azure/scripts/push-to-central.sh
+      env:
+        BUILD_REASON: "IndividualCI"
+        BRANCH: ${{ github.ref }}
+        GPG_PASSPHRASE: ${{ inputs.gpgPassphrase }}
+        GPG_SIGNING_KEY: ${{ inputs.gpgSigningKey }}
+        CENTRAL_USERNAME: ${{ inputs.centralUsername }}
+        CENTRAL_PASSWORD: ${{ inputs.centralPassword }}
```
ppatierno, line 51: out of curiosity, which username and password is now using? IIRC we set up a user token for that to be used within the pipelines, or is it your personal one now?

.github/actions/build/push-containers/action.yml
```diff
@@ -0,0 +1,143 @@
+name: "Push Access-Operator Containers"
```
ppatierno: I am really confused now :-D

.github/actions/build/release-artifacts/action.yml
```diff
@@ -0,0 +1,74 @@
+name: "Build Release Artifacts"
+description: "Builds release artifacts (tar.gz, zip, Helm chart)"
```
ppatierno: Bridge doesn't have Helm chart.

.github/workflows/cve-rebuild.yml
```diff
@@ -0,0 +1,132 @@
...
+  pull-java-artifacts-from-ghcr:
+    name: Pull Java artifacts from GHCR
+    if: ${{ startsWith(github.ref, 'refs/heads/release-') }}
+    runs-on: ubuntu-latest
+    permissions:
+      packages: read
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup ORAS
```
ppatierno, line 34: what's ORAS?

development-docs/RELEASING.md
```diff
@@ -23,38 +23,39 @@ This is implemented in the `Makefiles`, you just need to run the command `RELEAS
...
-After the release pipeline is finished, the release has to be created:
+After the release workflow is finished, the release has to be created:
 
 * Tag the right commit from the release branch with the release name (e.g. `git tag 1.2.0`) and push it to GitHub
-* Go to **Maven Central** > **Publish** to publish the release artifacts
-* Update the documentation on the website using the HTML files from the build pipeline artifacts 
-* On GitHub, create the release and attach the ZIP / TAR.GZ artifacts from the build pipeline artifacts
+* The Java artifacts are automatically pushed to Maven Central by the release workflow.
+Once this is done, go to the Sonatype portal and publish the artifacts.
+* The documentation is automatically published to the website by the build workflow
```
ppatierno: How is it going to happen? Does it open a PR? I mean, it's wonderful! But I would like to know more about how it works now :-D

.github/actions/build/build-containers/action.yml
Frawless: GHA allow users to re-use existing actions from another repositories. That's the case as well. Once I will finish first part of the migration I wan tot create new repo in Strimzi and move them these common actions we use across all repos. It will help us with usage and versioning.

.github/actions/build/build-docs/action.yml
Frawless: Copy/paste problem, I will fix it

.github/actions/build/deploy-java/action.yml
Frawless: TBH I have no idea. I just copy/paste stuff from Azure Pipelines. @im-konge added the variables into previous repositories so maybe he knows well. I guess the tooling doesn't care about username if token is used, but there has to be something?

.github/actions/build/push-containers/action.yml
Frawless: Copy/paste problem, I will fix it

.github/actions/build/release-artifacts/action.yml
Frawless: Right, again copy/paste issue,  I will fix it :)

.github/workflows/cve-rebuild.yml
Frawless: It's OCI image client. I found it very easy to use when manipulating with GHCR where we store build artifacts for CVE respins. GHA doesn't allow users to mark builds as permanent so this is proper way how to handle such situations - https://github.com/oras-project/oras

development-docs/RELEASING.md
Frawless: Good question, I think it is just halucination from my AI apprentice :D So `* The documentation is automatically published to the website by the build workflow` this is basically not true and can be removed or changed to something like `The documentation is stored as an artifact of the build` ? I don't really know how this works so feel free to suggest a correction or I can just drop the line, up to you.

.github/actions/build/deploy-java/action.yml
Frawless: BTW I did not test publishing into Sonatype, I am not sure if it's even possible to easily test it. However, it works fine for operators repo so unless there is a different process I would expect it will work here as well. All other parts were tested in my fork - see links in PR description.

.github/actions/build/build-containers/action.yml
ppatierno: If that's the final ok, then fine with me with this temporary solution. Who knows if we are going to change something in the strimzi-kafka-operator, by removing some actions and suddenly breaking other repos. A common repo would make sense. Thanks.

development-docs/RELEASING.md
ppatierno: I would just leave the previous sentence "Update the documentation on the website using the HTML files from the build pipeline artifacts" because this is what we actually do today. Download the documentation artifacts, unzip and move some content across the website repo and open a PR there.

Frawless: Alright, I will update it

.github/actions/build/deploy-java/action.yml
ppatierno: Ok. If copied from Azure than it should be a token with password I created for the bridge. We'll know on the next release if everything works fine or not :-)

Frawless: Yes, unfortunately this is not easy to test properly :/ But it is working for operators so I hope it will work here as well (assuming correct creds are setup for GHA)

PR #1067, opened 2026-02-01
Moved Javadoc validation to checkstyle

### LINKED ISSUE #1058: Move Javadoc validation to checkstyles and update Javadocs to be in-sync with the operators repo
Sooner or later the Bridge will have to move to Java 21 or newer not only as a container runtime, but also as build and language version. However, the Bridge Javdocs are currently not valid for Java 21. I think we should do something similar as we did in operators in https://github.com/strimzi/strimzi-kafka-operator/pull/10527:
* Disable Javadoc failures on warnings in the Java Javadoc plugin
* Move Javadoc validation to Checkstyle instead
* Use the same Javadoc rules in the Bridge as we do in operators (+ fix the Javadocs accordingly)

ppatierno:
This PR fixes #1058 (as preparation for Java 21) by:

* disabling the Javadoc failures on warnings in the maven Javadoc plugin
* moving Javadoc validation to checkstyle
* using the same Javadoc rules as the Strimzi operator

It adds Java 21 to the CI pipelines for validation but still using Java 17 for everything else (binaries push and the release process).

### DIFF
--- a/.github/actions/build/build-binaries/action.yml
+++ b/.github/actions/build/build-binaries/action.yml
@@ -65,12 +65,6 @@ runs:
         name: bridge-binaries.tar
         path: bridge-binaries.tar
 
-    - name: Upload binaries artifact with Java version
-      uses: actions/upload-artifact@v5
-      with:
-        name: bridge-binaries-java-${{ inputs.javaVersion }}.tar
-        path: bridge-binaries.tar
-
     - name: Clean external SNAPSHOT dependencies from Maven repository
       if: github.ref == 'refs/heads/main'
       shell: bash
--- a/.azure/templates/jobs/build_java.yaml
+++ b/.azure/templates/jobs/build_java.yaml
@@ -8,6 +8,10 @@ jobs:
           image: 'Ubuntu-22.04'
           jdk_version: '17'
           main_build: 'true'
+        'java-21':
+          image: 'Ubuntu-22.04'
+          jdk_version: '21'
+          main_build: 'false'
     # Set timeout for jobs
     timeoutInMinutes: 60
     # Base system
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,8 @@ jobs:
         java:
           - version: "17"
             mainBuild: true
+          - version: "21"
+            mainBuild: false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6


.github/actions/build/build-binaries/action.yml
```diff
@@ -65,12 +65,6 @@ runs:
         name: bridge-binaries.tar
         path: bridge-binaries.tar
 
-    - name: Upload binaries artifact with Java version
-      uses: actions/upload-artifact@v5
-      with:
-        name: bridge-binaries-java-${{ inputs.javaVersion }}.tar
-        path: bridge-binaries.tar
-
```
scholzj, line 73: Why are you removing this? It seems wrong as that will be needed for the image build.

Frawless: There are two archive targets - one with name `bridge-binaries.tar` and one with `bridge-binaries-java-${{ inputs.javaVersion }}.tar`. The reason I put there the second one was to store binaries for all versions in case someone will want to download it locally and check the functionality. For building images in CI we need only `bridge-binaries.tar` what is archive as part of Java `mainBuild` which is done few lines before these. This change will not break anything, the question is if we want to archive it for all Java versions or not.

ppatierno: I removed this because IIRC we didn't have something like this in the past on Azure with several binaries based on different Java versions. In the past, we had both Java 11 and Java 17 to compile but we produced and distributed binaries only for Java 11 as the supported version (the Java API source level was at 11). Of course, Java 11 binaries worked fine on Java 17 runtime as within the container images. I made the same assumption here (we had 11 and 17, now 17 and 21). The bridge will still be built as Java 17 binaries but we want a build (with no binaries) based on Java 21 for validation on code or dic changes to catch broken stuff earlier (before the final migration to Java 21).

scholzj: Ok, if it does not break anything, I guess I'll leave it to you to decide.

scholzj (approved):

.github/actions/build/build-binaries/action.yml
ppatierno: So @Frawless are you ok with that?

Frawless: Sure, I am fine with that. I will keep it in mind and will remove it also from other repos to keep consistency.

Frawless (approved):

PR #1070, opened 2026-02-11
Improve naming of the SSL options

tinaselenge:
Based on the recent discussion, naming for some of the SSL properties were confusing as they mix "keystore" terminology with PEM certificate/key terminology. This PR attempts to make them clearer, with more consistent terminology by making the current options only PEM specific. 

In the future, when we add support for mTLS, we could add the following options:
- http.ssl.trusted.certificates.location
- http.ssl.trusted.certificates

Also in the future, if we ever need to support more formats, we could add:
- http.ssl.keystore.location
- http.ssl.keystore.password
- http.ssl.keystore.type ( PKCS12 | JKS )
- http.ssl.truststore.location
- http.ssl.truststore.type ( PKCS12 | JKS )
- http.ssl.truststore.password

The separation of options for PEM and other formats help not mix the terminologies and provide clearer definitions.

### DIFF
--- a/documentation/modules/proc-configuring-http-bridge.adoc
+++ b/documentation/modules/proc-configuring-http-bridge.adoc
@@ -67,13 +67,13 @@ bridge.id=my-bridge
 http.host=0.0.0.0
 http.port=8443 <5>
 http.ssl.enable=true <6>
-http.ssl.certificate.path=/etc/ssl/certs/bridge.crt <7>
-http.ssl.key.path=/etc/ssl/private/bridge.key <8>
+http.ssl.certificate.location=/etc/ssl/certs/bridge.crt <7>
+http.ssl.key.location=/etc/ssl/private/bridge.key <8>
 ----
 <5> Configure HTTP Bridge to listen on port 8443 for TLS encrypted connections.
 <6> Set `true` to enable TLS encryption between HTTP clients and the HTTP Bridge.
-<7> Path to the certificate file in PEM format.
-<8> Path to the private key file in PEM format.
+<7> Location of the certificate file in PEM format.
+<8> Location of the private key file in PEM format.
 
 . Save the configuration file.
 
@@ -92,15 +92,15 @@ http.ssl.key.path=/etc/ssl/private/bridge.key <8>
 
 |http.ssl.enable
 |Enable TLS encryption between HTTP clients and HTTP bridge. It is `false` by default.
-If set to `true`, `http.ssl.certificate.path` and `http.ssl.key.path` configurations or
+If set to `true`, `http.ssl.certificate.location` and `http.ssl.key.location` configurations or
  `http.ssl.certificate` and `http.ssl.key` configurations are required.
 
-|http.ssl.certificate.path
-|The path to the HTTP Bridge server certificate file in PEM format.
+|http.ssl.certificate.location
+|The location of the HTTP Bridge server certificate file in PEM format.
 PEM is the only format supported for certificate.
 
-|http.ssl.key.path
-|The path to the HTTP Bridge server private key file in PEM format.
+|http.ssl.key.location
+|The location of the HTTP Bridge server private key file in PEM format.
 PEM is the only format supported for private key.
 
 |http.ssl.certificate
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -355,16 +355,16 @@ public class HttpBridge extends AbstractVerticle {
         if (this.bridgeConfig.getHttpConfig().isSslEnabled()) {
             httpServerOptions.setSsl(true);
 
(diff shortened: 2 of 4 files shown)


documentation/modules/proc-configuring-http-bridge.adoc
```diff
@@ -92,22 +92,22 @@ http.ssl.keystore.key.location=/etc/ssl/private/bridge.key <8>
...
-|http.ssl.keystore.key.location
-|The location of the HTTP Bridge server private key file in PEM format.
+|http.ssl.key.path
+|The path to the HTTP Bridge server private key file in PEM format.
 PEM is the only format supported for private key.
 
-|http.ssl.keystore.certificate.chain
-|The HTTP Bridge server keystore certificate in PEM format.
-PEM is the only format supported for keystore.
+|http.ssl.certificate
```
scholzj, line 106: So, if we call this `http.ssl.certificate` / `http.ssl.certificate.path` ... how would we call the _trusted certificates_ if we one day add support for mTLS authentication?

katheris:
I think the confusion here is that the settings are similar to the Kafka settings, but actually the way the user provides them is different from Kafka. So for example in Kafka you can provide these configurations for files:
`ssl.keystore.type` (PEM or JKS or PKCS12)
`ssl.keystore.location`
`ssl.keystore.password`
`ssl.key.password`
`ssl.truststore.type` (PEM or JKS or PKCS12)
`ssl.truststore.location`
`ssl.truststore.password`

Where `ssl.keystore.location` contains both the private key and the public certificate. Whereas for Bridge we have the key and certificate as separate files:
`http.ssl.keystore.certificate.location`
`http.ssl.keystore.key.location`

So given the starting point I'm not sure what can be changed to make it less confusing

scholzj:
@katheris I can live with 1 and 2.

I think the option 3 - while I know it aligns with Kafka - goes against how it is usually used. So I would avoid that. (I kind of understand why Kafka opted for it as they have these options first from using JKS/PKCS12 - but we do not have this history to follow on).

The `keystore` and `truststore` names in the options are slightly annoying as these are not really keystores and trusstores at this point. But they might be if we one day add support for PKCS12 or JKS. So I can learn to live with it. 🤷 

_Please also keep in mind that this follows on the discussion in https://github.com/strimzi/proposals/pull/201#discussion_r2786399444 - these are slightly related and while we do not have to use the same in both places, it might be worth considering it in the whole context._

scholzj:
@ppatierno Keystore/truststore are not the same as files. They refer to a special storage format designed to store cryptographic information in a secure way. So, not every file is a key or truststore (No, the `PEM` _keystore type_ in Kafka is not a keystore). And not every keystore/trusstore has to be a file (you can have hardware or network-based stores etc.).

> They follow exactly the same approach as Kafka.
> With FILE (store):
> 
> http.ssl.keystore.location --> does it contain only certs chain?
> http.ssl.keystore.key.location --> does it contain only the key?

While this brings it on par with how the PEM files are usually used (a separate public and private keys), I think this has a pretty bad usability given the confusing names which make it completely unclear what is the private key and what is the public key.

Also, this does not follow the same approach as Kafka as it is different 😉.

> Furthermore, what I see missing in the bridge, is a password for when the keystore is in a FILE (so something like in Kafka, http.ssl.keystore.password) and also a password if the key is encrypted (so something like in Kafka, http.ssl.key.password). Not sure if it was on purpose or not. Maybe not discussed in the proposal.

It is not missing. It has no value because we do not support any keystore. So there are no passwords in game. That would come only if we one day decide to support any keystores. But there is no such support in the implementation or in the proposal as far as I know.

scholzj:
> @ppatierno, from the way I'm reading them seem to suggest that the inline properties are just for PEM with the default implementation: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/ssl/DefaultSslEngineFactory.java#L280 but I could be understanding this wrong. Anyways, this would only matter if we ever need to support other formats :)

That is my read as well ... as in `anything can happen if you use custom SSL implementations, but by default it is only unencrypted PEMs`.

ppatierno:
> @ppatierno, from the way I'm reading them seem to suggest that the inline properties are just for PEM with the default implementation: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/security/ssl/DefaultSslEngineFactory.java#L280
but I could be understanding this wrong. Anyways, this would only matter if we ever need to support other formats :)

@tinaselenge you are right. I misread the doc and the code is much clearer :-)

> For a store such as PKCS12 or JKS, this is normal, yes. But for a PEM file, it is IMO not normal. The inline PEM options are separate in Kafka, which is expected and correct from my perspective.

@scholzj Not sure about the "it is IMO not normal" part. Is it your own opinion? I would assume so given the "IMO" part :-) Or do you have any source for that?
From my knowledge, it's accepted to put different "objects" within a PEM file and you can mix certs with key. They are all well separated by headers. If that's good or not, compared to use PKCS12 or JKS, it's a different story and it could have disadvantages so maybe it's not much used.

Said that, I think we can move forward with Tina suggestion.

PR #1074, opened 2026-02-16
Refactor integration tests - part 1 - changes to installation, refactor AdminIT, HttpCorsIT

im-konge:


src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,137 @@
...
+import io.strimzi.kafka.bridge.config.KafkaProducerConfig;
+import io.strimzi.kafka.bridge.http.HttpConfig;
+import io.strimzi.kafka.bridge.httpclient.HttpRequestHandler;
+import io.strimzi.kafka.bridge.metrics.MetricsType;
+import io.strimzi.test.container.StrimziKafkaCluster;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.*;
```
see-quick: replace this with specific imports?

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final Logger LOGGER = LogManager.getLogger(AbstractIT.class);
+    private static final String DEFAULT_OPENJDK_IMAGE = "registry.access.redhat.com/ubi9/openjdk-21-runtime:latest";
+
+    public static ObjectMapper objectMapper = new ObjectMapper();
+
+    public static HttpRequestHandler httpRequestHandler = null;
+    public static StrimziKafkaCluster kafkaCluster = null;
+    public static GenericContainer<?> bridge = null;
+
+    public static String topicName = null;
```
see-quick: You don't need to init those with null (that's implicit)

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,137 @@
...
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Random;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class AbstractIT {
```
see-quick: I think you should add that `@ParametrizedClass`, which we have to test multiple versions?

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,137 @@
...
+    public static HttpRequestHandler httpRequestHandler = null;
+    public static StrimziKafkaCluster kafkaCluster = null;
+    public static GenericContainer<?> bridge = null;
+
+    public static String topicName = null;
+    private final static Random RNG = new Random();
+
+    @BeforeAll
+    void beforeAll() {
+        setupKafkaCluster(null);
```
see-quick: I think we don't need that or? 
```java
StrimziKafkaCluster.StrimziKafkaClusterBuilder builder = new StrimziKafkaCluster.StrimziKafkaClusterBuilder()
            .withKafkaVersion(kafkaVersion)
            .withNumberOfBrokers(1)
            .withSharedNetwork();
```
as it would be like this? Or you want to make sure that default behavior of running ITs would be just with latest Kafka version? And then on CIs we would have multiple Kafka versions defined?

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,137 @@
...
+    public void setupBridge() throws IOException {
+        String propertiesPath = createPropertiesFile(getConfiguration());
+
+        GenericContainer<?> container = new GenericContainer<>(DEFAULT_OPENJDK_IMAGE)
+            .withFileSystemBind(getBridgeJarPath(), "/app/", BindMode.READ_ONLY)
+            .withFileSystemBind(propertiesPath, "/app/application.properties", BindMode.READ_ONLY)
+            .withCommand("/app/bin/kafka_bridge_run.sh", "--config-file=/app/application.properties")
+            .withExposedPorts(8080)
+            .withNetwork(Network.SHARED)
+            .waitingFor(Wait.forHttp("/"));
```
see-quick: I think better is to wait for `/ready` ?

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
im-konge: Omg, thanks 🤦

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
im-konge: I firstly started with this and then I want add this, yes.

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
im-konge: For now, it's like this, as I mentioned in the previous comment.
But I would like to do this:

- for local run use just latest Kafka version
- in CI run multiple the tests with multiple Kafka versions

To not make things confusing, I will add it in this PR.

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
im-konge: yeah, originally I had `/healthy` but it doesn't exist here :) `/ready` will do the trick. Thanks

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,8 @@
+package io.strimzi.kafka.bridge.http.base;
```
see-quick, line 5: header?

src/test/java/io/strimzi/kafka/bridge/http/extensions/configuration/BridgeConfiguration.java
```diff
@@ -0,0 +1,25 @@
+package io.strimzi.kafka.bridge.http.extensions.configuration;
```
see-quick: also here

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
see-quick: didn't checkstyle complain?

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
```diff
@@ -0,0 +1,8 @@
...
+
+import org.junit.jupiter.params.Parameter;
+
+public abstract class AbstractIT {
+    @Parameter
+    String kafkaVersion;
+}
```
Frawless, line 17: A comment why do we need this would be good

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface BridgeConfiguration {
+    String[] properties() default {
+        KafkaConsumerConfig.KAFKA_CONSUMER_CONFIG_PREFIX + ConsumerConfig.AUTO_OFFSET_RESET_CONFIG + "=earliest",
+        KafkaProducerConfig.KAFKA_PRODUCER_CONFIG_PREFIX + ProducerConfig.MAX_BLOCK_MS_CONFIG + "=10000",
+        HttpConfig.HTTP_CONSUMER_TIMEOUT + "=5",
+        BridgeConfig.METRICS_TYPE + "=strimziMetricsReporter",
+        BridgeConfig.BRIDGE_ID + "=my-bridge"
```
Frawless: I assume these are some defaults values and eventually can be overwritten? Comment would be great.

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeExtension.java
```diff
@@ -0,0 +1,190 @@
...
+            setupBridge(extensionContext);
+        }
+    }
+
+    @Override
+    public void afterEach(ExtensionContext extensionContext) {
+        if (!isPerClass(extensionContext)) {
+            stopBridge(extensionContext);
+        }
+    }
```
Frawless: what if we will have test class that won't be parametrized?

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeTest.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ExtendWith({KafkaExtension.class, BridgeExtension.class, TestStorageExtension.class})
+@ParameterizedClass
+@MethodSource("io.strimzi.kafka.bridge.http.extensions.KafkaExtension#kafkaVersions")
```
Frawless: Maybe some comment in readme or somewhere that BridgeTest loads kafka versions and this should be changed in case of changes in version loading, etc?

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeTest.java
```diff
@@ -0,0 +1,22 @@
...
+@BridgeConfiguration
+public @interface BridgeTest {
+}
```
Frawless: Having `BridgeTest` and `Test` after is strange, I would rename it to `Bridge` or `BridgeTestEnvironment` for example.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.ParameterInfo;
+
+import java.util.List;
+
+public class KafkaExtension implements
+    BeforeClassTemplateInvocationCallback,
+    AfterClassTemplateInvocationCallback {
+    private static final Logger LOGGER = LogManager.getLogger(KafkaExtension.class);
+    private static final Boolean RUN_WITH_ALL_KAFKA_VERSIONS_ENV = Boolean.valueOf(System.getenv().getOrDefault("RUN_WITH_ALL_KAFKA_VERSIONS", "false"));
```
Frawless: Default should be true, only if specified run a specific Kafka version. What if user specify Kafka version that is not supported by test-container?

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,80 @@
...
+    private void setupKafkaCluster(ExtensionContext extensionContext, String kafkaVersion) {
+        StrimziKafkaCluster.StrimziKafkaClusterBuilder kafkaBuilder = new StrimziKafkaCluster.StrimziKafkaClusterBuilder()
+            .withNumberOfBrokers(1)
+            .withSharedNetwork();
+
+        if (kafkaVersion != null && !kafkaVersion.isEmpty() && !kafkaVersion.equals("latest")) {
+            LOGGER.info("Using Kafka version: {}", kafkaVersion);
+            kafkaBuilder.withKafkaVersion(kafkaVersion);
+        } else {
+            LOGGER.info("Using default Kafka version from test container");
```
Frawless: I would use `latest` everywhere or `default` everywhere

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private String topicName;
+    private HttpRequestHandler httpRequestHandler;
+    private AdminClientFacade adminClientFacade;
+
+    public TestStorage(ExtensionContext extensionContext) {
+        topicName = "my-topic-" + RNG.nextInt(Integer.MAX_VALUE);
+        httpRequestHandler = BridgeExtension.getHttpRequestHandler(extensionContext);
+        adminClientFacade = AdminClientFacade.create(KafkaExtension.getKafkaCluster(extensionContext).getBootstrapServers());
+    }
```
Frawless: what if test will use multiple topics?

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeTestContextExtension.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class TestStorageExtension implements BeforeEachCallback, AfterEachCallback, ParameterResolver {
+
+    private static final String TEST_STORAGE_KEY = "testStorage";
+
+    @Override
+    public void beforeEach(ExtensionContext extensionContext) {
+        TestStorage testStorage = new TestStorage(extensionContext);
+        getStore(extensionContext).put(TEST_STORAGE_KEY, testStorage);
+    }
```
Frawless, line 30: In case there will be class-wide required configs, it can be extends in the future right?

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
im-konge: What do you mean? If test will use multiple topics, they will be created in the test directly and you will not have to use the `TestStorage`. That's the same thing as in operators repo. This is just for easier handling in tests that need one topic and that's all.

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeTestContextExtension.java
im-konge: Yes, you could add the `beforeClassTemplateInvocation` and add its class to the `implements` list.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
im-konge: There is no way right now how to specify the Kafka version, so it will either run on the latest or against all Kafka versions.
Currently, if you specify in the list of Kafka versions some version of Kafka not supported by Test-containers, it will fail at some point. However, once 0.115.0 will be released, we can use the method listing all the supported Kafka versions from the test-containers directly, so we will be sure with what versions we can test.

But yeah, now there is no way how we can check it.

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeExtension.java
im-konge: I will check that, but you should be able to add `beforeAll` and `afterAll` overrides - the reason why the `beforeClassTemplateInvocation` and `afterClassTemplateInvocation` are used in Bridge is that if you use `beforeAll` the Bridge would be deployed before the Kafka cluster - as the order is `beforeAll` -> `beforeClassTemplateInvocation`.

But I will check if we can specify `beforeAll` here as well (but I would be really curious what will be the case of not having one parametrized class).

im-konge: So in case that you want to have before all, it's definitely possible. It will need check if we are in the `ParameterizedClass` or not:
```
    private boolean isParameterizedClass(ExtensionContext context) {
        return context.getRequiredTestClass()
            .isAnnotationPresent(ParameterizedClass.class);
    }
```

and then you can do
```
@Override
    public void beforeAll(ExtensionContext extensionContext) throws Exception {
        if (!isParameterizedClass(extensionContext) && isPerClass(extensionContext)) {
            setupBridge(extensionContext);
        }
    }

    @Override
    public void afterAll(ExtensionContext extensionContext) throws Exception {
        if (!isParameterizedClass(extensionContext) && isPerClass(extensionContext)) {
            stopBridge(extensionContext);
        }
    }
```
the same thing would have to be in the `KafkaExtension`.

For the tests itself you would need another annotation similar to the `BridgeTest` that would just not have the `ParameterizedTest` annotation.

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeTest.java
see-quick: Yeah agree with Bridge, also `BridgeSuite` sounds okayish? Such annotation will be always class-level right?

pom.xml
```diff
@@ -78,7 +78,8 @@
 		<prometheus.version>1.3.6</prometheus.version>
 		<commons-cli.version>1.4</commons-cli.version>
         <!-- Update also list of Kafka version in HttpBridgeITAbstract.java together with test-container version! -->
-		<test-container.version>0.114.0</test-container.version>
+		<strimzi-test-container.version>0.114.0</strimzi-test-container.version>
```
see-quick: probably you can try use rc1 to also update those versions?

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
see-quick: Currently our picture of `TestStorage` is that is has fixed set of topics (there is no underlying logic).  I like the idea abstracting such logic into `TestStorage` i.e.,  we could extend this to have option like this 
```java
.withAdditionalTopics("topic-1", "topic-2", "topic-3", "topic-4")
```
But I think that's not scope of this PR. I am fine having it as we currently have it in operator repo as Lukas already said.

src/test/java/io/strimzi/kafka/bridge/http/base/AbstractIT.java
im-konge: yeah the checkstyle seems to be switched-off for the tests :D so I think you will find more of it. I will check what is happening and at some point turn it on.

pom.xml
im-konge: good point.

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
im-konge: That would be great, but I'm injecting the `TestStorage` to the tests. So that would need some other annotation or something where you would specify other topics, which will be then in the `TestStorageExtension` taken from the annotation and passed to the `TestStorage` that will be injected into the test.

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeExtension.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.reflect.Method;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Properties;
+
+public class BridgeExtension implements
```
ppatierno: Let's add some Javadoc. I won't repeat myself, it's valid for all the other new classes/extensions.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,80 @@
...
+    BeforeClassTemplateInvocationCallback,
+    AfterClassTemplateInvocationCallback {
+    private static final Logger LOGGER = LogManager.getLogger(KafkaExtension.class);
+    private static final Boolean RUN_WITH_ALL_KAFKA_VERSIONS_ENV = Boolean.valueOf(System.getenv().getOrDefault("RUN_WITH_ALL_KAFKA_VERSIONS", "false"));
+
+    private static final String KAFKA_CLUSTER_KEY = "kafkaCluster";
+
+    static List<String> kafkaVersions() {
+        // TODO: replace this with method from strimzi-test-container once we have 0.115.0
+        List<String> listOfVersions = List.of("4.0.1", "4.1.1");
```
ppatierno: I think already asked this question somewhere else but ... isn't it possible to load these versions from a pom.xml property instead of having it in the code?

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,80 @@
...
+            return listOfVersions;
+        }
+
+        // run just with `latest` version of Kafka
+        return List.of("latest");
+    }
+
+    @Override
+    public void beforeClassTemplateInvocation(ExtensionContext context) throws Exception {
+        // ParameterInfo is available HERE because we're at the right level!
```
ppatierno: what does "right level" means here?

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package io.strimzi.kafka.bridge.http.extensions;
+
+import io.strimzi.kafka.bridge.facades.AdminClientFacade;
+import io.strimzi.kafka.bridge.httpclient.HttpRequestHandler;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.util.Random;
+
+public class TestStorage {
```
ppatierno: As mentioned during our offline chat, I would say it's more a `TestContext` than a `TestStorage` ... using the "storage" term always confuses me (the same is in the Strimzi kafka operator repo).

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpRequestHandler.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.logging.log4j.Logger;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.ArrayList;
+import java.util.List;
+
+public class HttpRequestHandler {
```
ppatierno: For me a request handler is something to handle a request but not something like a full HTTP client which sends a request and also return the response to the caller. I think we should find a better name. Isn't it more an HTTP service or an HTTP client facade or something similar?

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
im-konge: Do you want to specify it somewhere else? I mean, the next strimzi-test-container version 0.115.0 will have the method returning all supported versions of Kafka by the version of the test-containers, that we can use (and possibly filter out the older patch releases of the minor versions). IDK if it's a good thing to have them stored somewhere else.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
im-konge: that's just my comment unrelated :D I will remove it.

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpRequestHandler.java
im-konge: HttpService looks good to me.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
ppatierno: ah ok, getting them from the test container would be even better I guess.

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeExtension.java
Frawless: I see, so it doesn't make sense to add it here. Maybe a comment in README about this would be great in case someone will need it in the future. Or in Javadoc, up to you.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
Frawless: So default will be `all` and user can switch to `RUN_WITH_ALL_KAFKA_VERSIONS=false` and latest will be used?

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
Frawless: Don't make the things more complex than it has to be. 

If in tests that needs more topics will user define one as method-scoped variable and for second will use one form TS....it is strange from my POV. I understand the point why it is useful, but I am not sure if it's the best idea.

src/test/java/io/strimzi/kafka/bridge/http/extensions/KafkaExtension.java
im-konge: the current default is to run against only the latest version (with `RUN_WITH_ALL_KAFKA_VERSIONS` being set to `false` by default). But if you switch it to `true`, it will run against all the versions.

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
im-konge: I don't want to make it complex.
In the `AdminIT` there is one test creating two topics, other tests used just one topic. From my side, I don't feel like it's a big deal not having more than just one topic in the `TestStorage`. If you need in one or two tests more topics, cool, you can create them in the test itself. But I would not complicate it that much because of the topic names.

Or what was your idea about that -> you want to remove it from the `TestStorage` completely and use some utils method or something?

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeExtension.java
im-konge: I was thinking about having it in Javadoc, but FMPOV it would be long and maybe a bit messy. So I will add note that if someone would like to implement the beforeAll/afterAll stuff, there is doc on path which they can check. And I will write it to some README.md

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeTest.java
im-konge: I used `BridgeSuite` for now, but I'm open for any other suggestion as well.

src/test/java/io/strimzi/kafka/bridge/http/extensions/BridgeTest.java
im-konge: I can mention in Javadoc from where the Kafka versions are taken. But the versions will not be changed once we have 0.115.0 of Strimzi Test Container released -> we will use the versions directly from the method it provides.

src/test/java/io/strimzi/kafka/bridge/http/extensions/TestStorage.java
im-konge: @Frawless @see-quick WDYT? Should I change it to `TestContext` or do you have any other idea?

see-quick: I mean it would be great that we are consistent across the whole codebase. And as creator of that stuff I think the term `TestStorage` here is okay.  The class is **data holder** it doesn't control test lifecycle (i.e., it's basically bag of pre-built objects that test pull from => storage). 

I think I previously consider such alterantive i.e., naming it TestContext but I think would create more confusion with JUnit5 as it has it's own `ExtensionContext` and the `ExtensionContext.Store`. 

Moreover the class has no "context" behaviour ... I mean IMO context typically implies something that influence execution but here is just plain storage of objects which are then used in the tests. 

While I can understand that storage might also evoke Kafka/Kuberentes storage concepts ... but in the testing context it should be clear :)

im-konge: So what about `TestDataHolder` 😀

src/test/java/io/strimzi/kafka/bridge/enums/BridgeRunMode.java
```diff
@@ -0,0 +1,25 @@
...
+    public static BridgeRunMode fromString(String value) {
+        String lowerCase = value.toLowerCase(Locale.ROOT);
+
+        if (CONTAINER.name().toLowerCase(Locale.ROOT).equals(lowerCase)) {
+            return CONTAINER;
+        } else if (IN_MEMORY.name().toLowerCase(Locale.ROOT).equals(lowerCase)) {
+            return IN_MEMORY;
+        }
+
+        return NONE;
```
see-quick: wouldn't be better to throw ISE here?

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Method returning list of Kafka versions for which we will run all the tests - because of the {@link org.junit.jupiter.params.ParameterizedClass}.
+     * Based on {@link #RUN_WITH_ALL_KAFKA_VERSIONS_ENV} the tests will be executed against all Kafka versions, or the latest one.
+     *
+     * @return  list of Kafka versions for which we will run the tests.
+     */
+    static List<String> kafkaVersions() {
+        // TODO: replace this with method from strimzi-test-container once we have 0.115.0
+        List<String> listOfVersions = List.of("4.0.1", "4.1.1");
```
see-quick: you can replace with method :) as 0.115.0 is released

src/test/java/io/strimzi/kafka/bridge/objects/MessageRecord.java
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.bridge.objects;
+
+public record MessageRecord(
+    String key,
+    String value
+) {
+}
```
see-quick: this looks weird... maybe having it as `MessageRecord(String key, String value) {}`?  Like message record would not have more than two parameters and if so it doesn't matter (just my preference).

src/test/java/io/strimzi/kafka/bridge/enums/BridgeRunMode.java
im-konge: Yeah makes sense

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
im-konge: Yeah I was waiting for the PR to merged in `main` branch, so I can rebase and use it :)

src/test/java/io/strimzi/kafka/bridge/objects/MessageRecord.java
im-konge: that was checkstyle which said that it should be on newline .. I can have a look if it can be a bit different

im-konge: yeah I changed everything to have better styling, the checkstyle passed

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,168 @@
...
+ * Extension (or fixture) for managing Kafka cluster in tests.
+ * Because of the {@link org.junit.jupiter.params.ParameterizedClass}, this extension currently supports deploying and
+ * stopping the Kafka cluster in {@link #beforeClassTemplateInvocation(ExtensionContext)} and {@link #afterClassTemplateInvocation(ExtensionContext)}.
+ * However, in case that we decide to not support running multiple Kafka versions, we can implement beforeAll, afterAll, beforeEach, or afterEach methods.
+ */
+public class KafkaExtension implements
+    BeforeClassTemplateInvocationCallback,
+    AfterClassTemplateInvocationCallback {
+    private static final Logger LOGGER = LogManager.getLogger(KafkaExtension.class);
+    private static final Boolean RUN_WITH_ALL_KAFKA_VERSIONS_ENV = Boolean.valueOf(System.getenv().getOrDefault("RUN_WITH_ALL_KAFKA_VERSIONS", "true"));
```
Frawless, line 30: I would mention it in some README that users can use just latest kafka by setting the env var to false.

Frawless (approved):

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+     *         HttpConfig.HTTP_CORS_ALLOWED_METHODS + "=GET,POST,PUT,DELETE,OPTIONS,PATCH"
+     *     }
+     * )
+     * </pre>
+     *
+     * Note: it's a String array where each element is one key-value pair separated by `=`.
+     *
+     * @return  array of Strings containing configuration of Bridge
+     */
+    String[] properties() default {
```
ppatierno: Out of curiosity, if we need to have key-value pairs, why it can't be a Map or an instance of a Properties class?

src/test/java/io/strimzi/kafka/bridge/enums/BridgeRunMode.java
```diff
@@ -0,0 +1,25 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.bridge.enums;
+
+import java.util.Locale;
+
+public enum BridgeRunMode {
+    CONTAINER,
+    IN_MEMORY,
+    NONE;
```
ppatierno: what does NONE mean here?

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
ppatierno: Yeah, agree.

src/test/java/io/strimzi/kafka/bridge/objects/TestStorage.java
```diff
@@ -0,0 +1,59 @@
...
+import io.strimzi.kafka.bridge.facades.AdminClientFacade;
+import io.strimzi.kafka.bridge.httpclient.HttpService;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.util.Random;
+
+/**
+ * Class used for storing variables, services, and all information used in tests.
+ */
+public class TestStorage {
```
ppatierno: As mentioned offline I think "storage" is confusing here, as well as it's used in the Strimzi operator. For me, it represents a "context" because as it mentioned in the Java doc it has "variables, services, and all information used in tests.". I would like to know what other @strimzi/tenured-maintainers think about that before making this change.

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
im-konge: The annotation takes just primitive types, classes, or another annotations.
So instead of having the String array, you can have for example something like this:

```
@Config({
    @Entry(key = "host", value = "localhost"),
    @Entry(key = "port", value = "5432")
})
```
but I thought that it can get really long... but I'm not against it TBH :) it's just up to you what makes more sense to you.

src/test/java/io/strimzi/kafka/bridge/enums/BridgeRunMode.java
im-konge: It's just leftover, instead of `NONE` we are throwing an exception.

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
ppatierno: Ok, despite being longer, I don't expect tests changing a lot of properties, so I am more for the key-value entries approach instead of strings but as usual, let's see what other @strimzi/maintainers think.

src/test/java/io/strimzi/kafka/bridge/enums/BridgeRunMode.java
ppatierno: So does it need to be removed right?

im-konge: Yeah I will remove it

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
see-quick: I don't have strong opinion here ... choose whatever you feel is better this is just syntax sugar.

im-konge: It seems that I will add these key-value annotations as you suggested @ppatierno (other than Maros there wasn't anyone else replying, so I will take it as it is).

src/test/java/io/strimzi/kafka/bridge/objects/TestStorage.java
im-konge: Okay so it's been 4 days without anyone replying. So should we keep it as it is and once we come up with better name, we can change it? I don't want to be stuck for so long on this, given that nobody else replies (or there is no strong opinion).

Frawless: I am not big fan of `TestStorage` name, but `TestContext` is probably not great naming. We operate with kube context within `kubetest4j` within operators repository and there is also `ExtensionContext` from jUnit that is kinda "test context". Maybe `BridgeTestContext` ?

im-konge: I don't have better name at my mind right now, I think I can live with that. Other names would be `TestEnvironment` or `TestResources`. But those don't look much better then `TestStorage`

So I guess `BridgeTestContext` can do the trick.

Frawless: Fine by me.

im-konge: @ppatierno @see-quick WDYT? Should I go with `BridgeTestContext`?

ppatierno: Fine with me

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
im-konge: Btw it looks like this:
```
    @BridgeConfiguration(
        properties = {
            @ConfigEntry(key = HttpConfig.HTTP_CORS_ENABLED, value = "false"),
            @ConfigEntry(key = HttpConfig.HTTP_CORS_ALLOWED_ORIGINS, value = "https://strimzi.io"),
            @ConfigEntry(key = HttpConfig.HTTP_CORS_ALLOWED_METHODS, value = "GET,POST,PUT,DELETE,OPTIONS,PATCH")
        }
    )
```
I think that's nice.

src/test/java/io/strimzi/kafka/bridge/objects/TestStorage.java
see-quick: I still think we should keep naming consistent with the operators repo (i.e., TestStorage) to avoid contributors having to learn different names for the same pattern across Strimzi repos, but I don't want to block progress on a naming discussion. Feel free to go ahead with BridgeTestContext.

ppatierno: On the other side I think that the operator tests should be changed to use more "context" naming than "storage" which, inside the operator itself is really confusing, but I don't want to inject my thoughts in the QE part if you like "storage".

im-konge: TBH I thought we have consensus here :D so I stopped watching the discussion. Anyway, in operators we can have something like `StrimziTestContext`, we can change the naming.

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
im-konge: I updated the README.

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation handling the configuration of the Bridge cluster.
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
```
ppatierno, line 22: Why it's not just `ElementType.TYPE` to be used on a class. I can't see we use the `BridgeConfiguration` annotation on a method.

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
```diff
@@ -0,0 +1,380 @@
...
+     * It is then used in the before and after methods to check if the Bridge cluster should be started/stopped at the particular stage.
+     *
+     * @param extensionContext  context of the test.
+     *
+     * @return  boolean value determining if the {@link org.junit.jupiter.api.TestInstance.Lifecycle} is set to {@link org.junit.jupiter.api.TestInstance.Lifecycle#PER_CLASS}
+     */
+    private boolean isPerClass(ExtensionContext extensionContext) {
+        return extensionContext.getTestInstanceLifecycle()
+            .map(lifecycle -> lifecycle == TestInstance.Lifecycle.PER_CLASS)
+            .orElse(false);
```
ppatierno, line 359: is the returning "false" ok? I mean, is there others lifecycle values other than PER_CLASS or PER_METHOD?

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
```diff
@@ -0,0 +1,380 @@
...
+     * @param extensionContext  context of the test.
+     * @param configuration     configuration of the Bridge.
+     *
+     * @throws IOException    exception during creating properties file.
+     */
+    private void deployBridgeInContainer(ExtensionContext extensionContext, Map<String, Object> configuration) throws IOException {
+        Properties properties = new Properties();
+        properties.putAll(configuration);
+        String propertiesPath = createPropertiesFile(properties);
+
```
ppatierno, line 282: Let's add a log like we have for the in-memory, like `LOGGER.info("Deploying in-memory Bridge");`

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
```diff
@@ -0,0 +1,380 @@
...
+        Vertx vertx =  Vertx.vertx(vertxOptions);
+
+        vertx.deployVerticle(httpBridge)
+            .onComplete(ar -> {
+                if (ar.succeeded()) {
+                    LOGGER.info("Bridge deployed: " + ar.result());
+                } else {
+                    LOGGER.error("Failed to deploy Bridge", ar.cause());
+                }
+            }).await();
```
ppatierno, line 262: looks like if there is any error to deploy the verticle, we just log the error but didn't raise any exception or stop testing?

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
```diff
@@ -0,0 +1,168 @@
...
+     * Method that setups and starts the Kafka cluster based on the {@param kafkaVersion}, using the {@link StrimziKafkaCluster}.
+     * In case that {@param kafkaVersion} is null, empty, or set to `latest`, we will skip configuring the Kafka version, so the latest
+     * version supported in Strimzi Test Containers is used.
+     *
+     * @param extensionContext  context of the test.
+     * @param kafkaVersion      version of Kafka that should be used in Kafka cluster.
+     */
+    private void setupKafkaCluster(ExtensionContext extensionContext, String kafkaVersion) {
+        StrimziKafkaCluster.StrimziKafkaClusterBuilder kafkaBuilder = new StrimziKafkaCluster.StrimziKafkaClusterBuilder()
+            .withNumberOfBrokers(1)
```
ppatierno, line 131: shouldn't we make the number of brokers configurable? I am not sure but what if we have tests creating topics with higher replication factor? Their creation will fail because just one broker.

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
im-konge: It's used on methods - https://github.com/strimzi/strimzi-kafka-bridge/pull/1074/changes#diff-d1ea9d42bda17c762e1ff33f40c057d8d9b55feff78c83c5874d5c8ee37d94d4R88
For now it's mainly used in `HttpCorsIT` where every test has different configuration.

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
im-konge: It's per class and per method. You only care if it's configured with `PER_CLASS`, so `false` in the else is completely fine.

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
im-konge: hmm good point, I will throw exception there. Thanks!

src/test/java/io/strimzi/kafka/bridge/extensions/KafkaExtension.java
im-konge: I was thinking about adding `KafkaConfiguration` as we have for Bridge, but currently we are using the same number of brokers and the same config - I mean in the tests I changed. I didn't want to add something that will not be used in the current code - or just the default will be used. So maybe I can add it there once it will be needed (during refactoring of other tests)?

src/test/java/io/strimzi/kafka/bridge/configuration/BridgeConfiguration.java
ppatierno: Ok I am blind, sorry ;-)

PR #1087, opened 2026-03-11
Various Java cleanups

mimaison:
- Remove unused methods
- Javadoc cleanups
- Fix non constant log calls
- Fix typos

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -62,6 +62,7 @@ import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -194,7 +195,9 @@ public class HttpBridge extends AbstractVerticle {
         Long timeoutInMs = timeout * 1000L;
         vertx.setPeriodic(timeoutInMs / 2, ignore -> {
             LOGGER.debug("Looking for stale consumers in {} entries", timestampMap.size());
-            for (Map.Entry<ConsumerInstanceId, Long> item : timestampMap.entrySet()) {
+            Iterator<Map.Entry<ConsumerInstanceId, Long>> it = timestampMap.entrySet().iterator();
+            while (it.hasNext()) {
+                Map.Entry<ConsumerInstanceId, Long> item = it.next();
                 if (item.getValue() + timeoutInMs < System.currentTimeMillis()) {
                     HttpSinkBridgeEndpoint<byte[], byte[]> deleteSinkEndpoint = this.httpBridgeContext.getHttpSinkEndpoints().get(item.getKey());
                     if (deleteSinkEndpoint != null) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/strimzi/kafka/bridge/http/converter/JsonEncodeException.java
```diff
@@ -25,13 +19,4 @@ public JsonEncodeException(String message) {
...
 
-    /**
-     * Constructor
-     *
-     * @param message Exception message
-     * @param cause Inner cause of the exception
-     */
-    public JsonEncodeException(String message, Throwable cause) {
-        super(message, cause);
-    }
```
ppatierno, line 36: instead of removing this one, can you update the code within the `JsonUtils` where the other ctor is used and using this instead? This way we have consistency with the ctor we use for the `JsonDecodeException`.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -195,9 +194,7 @@ private void startInactiveConsumerDeletionTimer(Long timeout) {
         Long timeoutInMs = timeout * 1000L;
         vertx.setPeriodic(timeoutInMs / 2, ignore -> {
             LOGGER.debug("Looking for stale consumers in {} entries", timestampMap.size());
-            Iterator<Map.Entry<ConsumerInstanceId, Long>> it = timestampMap.entrySet().iterator();
-            while (it.hasNext()) {
-                Map.Entry<ConsumerInstanceId, Long> item = it.next();
+            for (Map.Entry<ConsumerInstanceId, Long> item : timestampMap.entrySet()) {
```
ppatierno: I remember that the loop was done this way because Java didn't allow to remove elements from the collection you are going through and it raised an exception. Maybe is something changed now?

src/test/java/io/strimzi/kafka/bridge/http/ConsumerIT.java
```diff
@@ -1719,7 +1719,7 @@ void consumerDeletedAfterInactivity(VertxTestContext context) {
                         assertThat(consumerInstanceId, is(name));
                         assertThat(consumerBaseUri, is(Urls.consumerInstance(groupId, name)));
 
-                        vertx.setTimer(timeout * 2 * 1000L, timeouted -> {
+                        vertx.setTimer(timeout * 2 * 1000L, timedouted -> {
```
ppatierno: isn't it more "timedout"?

src/main/java/io/strimzi/kafka/bridge/http/converter/JsonEncodeException.java
mimaison: Good idea!

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
mimaison: That's a good point. Since it's a cleanup PR, let's just revert to the iterator logic and not risk breaking this

src/test/java/io/strimzi/kafka/bridge/http/ConsumerIT.java
mimaison: Yes it's what I meant to use but I somehow mess it up. As the variable is not used, I've renamed it to `ignore`

PR #1089, opened 2026-03-13
Use Strimzi GitHub Actions v1

Frawless:
This PR freezes version of Strimzi GitHub actions to [v1](https://github.com/strimzi/github-actions/releases/tag/v1.0)

This PR also adds integration tests to ensure that build is working all the time.

Closes https://github.com/strimzi/github-actions/issues/10

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,6 @@ jobs:
         with:
           mainJavaBuild: ${{ matrix.java.mainBuild }}
           artifactSuffix: "kafka-bridge"
-          clusterOperatorBuild: "false"
 
   build-docs:
     name: Build docs
--- a/.github/workflows/github-actions-integration.yml
+++ b/.github/workflows/github-actions-integration.yml
@@ -37,6 +37,5 @@ jobs:
       helmChartName: "none"
       releaseVersion: "6.6.6"
       imagesDir: "kafka-bridge-amd64.tar.gz"
-      clusterOperatorBuild: false
       githubActionsRef: "v1"
     secrets: inherit


.github/workflows/build.yml
```diff
@@ -32,11 +32,18 @@ jobs:
...
           javaVersion: ${{ matrix.java.version }}
-          mainBuild: ${{ matrix.java.mainBuild }}
-          runnerArch: "amd64"
+
+      - name: Build binaries using build-binaries action
+        uses: strimzi/github-actions/.github/actions/build/build-binaries@v1
+        with:
+          mainJavaBuild: ${{ matrix.java.mainBuild }}
+          artifactSuffix: "kafka-bridge"
+          clusterOperatorBuild: "false"
```
ppatierno: what's this parameter? if it's used just within the repo to build the operator, can we have its default value to false to avoid to put it into all the others?

.github/workflows/github-actions-integration.yml
```diff
@@ -0,0 +1,42 @@
...
+      ref: ${{ github.sha }}
+      architecture: "amd64"
+      artifactSuffix: "kafka-bridge"
+      buildContainers: true
+      modules: "./"
+      nexusCheck: "kafka-bridge"
+      javaVersion: "17"
+      helmChartName: "none"
+      releaseVersion: "6.6.6"
+      imagesDir: "kafka-bridge-amd64.tar.gz"
```
ppatierno, line 39: the naming seems to ask for a dir with images? but then I see a tar.gz file. What is this then?

.github/workflows/github-actions-integration.yml
```diff
@@ -0,0 +1,42 @@
...
+      clusterOperatorBuild: false
```
ppatierno: again I think if it's used in the operator only we should have false as default to avoid having this in the bridge (or maybe all the other repositories)?

.github/workflows/github-actions-integration.yml
Frawless: More precise would be `imagesLocation`. I can open issue and fix it in the upcoming releases, but v1 is already freezed.

.github/workflows/build.yml
Frawless: Default is false. It is here just for better visibility, same as matrix for Java config in other repos as we talked about in the past. If you prefer removal of it, I can remove it in the bridge workflows.

.github/workflows/github-actions-integration.yml
ppatierno: Yeah that would be better. Thanks.

.github/workflows/build.yml
ppatierno: I think that it makes only confusion. If it defaults to "false" than it should not be there, not only here in the bridge but across all the other components, except the operator where you would set it to "true" AFAIU.

.github/workflows/github-actions-integration.yml
Frawless: I opened https://github.com/strimzi/github-actions/issues/26

.github/workflows/build.yml
Frawless: OK, it is removed from the workflows.

ppatierno (approved):

PR #1116, opened 2026-04-29
Refactor integration tests vol.2 - `ConsumerGeneratedNameIT` and `ConsumerIT`

im-konge:
This PR refactors another part of the integration tests - mainly the `ConsumerGeneratedNameIT` and `ConsumerIT`.
There is nothing much special about `ConsumerGeneratedNameIT`, however the second class - `ConsumerIT` - was longer and there were the most changes. As part of refactoring the tests, I found out there are some duplicates - so I removed them. Or I moved some checks to different tests to not do the same from a different side and also to save some time.

I created few new classes - `HttpConsumerService` and `HttpProducerService` - those should be replacements to the classes that use VertX - `ConsumerService` and `ProducerService`. I didn't removed these two yet as there are still some tests using these classes, the whole clean-up will be something for last PR.

Also, I created `Endpoints` class where we will store all the possible endpoints we want to test - the reason why I moved away from `Urls` class is that we are building the requests differently now.

There were also some changes inside the Kafka clients stuff used in tests - mainly to remove the VertX stuff and as form of a cleanup. The best thing would be to use Kafka clients from test-clients, so we will not write the same thing over and over again.

I created `HttpError` instead of the `HttpBridgeError` - again, we are getting different type of responses - with VertX we are getting VertX's JSON object, with the Java's HTTP stuff we are getting String inside the `HttpResponse`.

Finally, I added possibility to remove default configuration for Bridge - in cases that we don't want to create the whole Bridge configuration again and we want to be sure these fields will be really deleted.


src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
```diff
@@ -5,21 +5,16 @@
...
 import java.util.List;
 import java.util.Properties;
 import java.util.Random;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.function.IntPredicate;
 
 public class BasicKafkaClient {
```
ppatierno, line 19: Looking into these testing classes more thanks to your refactoring effort I was wondering why we need this class. I guess the goal is having some Kafka clients to be used within tests to produce/consume messages via Kafka protocol (so wrappers around the Kafka client itself). I can't understand why we have this `BasicKafkaClient` exposing methods to send messages which uses a `Producer` instance. Can't we use just the `Producer` instance in the tests when needed?

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpError.java
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.bridge.httpclient;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public record HttpError(int code, String message, List<String> validationErrors) {
```
ppatierno: I read your comment in the description about this addition but looking at `HttpBridgeError` class, its `fromJson` method is used just within the tests. In the past I already did a refactoring work to not using Vert.x `JsonObject` anymore but pure Jackson classes. Can't you modify the `HttpBridgeError.fromJson` accordingly? It could get an `ObjectNode` as from Jackson which you could build from the HTTP response which is a String but still representing a JSON. Maybe the `JsonUtils` class can be helpful. I am thinking aloud to avoid adding classes that we actually don't need if we can refactor what we already have.

src/test/java/io/strimzi/kafka/bridge/objects/ReceivedMessage.java
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.bridge.objects;
+
+public record ReceivedMessage(String topic, Object key, Object value, Integer partition, Long offset, String timestamp, MessageHeader[] headers) {
```
ppatierno, line 7: Not sure we really need a class to wrap received messages. Can we stick with handling JSON as with Vert.x? Of course, not using `JsonObject` but maybe `ObjectNode` from Jackson. As mentioned in a different comment I did a refactoring to use `ObjectNode` instead of Vert.x JSON stuff, maybe we should do the same within the tests.

src/test/java/io/strimzi/kafka/bridge/Constants.java
```diff
@@ -10,4 +10,9 @@ public interface Constants {
      * Tag for http bridge tests, which are triggered for each push/pr/merge on travis-ci
      */
     String HTTP_BRIDGE = "httpbridge";
+
+    String DEFAULT_BRIDGE_ID = "my-bridge";
+
+    long DEFAULT_CONSUMER_TIMEOUT = 5;
+    String DEFAULT_CONSUMER_TIMEOUT_STRING = "5";
```
ppatierno, line 24: why do we need both as long and String?
Also, can we add doc for these constants?

ppatierno (commented):
I had a pass, difficult to check everything because of a lot of changes, but I left some first comments mostly related to JSON (from HTTP response) handling.

src/test/java/io/strimzi/kafka/bridge/clients/BasicKafkaClient.java
im-konge: Well ultimately I would like to use test-clients for the Kafka part, which doesn't need changing. But I guess it's not for one PR - when you already said it's long enough. So I will do it in different PR, I tried to reduce the changes as much as I could.

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpError.java
im-konge: How about adding `fromString` to the `HttpBridgeError` instead of parsing it every time in every test to Json and then to the object?
Maybe you did refactoring, but I see that `HttpBridgeError` contains Vertx object.
I see it even on the main branch:

```
import com.fasterxml.jackson.databind.node.ObjectNode;
import io.strimzi.kafka.bridge.http.converter.JsonUtils;
import io.vertx.core.json.JsonObject;

import java.util.ArrayList;
import java.util.List;

/**
 * Represents an error related to HTTP bridging
 *
 * @param code  code classifying the error itself
 * @param message   message providing more information about the error
 * @param validationErrors  list of detailed validation errors
 */
public record HttpBridgeError(int code, String message, List<String> validationErrors) {
```

src/test/java/io/strimzi/kafka/bridge/Constants.java
im-konge: One is for the BridgeConfiguration, which needs to be constants (so without `String.valueOf()` stuff) - as we have key/value pairs in String (it needs to be primitive data types, you cannot pass there `Object` for example) and the other one (long) is used in tests.

src/test/java/io/strimzi/kafka/bridge/objects/ReceivedMessage.java
im-konge: I guess I can do it, but FMPOV this was much clearer than parsing it to map, then using the Map everywhere. You can say the same thing for other classes I created for easier handling of things. And IMO it should be consistent - so we should not have something in classes, something in Maps. We should pick just one approach.

im-konge:
> I had a pass, difficult to check everything because of a lot of changes, but I left some first comments mostly related to JSON (from HTTP response) handling.

Sorry that it has a lot of changes, but I couldn't split it more given that the ConsumerIT itself had 2200 lines of code to be refactored. I'm not sure in how much smaller pieces I should do the PRs.

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpError.java
ppatierno: > Maybe you did refactoring, but I see that HttpBridgeError contains Vertx object.

Yes the Vert.x `JsonObject` is used just within the `fromJson` method which is used in tests only. When I did the refactoring I couldn't remove it because of tests still based on Vert.x. Thanks to your refactoring we can remove this last piece of Vert.x JSON stuff within the `HttpBridgeError` class and you could reuse it.

src/test/java/io/strimzi/kafka/bridge/objects/ReceivedMessage.java
ppatierno: Are you talking about classes from the previous PR about refactoring or something new in this PR?

src/test/java/io/strimzi/kafka/bridge/Constants.java
ppatierno: But if they are going to be the same can't you get one from the other?

ppatierno:
> Sorry that it has a lot of changes, but I couldn't split it more given that the ConsumerIT itself had 2200 lines of code to be refactored. I'm not sure in how much smaller pieces I should do the PRs.

No worries, I am not complaining at all. It was just to set expectations on my review ... it's not going to be so easy and fast ;-)

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpError.java
im-konge: I guess it will not be in this PR - like fully removing it. I would rather have method to just use `fromResponse` or something instead of building it somewhere else. For now, I can add another method to handle it in the HttpBridgeError -> but until it is finished, we cannot really remove the Vertx stuff from there unfortunately.

src/test/java/io/strimzi/kafka/bridge/objects/ReceivedMessage.java
im-konge: From the previous PR and this.
Here is the full list:
```
src/test/java/io/strimzi/kafka/bridge/objects
├── BridgeTestContext.java
├── MessageHeader.java
├── MessageRecord.java
├── Offsets.java
├── Partition.java
├── ReceivedMessage.java
├── Records.java
├── Replica.java
└── Topic.java
```
The `BridgeTestContext` is actually holder for the info for each test, it should be kept as it is.

src/test/java/io/strimzi/kafka/bridge/Constants.java
im-konge: I can get long from the String, but not the other way around - as I said the config has to be String

src/test/java/io/strimzi/kafka/bridge/httpclient/HttpError.java
ppatierno: I think it's important and more consistent that the tests use the `HttpBridgeError`. In the end the `fromJson` method is there just for the tests, because the bridge always produces an `HttpBridgeError` so it always uses only the `toJson`. I am fine with adding a new overload for `fromJson` getting a String representing a JSON.

src/test/java/io/strimzi/kafka/bridge/objects/ReceivedMessage.java
ppatierno: So ... my plan (at least in my mind) was always to refactor the bridge to move POJO around to be serialized/deserialized as JSON across HTTP requests/responses payload but having POJOs within the bridge. The final end could be using such POJOs within the tests as well. I was wondering if it's better to stick with JSON handling in the tests as it happens today in the bridge itself, and moving to POJOs when the core bridge code has them.

im-konge: Or you can then reuse them in the core Bridge :D depends on what is better for you.

ppatierno: :-D Let's stick with what you have for now by adding these classes. I had a look at the core bridge code and I am not so sure we would need POJOs anymore. We are anyway moving across JSON and Kafka classes using the message converters, not sure an additional layer of POJOs would bring advantages.

src/test/java/io/strimzi/kafka/bridge/Constants.java
ppatierno: so you can define the long constant by converting the String one instead of having two separate values, right?

im-konge: Yeah will do that :)

src/test/java/io/strimzi/kafka/bridge/objects/ReceivedMessage.java
im-konge: :D Anyway, you can say at the end of the refactoring that you don't like it and I will remove it from everywhere. I just thought it would be nice handling in the tests, which would help to not manipulate with Maps directly (so you don't need to remember the structure, run the tests to see the structure etc.)

src/test/java/io/strimzi/kafka/bridge/http/ConsumerIT.java
```diff
@@ -6,2037 +6,1225 @@
...
-
-        context.completeNow();
-        assertThat(context.awaitCompletion(TEST_TIMEOUT, TimeUnit.SECONDS), is(true));
+        HttpResponse<String> httpResponse = httpConsumerService.createConsumerRequest(groupId, consumerWithWrongFormat);
+
+        assertThat(httpResponse.statusCode(), is(HttpResponseStatus.UNPROCESSABLE_ENTITY.code()));
+        HttpBridgeError httpBridgeError = HttpBridgeError.fromJson(HttpResponseUtils.getResponseAsMap(httpResponse.body()));
+
+        assertThat(httpBridgeError.code(), is(HttpResponseStatus.UNPROCESSABLE_ENTITY.code()));
+        assertThat(httpBridgeError.message(), is("Invalid format type."));
```
ppatierno, line 97: I see this formatting in several places would be better to have the line which get the response followed by the assertion, then new line, then line getting the error then assertion. The blank lines position is weird. I would expect something more like:

```java
HttpResponse<String> httpResponse = httpConsumerService.createConsumerRequest(groupId, consumerWithWrongFormat);
assertThat(httpResponse.statusCode(), is(HttpResponseStatus.UNPROCESSABLE_ENTITY.code()));

HttpBridgeError httpBridgeError = HttpBridgeError.fromJson(HttpResponseUtils.getResponseAsMap(httpResponse.body()));
assertThat(httpBridgeError.code(), is(HttpResponseStatus.UNPROCESSABLE_ENTITY.code()));
assertThat(httpBridgeError.message(), is("Invalid format type."));
```

It's clearer where each assertions belongs to.

im-konge: Ok, done

ppatierno (approved):
LGTM. Thanks!

PR #1117, opened 2026-05-02
Use switch pattern matching in buildOffsets method

### LINKED ISSUE #1112: Using types Java pattern matching within `HttpSourceBridgeEndpoint.buildOffsets` method
The `HttpSourceBridgeEndpoint.buildOffsets` has an `if/else if` block depending on the `result.result()` type which could be improved by using Java pattern matching on the type.
It would have a switch on the `result.result()` and cases for each of the two types `RecordMetadata` and `HttpBridgeError`.
The result of the switch would be assigned to the `offset` local variable.

guancioul:
Close #1112,

Refactor buildOffsets to use switch pattern matching.

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
@@ -212,14 +212,13 @@ public class HttpSourceBridgeEndpoint<K, V> extends HttpBridgeEndpoint {
         ArrayNode offsets = JsonUtils.createArrayNode();
 
         for (HttpBridgeResult<?> result : results) {
-            ObjectNode offset = null;
-            if (result.result() instanceof RecordMetadata metadata) {
-                offset = JsonUtils.createObjectNode()
+            ObjectNode offset = switch (result.result()) {
+                case RecordMetadata metadata -> JsonUtils.createObjectNode()
                         .put("partition", metadata.partition())
                         .put("offset", metadata.offset());
-            } else if (result.result() instanceof HttpBridgeError error) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
```diff
@@ -212,14 +212,13 @@ private ObjectNode buildOffsets(List<HttpBridgeResult<?>> results) {
...
+            ObjectNode offset = switch (result.result()) {
+                case RecordMetadata metadata -> JsonUtils.createObjectNode()
                         .put("partition", metadata.partition())
                         .put("offset", metadata.offset());
-            } else if (result.result() instanceof HttpBridgeError error) {
-                offset = error.toJson();
-            }
+                case HttpBridgeError error -> error.toJson();
+                default -> null;
+            };
```
scholzj, line 221: I would probably prefer a more old-school switch here ...

```
ObjectNode offset;

switch ...
    case ...:
        offset = ...
    ...
```

ppatierno: I am fine with the current proposed code using the switch expressions.

ppatierno (approved):
LGTM. Thanks!

src/main/java/io/strimzi/kafka/bridge/http/HttpSourceBridgeEndpoint.java
scholzj: Ok, lets stick with it here.

scholzj (approved):

PR #1121, opened 2026-05-12
Refactor `ConsumerSubscriptionIT` and `DisablingConsumerProducerIT`

im-konge:
This PR continues with refactoring integration tests in this repository - now with `ConsumerSubscriptionIT` and `DisablingConsumerProducerIT`

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
+++ b/src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
@@ -4,6 +4,7 @@
  */
 package io.strimzi.kafka.bridge.http;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.strimzi.kafka.bridge.extensions.BridgeSuite;
 import io.strimzi.kafka.bridge.http.base.AbstractIT;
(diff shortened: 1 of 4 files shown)


src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -5,561 +5,280 @@
...
-                .deleteConsumer(context, groupId, name);
-        context.completeNow();
+        httpConsumerService.consumeRecordsRequest(groupId, name);
+
+        // validate subscription list has 1 topic
+        HttpResponse<String> httpResponse = httpConsumerService.listSubscriptionsRequest(groupId, name);
+        assertThat(httpResponse.statusCode(), is(HttpResponseStatus.OK.code()));
+
+        Map<String, Object> responseBody = HttpResponseUtils.getResponseAsMap(httpResponse.body());
+        Object[] topics = (Object[]) responseBody.get("topics");
```
ppatierno: This is not specific to this PR and maybe something we can address later but I am not so keen to have generic Object array moving around. Any response from the bridge is JSON so I think we should aim to deal with it, if it makes sense.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
```diff
@@ -5,561 +5,280 @@
...
-        context.completeNow();
+        Object[] partitions = (Object[]) responseBody.get("partitions");
+        assertThat(partitions.length, is(2));
+
+        // partitions contain objects like {topicName: [0, 1, 2, 3]}
+        Map<String, Object> part0 = (Map<String, Object>) partitions[0];
+        Map<String, Object> part1 = (Map<String, Object>) partitions[1];
+
+        if (part0.containsKey(topic2)) {
+            assertThat(((Object[]) part0.get(topic2)).length, is(4));
```
ppatierno: yeah talking about Object array ... I think this line with casting to that is pretty bad comparing to the previous JSON stuff.

ppatierno (approved):
Going to approve this but I left comments about the usage of Object array. I think the code should be improved by coming back to deal with JSON as it's the representation we get from the bridge. Not for this PR but after refactoring the overall tests we should have another pass imho.

src/test/java/io/strimzi/kafka/bridge/http/ConsumerSubscriptionIT.java
im-konge: Let me try to fix it partially in this PR, so we don't have it there for the whole time. Thanks

im-konge:
@ppatierno I added method for returning the response as JsonNode and updated the tests accordingly. WDYT?

ppatierno:
> @ppatierno I added method for returning the response as JsonNode and updated the tests accordingly. WDYT?

It looks much better. Thanks!

PR #1122, opened 2026-05-13
Refactor `InvalidProducerIT`, `OtherServicesIT`, and `TlsIT` with changes to BridgeExtension due to TLS Bridge

im-konge:
This PR continues with refactoring the integration tests, now with `InvalidProducerIT`, `OtherServicesIT`, and `TlsIT`.
There were needed changes for the `TlsIT`, as the handling of the Bridge stuff in container (and also in-memory Bridge) is a bit different (in terms of the HTTP ports etc.), so I had to add changes to the `BridgeExtension` as well.

Also, I had to generate correct certificate for the `TlsIT` with SANs, that are required when doing the HTTP requests with the Java's HTTP client - that could be disabled with `.setVerifyHost(false)` in the VertX's Webclient.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
+++ b/src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
@@ -10,7 +10,9 @@ import io.strimzi.kafka.bridge.configuration.BridgeConfiguration;
 import io.strimzi.kafka.bridge.configuration.ConfigEntry;
 import io.strimzi.kafka.bridge.enums.BridgeRunMode;
 import io.strimzi.kafka.bridge.http.HttpBridge;
+import io.strimzi.kafka.bridge.http.HttpConfig;
 import io.strimzi.kafka.bridge.httpclient.HttpService;
+import io.strimzi.kafka.bridge.utils.Urls;
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxOptions;
 import io.vertx.micrometer.Label;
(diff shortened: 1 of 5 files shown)


src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
```diff
@@ -274,10 +279,16 @@ private void deployBridgeInMemory(ExtensionContext extensionContext, Map<String,
                 }
             }).await();
 
-        HttpService httpService = new HttpService("127.0.0.1", 8080);
+        boolean sslEnabled = "true".equals(String.valueOf(configuration.get(HttpConfig.HTTP_SERVER_SSL_ENABLE)));
```
ppatierno: what about `Boolean.TRUE.equals(configuration.get(HttpConfig.HTTP_SERVER_SSL_ENABLE));` ?

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
```diff
@@ -293,6 +304,9 @@ private void deployBridgeInContainer(ExtensionContext extensionContext, Map<Stri
         properties.putAll(configuration);
         String propertiesPath = createPropertiesFile(properties);
 
+        boolean sslEnabled = "true".equals(String.valueOf(configuration.get(HttpConfig.HTTP_SERVER_SSL_ENABLE)));
```
ppatierno: Ditto as above.

ppatierno (approved):

src/test/java/io/strimzi/kafka/bridge/extensions/BridgeExtension.java
im-konge: Yeah good point, thanks.

im-konge: I had to do it like this:
```
Boolean.TRUE.toString().equals(configuration.get(HttpConfig.HTTP_SERVER_SSL_ENABLE).toString());
```
because in the configuration it is in `"true"` Object (so String) and otherwise it would also throw the NPE (if used differently)

im-konge: Or maybe this:
```
Boolean.parseBoolean(String.valueOf(configuration.get(HttpConfig.HTTP_SERVER_SSL_ENABLE)));
```
will be better, the `parseBoolean` handles the null as well.

PR #1124, opened 2026-05-14
Refactor of Bridge integration tests - `MetricsIT` and `SeekIT`

im-konge:
This PR refactors `MetricsIT` and `SeekIT`. 
The only class left now is the `ProducerIT` which will be done in separate PR.
It also adds the separate `HttpSeekService` (similar to how it is done until now), to not add more of the methods to the `HttpConsumerService` class.
Also, because I decided to have the management service in the `BridgeTestContext` (as it is used in the SeekIT) I decided to remove the management port key etc. and change it in the previous classes (OtherServicesIT and TlsIT)

It also fixes some issues with leaking connections to the Bridge and Kafka, which weren't closed in Kafka clients and HTTP services.

### DIFF
--- a/src/test/java/io/strimzi/kafka/bridge/httpclient/HttpSeekService.java
+++ b/src/test/java/io/strimzi/kafka/bridge/httpclient/HttpSeekService.java
@@ -4,13 +4,13 @@
  */
 package io.strimzi.kafka.bridge.httpclient;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.strimzi.kafka.bridge.BridgeContentType;
 import io.strimzi.kafka.bridge.http.model.HttpBridgeError;
 import io.strimzi.kafka.bridge.utils.Endpoints;
 
 import java.net.http.HttpResponse;
-import java.util.Map;
 
 /**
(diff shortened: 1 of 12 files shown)


src/test/java/io/strimzi/kafka/bridge/httpclient/HttpSeekService.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Seeks consumer to specified positions (offsets).
+     * Verifies that the response status is 204 NO_CONTENT.
+     *
+     * @param groupId       group ID to which consumer belongs to.
+     * @param consumerName  name of the consumer.
+     * @param body          request body containing offsets.
+     */
+    public void seekToPositions(String groupId, String consumerName, Map<String, Object> body) {
```
ppatierno: not sure if that's the same across all the other services but as you did already for the responses to get JSON, can we build a JSON as well to be passed as body instead of the Map? We can even proceed with this for now and then doing another pass of refactoring to change it.

im-konge: For me it was much easier to use Maps instead of building JSONs for the requests. But yes, it is actually everywhere in the HttpServices like Consumer. So it would require to fix it everywhere.

ppatierno (approved):

PR #1127, opened 2026-05-19
Refactor integration tests - final changes - clean up of Admin client stuff, use JsonNode instead of record classes for each response

im-konge:
This final PR cleans up the remaining things in the integration tests, mainly:

- Admin client facade class and AbstractIT - I made the `createTopic` method in the `AdminClientFacade` blocking and removed the methods from the `AbstractIT`
- cleaned up `AbstractIT` from the send message method - the `HttpProducerService` is used instead in the _one test_ that needed it
- I removed all of the `record` classes that were used for parsing every response for easier handling - instead I'm using JsonNode directly
- I changed from passing `Map<String, Object>` everywhere to `JsonNode` as discussed in previous PRs
- clean up of unused stuff

This is the last PR for the integration tests refactor, once this is merged, I'm going to open a PR to `main` branch.


ppatierno (approved):
LGTM overall. As said before we can refine when everything will be merged. When the big PR against main will be open I will also try everything again locally.

PR #1149, opened 2026-08-04
Fixed the script to get GC options and Terabytes to `to_byte` script

ppatierno:
### Type of change

- Task

### Description

The `get_gc_opts` still had an old check on Java 9 which is not supported anymore.
This PR aligns the function to what we use within the Strimzi operator without such check.
It also add the support for Terabytes to the 'to_byte.gawk` script as for the Strimzi operator.

### Checklist


scholzj (approved):
LGTM. But you should probably use a more corresponding name as it foes not change only the GC option.

PR #1170, opened 2026-08-15
Use a pattern matching switch in HttpBridge.errorHandler

### LINKED ISSUE #1113: Using types Java pattern matching within `HttpBridge.errorHandler` method
The `HttpBridge.errorHandler` method has a deep chain of `if/else if` based on the type of the `routingContext.failure()`.
It could be refactored by using a pattern matching on the corresponding possible types.
The solution would be a switch on `routingContext.failure()` and several case corresponding to each type.

gggeon96:
Replaces the if/else chain in `HttpBridge.errorHandler` with a pattern matching switch over `routingContext.failure()`, as proposed in the issue.

Two branches match on the failure itself and three on its cause, so the switch uses guarded patterns (`when`) to keep the original matching order. The schema validation errors are collected with a stream, which also keeps the method within the checkstyle cyclomatic complexity limit (guarded patterns each count toward it). Behaviour and response codes are unchanged; the existing `ProducerIT`/`ConsumerIT` assertions cover the error messages.

Closes #1113

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -59,7 +59,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


ppatierno (approved):
LGTM. Thanks!

tinaselenge (approved):
thanks for the PR. LGTM!

PR #1172, opened 2026-08-19
Fixed HTTP sink endpoint cleaning by centralizing the logic

ppatierno:
### Type of change

- Refactoring

### Description

I enabled some additional log in the current bridge and noticed the following:

```shell
2026-08-19 13:05:37 INFO  [vert.x-eventloop-thread-0] deleteConsumer:49 - [924287670] DELETE_CONSUMER Request: from 0:0:0:0:0:0:0:1:39994, method = DELETE, path = /consumers/my-group/instances/consumer1
2026-08-19 13:05:37 INFO  [vert.x-eventloop-thread-0] HttpBridge:472 - HttpBridge.deleteConsumer (thread = Thread[#53,vert.x-eventloop-thread-0,5,main] remove = ConsumerInstanceId(groupId=my-group, instanceId=consumer1)
2026-08-19 13:05:37 INFO  [vert.x-eventloop-thread-0] HttpBridge:473 - ---> is in the list? true
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] ConsumerCoordinator:1062 - [Consumer clientId=consumer1, groupId=my-group] Resetting generation and member id due to: consumer pro-actively leaving the group
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] ConsumerCoordinator:1109 - [Consumer clientId=consumer1, groupId=my-group] Request joining group due to: consumer pro-actively leaving the group
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] Metrics:684 - Metrics scheduler closed
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] Metrics:688 - Closing reporter org.apache.kafka.common.metrics.JmxReporter
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] Metrics:688 - Closing reporter org.apache.kafka.common.telemetry.internals.ClientTelemetryReporter
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] Metrics:694 - Metrics reporters closed
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] AppInfoParser:89 - App info kafka.consumer for consumer1 unregistered
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] HttpBridge:433 - HttpBridge.createConsumer closeHandler (thread = Thread[#68,kafka-bridge-executor-1,5,main] remove = ConsumerInstanceId(groupId=my-group, instanceId=consumer1)
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] HttpBridge:434 - ---> is in the list? false
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] HttpSinkBridgeEndpoint:310 - Deleted consumer consumer1 from group my-group
2026-08-19 13:05:37 INFO  [kafka-bridge-executor-1] deleteConsumer:58 - [924287670] DELETE_CONSUMER Response: statusCode = 204, message = No Content
```

The removal of a consumer when the DELETE command is issued by the HTTP client happens in two places. On the `deleteConsumer` (on the Vert.x event loop) but also on the sink `closeHandler` (on the async executor thread). The latter does usually nothing (because the consumer was already removed by the first call) but the current design can cause race condition. Finally, another deletion operation happens within the inactive consumer timer.

In more details, the HTTP Bridge maintains two maps for tracking active consumer endpoints: `httpSinkEndpoints` (storing endpoint instances) and `timestampMap` (tracking last activity for timeout detection). The cleanup logic for these maps is duplicated across the three above different locations.
Additionally, the cleanup is inconsistent with `timestampMap` only removed from in `deleteConsumer()` and the timer, but not in the `closeHandler`, while `httpSinkEndpoints` is cleaned up in all three places.

This PR fixes this issue by centralizing all cleanup logic into the `closeHandler` callback, which is the natural owner of this responsibility since it's invoked whenever an endpoint closes. The `deleteConsumer()` method and inactive consumer timer now simply call `close()` on the endpoint, and the `closeHandler` performs all map cleanup for both `httpSinkEndpoints` and `timestampMap`.

As a side benefit, this also eliminates a potential race condition where multiple threads could modify the HashMaps simultaneously.

### Checklist

- [x] AI assistance was used to create this PR (see the [Strimzi AI policy](https://github.com/strimzi/governance/blob/main/AI_POLICY.md))

### DIFF
--- a/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
+++ b/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
@@ -62,7 +62,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


tinaselenge (approved):
Thank you for the PR @ppatierno. It looks good to me. 

I just have a question that does not affect the PR necessarily regarding this bit in the description
```
As a side benefit, this also eliminates a potential race condition where multiple threads could modify the HashMaps simultaneously.
```
Does this mean, the clean up is now happening only in executor thread? Or is it still possible for the clean up to be executed in 2 different threads (therefore 2 threads modifying the maps), since the inactive consumer timer runs on the Vert.x event loop thread and calls deleteSinkEndpoint.close()?

ppatierno:
@tinaselenge what you said can happen if at the same time the inactive timer is triggered, so the removal from the map happens within the Vert.x event loop while at the same tile a DELETE operation is issued which runs in an executor thread. It could happen even when a new consumer is created so there is access to the same map. This issue was already here before the PR, so I think one way to fix it is using a `ConcurrentHashMap` instead. I will make the change in this PR. Thanks!

PR #1175, opened 2026-08-24
Add support for configuring TLS named groups on the HTTP server

ppatierno:
### Type of change

- Enhancement / new feature

### Description

This PR adds a new `http.ssl.named.groups` configuration property to control the SSL/TLS named groups (key-exchange groups) offered by the HTTP server, alongside the existing options for protocols and cipher suites.
The value is a comma-separated list (e.g. `x25519,secp256r1,secp384r1`) passed to Vert.x via `setKeyExchangeGroups(...)` method. When not set, the JVM defaults are used and the current behavior is unchanged.

This addition makes it possible to enable hybrid post-quantum key-exchange groups such as `X25519MLKEM768` on JVMs that support them (JDK 24+)  with no further code changes. This is valid for future JVMs but also for the current Java 21 on which it seems to be a commitment to backport support for PQC related JEPs at some point.

### Checklist

- [x] Update documentation
- [x] Update CHANGELOG.md (if present)
- [x] Write tests


src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
```diff
@@ -386,7 +386,9 @@ private ServerSSLOptions serverSSLOptions() {
         if (sslCipherSuites != null) {
             sslCipherSuites.forEach(serverSSLOptions::addEnabledCipherSuite);
         }
+        serverSSLOptions.setKeyExchangeGroups(this.bridgeConfig.getHttpConfig().getHttpServerSslNamedGroups());
```
tinaselenge, line 389: Do we not need to check if it's not null before setting it? Or setting it to null makes it JVM default?

tinaselenge (commented):
Mostly looks good to me, just have one question about setting the field.

src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java
ppatierno: Yes, setting null makes it using the JVM defaults as today.

tinaselenge (approved):
LGTM. Thanks, Paolo.

PR #1030, opened 2025-10-03
Add support for graceful shutdown

mlallaouret:
Fix https://github.com/strimzi/strimzi-kafka-bridge/issues/1008

Issue #1008, opened 2025-08-07
Add support for graceful shutdown


mlallaouret:
Hello,

When doing some load testing of kafka bridge on kubernetes and doing some rolling restart, some messages are not successful (http code 502) and so some data are lost..

Steps to reproduce:
- Deploy 2 replicas of kafka-bridge
- Use tool like fortio to generate multiple http calls over some time.
- While generating http calls, launch a rolling restart of the kafka bridge deployment

I tested to patch the kafka-bridge deployment with a preStop hook like the one below and didn't see any dropped messages anymore. 
```
lifecycle:
  preStop:
    exec:
      command: ["sh", "-c", "sleep 10"]
```
The issue is we cannot specify pre stop hook in the kafkabridge custom resource for now.
However, a cleaner solution seems to be to implement [vertx graceful shutdown](https://vertx.io/docs/vertx-core/java/#_graceful_shutdown) but I didn't tested this one as it requires code changes.

What do you think ?

ppatierno:
Hi, it's not clear to me how you are deploying the bridge.
You are mentioning Deployment but then referring the KafkaBridge custom resource, so it's not clear if you are using the bridge within a container image and then using your own Deployment or using the Strimzi operator with the KafkaBridge custom resource.
Another thing is that the KafkaBridge can't have more than one replicas because how it works, mostly from a consumer perspective. Each consumer is tight to one instance and if you put a Service in front of it which goes in round robin you would not reach the right HTTP bridge instance where your consumer "lives".
Said that, a graceful shutdown would be useful ... are you willing to contribute to it, by making the changes, testing and open a PR? Thanks!

ppatierno:
Ok now it's clearer thanks! :-)

> For now, I use it only to produce messages so I am not impacted with the instance stickyness issue (unless I am missing something ?)

Yes the producer side is not impacted by the instance stickyness so you are "safe" from this pov.

Anyway, I think having a graceful shutdown would be good.
Not sure when someone will come to fix it (I can't right now but I could in the next weeks) unless you are willing to contribute to Strimzi. It would be really welcome! :-)

KeepItSimpleStupid:
Jumping in as a colleague of @mlallaouret who opened this issue : we're still interested to contribute this built-in graceful shutdown to the bridge :)

[...]

PR #1183, opened 2026-09-17
Block the shutdown hook until vertx.close() completes

### LINKED ISSUE #1184: [Bug]: Graceful shutdown never runs
### Bug Description

The graceful shutdown added by #1030 (for #1008) is triggered on SIGTERM and then cut short. The shutdown hook calls `vertx.close()` and attaches an `onComplete` callback without waiting on the returned future, so the hook thread returns immediately. The JVM waits only for shutdown hook *threads* to finish and then halts, so verticle undeployment never completes: the bridge exits roughly half a second after SIGTERM, regardless of how much grace period it is given.

Undeployment does start. `HttpBridge.stop()` gets through closing the sink and source endpoints (and the `KafkaProducer` each source endpoint owns) and the admin client endpoint, and the JVM then halts before the joined `apiServer.shutdown()` and `managementServer.shutdown()` complete. Those two are exactly what #1030 changed from `close()` to `shutdown()` in order to obtain a graceful drain, so the 30 second Vert.x grace period for in-flight requests never applies and clients see connections cut mid-request.

`vertx.close()` is asynchronous. The hook registers a callback and its thread terminates at once. Once JVM shutdown has begun, non-daemon threads no longer keep the process alive, so the JVM halts as soon as the hook threads are done and whatever `vertx.close()` had started is dropped. Relevant code: [`Application.java#L92-L102`](https://github.com/strimzi/strimzi-kafka-bridge/blob/1.1.0/src/main/java/io/strimzi/kafka/bridge/Application.java#L92-L102) and [`HttpBridge.stop()#L319-L356`](https://github.com/strimzi/strimzi-kafka-bridge/blob/1.1.0/src/main/java/io/strimzi/kafka/bridge/http/HttpBridge.java#L319-L356).

I have taken the liberty of opening a PR here: https://github.com/strimzi/strimzi-kafka-bridge/pull/1183

### Steps to reproduce

Docker only, no Kubernetes required, about two minutes end to end.

**Set `-Dlog4j2.shutdownHookEnabled=false`**, as the steps below do. log4j2 registers its own shutdown hook, shutdown hooks run concurrently, and logging is torn down while the bridge is still shutting down. Without the flag every log line after the first disappears whether or not it was reached, which makes the failure impossible to read accurately.

### 1. Config files

`custom-config/application.properties`:

```properties
bridge.id=repro-bridge
kafka.bootstrap.servers=kafka:9092
kafka.producer.acks=1
http.host=0.0.0.0
http.port=8080
http.management.port=8081
http.producer.enabled=true
http.consumer.enabled=true
```

`custom-config/log4j2.properties` (the entrypoint points log4j at this path; without it you get a fallback config):

```properties
name = BridgeConfig
appender.console.type = Console
appender.console.name = STDOUT
appender.console.layout.type = PatternLayout
appender.console.layout.pattern = %d{ISO8601} %-5p [%t] %c{1}:%L - %m%n
rootLogger.level = INFO
rootLogger.appenderRefs = console
rootLogger.appenderRef.console.ref = STDOUT
```

### 2. Single-node Kafka

```bash
docker network create bridge-repro

docker run -d --name kafka --network bridge-repro \
  -e KAFKA_NODE_ID=1 -e KAFKA_PROCESS_ROLES=broker,controller \
  -e KAFKA_LISTENERS=PLAINTEXT://:9092,CONTROLLER://:9093 \
  -e KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://kafka:9092 \
  -e KAFKA_CONTROLLER_QUORUM_VOTERS=1@kafka:9093 \
  -e KAFKA_CONTROLLER_LISTENER_NAMES=CONTROLLER \
  -e KAFKA_LISTENER_SECURITY_PROTOCOL_MAP=CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT \
  -e KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1 \
  -e KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR=1 \
  -e KAFKA_TRANSACTION_STATE_LOG_MIN_ISR=1 \
  apache/kafka:3.9.0
```

### 3. The bridge

The command mirrors what the cluster operator runs: `bin/docker/kafka_bridge_run.sh` reads `$STRIMZI_HOME/custom-config/application.properties` and tees it to `/tmp/kafka-bridge.properties`, which is the path `--config-file` must point at.

```bash
docker run -d --name bridge --network bridge-repro -p 18080:8080 \
  -e STRIMZI_JAVA_OPTS="-Dlog4j2.shutdownHookEnabled=false" \
  -v "$PWD/custom-config:/opt/strimzi/custom-config:ro" \
  quay.io/strimzi/kafka-bridge:1.1.0 \
  /opt/strimzi/bin/docker/kafka_bridge_run.sh --config-file=/tmp/kafka-bridge.properties
```

### 4. Produce once, then SIGTERM with a generous grace period

```bash
curl -s -X POST http://localhost:18080/topics/repro-topic \
  -H 'Content-Type: application/vnd.kafka.json.v2+json' \
  -d '{"records":[{"value":{"hello":"world"}}]}'

time docker stop --time 60 bridge

for m in "Shutdown hook triggered" \
         "Stopping HTTP bridge verticle" \
         "HTTP bridge has been shut down successfully" \
         "Vertx instance closed successfully"; do
  printf '%-46s %s\n' "$m:" "$(docker logs bridge 2>&1 | grep -c "$m")"
done
```

### Result

```
container exited after 0.55s of a 60s grace period, exit code 0

Shutdown hook triggered:                        1
Stopping HTTP bridge verticle:                  1
HTTP bridge has been shut down successfully:    0
Vertx instance closed successfully:             0
```

The last line logged before the process exits is the admin client unregistering, so `stop()` runs its endpoint and admin client cleanup and the JVM then halts before the two HTTP server shutdowns join.

Notes:

- No in-flight request is needed, and the broker does not need to be unhealthy. I first built this with `docker pause` on Kafka to make `KafkaProducer.close()` block and force a long drain; that turned out to be unnecessary, since the behaviour is identical with a healthy broker and no traffic.
- Exit code is 0, not 143, so this is the JVM finishing its own shutdown sequence rather than being SIGKILLed by the container runtime after the grace period.

### With #1183 applied

Same environment and steps, with `io.strimzi.kafka-bridge-1.1.0.jar` in the image swapped for a jar built from #1183:

| Log line | Stock 1.1.0 | With #1183 |
|---|---|---|
| `Shutdown hook triggered, starting graceful shutdown...` | 1 | 1 |
| `Stopping HTTP bridge verticle ...` | 1 | 1 |
| `HTTP bridge has been shut down successfully` | 0 | 1 |
| `Vertx instance closed successfully` | 0 | 1 |

### Also observed on Kubernetes

Single-replica bridge deployed by the cluster operator, `terminationGracePeriodSeconds: 30`. Note this capture predates the log4j2 finding above, so only the first line survives:

```
10:50:34.000Z  pod deletion issued
10:50:36.195Z  GET / answered 200        <- serving normally
10:50:36.208Z  "Shutdown hook triggered, starting graceful shutdown..."
               (no further output, container gone)
```

### Expected behavior

On SIGTERM the bridge drains: `HttpBridge.stop()` closes the sink and source endpoints and the admin client, shuts down the async executor, then calls `shutdown()` on the API and management HTTP servers, which on Vert.x 5 carries a 30 second default grace period for in-flight requests. The process exits once that completes.

The log should show, in order:

```
Shutdown hook triggered, starting graceful shutdown...
Stopping HTTP bridge verticle ...
HTTP bridge has been shut down successfully
Vertx instance closed successfully
```

Instead the last two never appear, and the process exits about half a second after SIGTERM with the remaining grace period unused.

### Strimzi version

Kafka Bridge 1.1.0 (`quay.io/strimzi/kafka-bridge:1.1.0`, `sha256:3a03af635b3ba64505527864bd0f6ad7b5118637491d8a5b32be78e11733e65a`), deployed by Cluster Operator 1.2.0. Vert.x 5.1.6, OpenJDK 21.0.12.

### Kubernetes version

1.35.7 (GKE) for the Kubernetes capture above. The Docker reproduction does not involve Kubernetes.

### Installation method

Helm chart

### Infrastructure

EKS, GKE

### Configuration files and logs

_No response_

### Additional context

Possibly a separate issue: log4j2's own shutdown hook tears logging down concurrently with the bridge's, so shutdown-time logs are unreliable by default. That is what made this failure hard to read. It may be worth setting `shutdownHook="disable"` in the shipped log4j2 configuration so operators can see what happened during termination.

spectrogram:
The shutdown hook added in #1030 (for #1008) calls `vertx.close()` and attaches an `onComplete` callback without waiting on the returned future, so the hook thread returns immediately. The JVM waits only for shutdown hook threads before halting, so the graceful shutdown is cut off partway through verticle undeployment.

This waits on the close future with a timeout, so the hook holds the JVM open until the shutdown either completes or gives up.

## Evidence

A/B on `quay.io/strimzi/kafka-bridge:1.1.0` against a single-node Kafka in Docker, identical config, one record produced, then `docker stop --time 60`. The only difference is `io.strimzi.kafka-bridge-1.1.0.jar` swapped for a jar built from this branch.

Both runs set `-Dlog4j2.shutdownHookEnabled=false`. Without it the results are misleading: log4j2 registers its own shutdown hook, hooks run concurrently, and logging is torn down while the bridge is still shutting down, so the later lines are lost whether or not they were reached.

| Log line | Stock 1.1.0 | This branch |
|---|---|---|
| `Shutdown hook triggered, starting graceful shutdown...` | 1 | 1 |
| `Stopping HTTP bridge verticle ...` | 1 | 1 |
| `HTTP bridge has been shut down successfully` | **0** | 1 |
| `Vertx instance closed successfully` | **0** | 1 |

On the stock jar the last thing logged before the process exits is the admin client unregistering, so `HttpBridge.stop()` gets through the endpoint and admin client closes and then the JVM halts before `apiServer.shutdown()` and `managementServer.shutdown()` join. Exit is 0.55s into a 60 second grace period, with exit code 0, so this is the JVM completing its own shutdown rather than being killed by the runtime.

The same truncation is visible in a Kubernetes deployment with the cluster operator and `terminationGracePeriodSeconds: 30`: last request served at `10:50:36.195Z`, hook logged at `10:50:36.208Z`, container gone immediately after.

## Open question

The timeout is a fixed 30 seconds here, matching Vert.x's own default HTTP server grace period. That may want to be configurable: the Kubernetes default `terminationGracePeriodSeconds` is also 30, which leaves no margin for the endpoint closes, producer closes, admin client and executor that `stop()` runs first, so an operator wanting a guaranteed clean drain has to raise the pod grace period without any way to align the two. Happy to add a bridge property instead, or to lower the constant, whichever you prefer.

## Verification

`mvn verify -DskipTests` passes locally (checkstyle, compile, testCompile, javadoc, dependency analyze) on JDK 21. The test suite was left to CI.


src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -30,13 +30,21 @@
...
 /**
  * Apache Kafka bridge main application class
  */
 public class Application {
     private static final Logger LOGGER = LogManager.getLogger(Application.class);
 
+    /**
+     * How long the shutdown hook waits for the Vert.x instance to close before giving up
+     */
+    private static final long SHUTDOWN_TIMEOUT_SECONDS = 30;
```
scholzj: Hardcoding the shutdown period here means it is not configurable in any environment (e.g. trough termination grace period in Kubernetes) and it would terminate early even if you want to give it more time, or?

scholzj (commented):
Thanks for the PR. You are right that the old code that just sets the handler does not look good. But I left some comment.

src/main/java/io/strimzi/kafka/bridge/Application.java
spectrogram: I'm happy to make it configurable - I wasn't sure if that's what we wanted but it sounds like we'd prefer that. 😄 

EDIT: done. It is `bridge.shutdown.timeout.seconds` which defaults to the current 30s.

scholzj: I do not think it should be configurable. I think it should wait until it shutdowns or is force killed.

spectrogram: My bad. I've taken it out completely.

ppatierno:
@spectrogram thanks for the PR. Can you rebased against the main branch please? While testing your PR I found another bug. Thanks!

src/main/java/io/strimzi/kafka/bridge/Application.java
```diff
@@ -91,14 +92,24 @@ private static Future<HttpBridge> deployHttpBridge(BridgeConfig bridgeConfig)
...
+                        // hook threads and then halts without waiting for any other thread, so returning
+                        // while vertx.close() is still in flight abandons the graceful shutdown.
+                        // There is no timeout here on purpose: the environment already bounds the wait,
+                        // for example by the Kubernetes termination grace period after which the process
+                        // is force killed.
+                        try {
+                            vertx.close()
+                                    .toCompletionStage()
+                                    .toCompletableFuture()
+                                    .get();
```
ppatierno: I was wondering if a simple:

```java
vertx.close().await();
```

would be enough here. I tried and it has the same effect than going through `CompletionStage` and `CompletionFuture`. Wdyt?

spectrogram: Makes sense since we no longer set the timeout here. The less code the better 😄

ppatierno: Anyway, if timeout was needed, there is the `await(long timeout, TimeUnit unit)` overload as well ;-)

ppatierno:
@spectrogram everything looks fine and thanks for the catch but I think this also deserve one line within the CHANGELOG. Would you mind adding it? Maybe something along this lines: "Fixed graceful shutdown on SIGTERM not completing, causing in-flight requests to be dropped." ?

scholzj (approved):

ppatierno (approved):
LGTM. Thanks for the PR!

